PR #2, opened 2015-09-16
[FEATURE] BooleanNode accepts non-Nodes in constructor

NamelessCoder:
Relaxing the required input type for the constructor of
BooleanNode allows simple values to be passed, avoiding
the need for additional wrapping of plain types in Nodes
before being able to evaluate them as booleans.


NamelessCoder:
@bwaidelich would you mind taking this one for a spin? It should solve the issue reported in https://github.com/NamelessCoder/TYPO3.Fluid/pull/22

bwaidelich:
:+1: looking good ;)

PR #3, opened 2015-09-18
[BUGFIX] Remove ``render()`` method from ``AbstractConditionViewHelper``

bwaidelich:
This removes the render() method from the AbstractViewHelper
class in order to prevent PHP errors when a subclass overrides the
method with a different signature (i.e. including parameters).

Note:
This was lost when re-applying https://github.com/NamelessCoder/TYPO3.Fluid/pull/18
to this repository.
This change also adds the ``AbstractViewHelperTest`` changes that got
lost.

### DIFF
--- a/src/Core/ViewHelper/AbstractConditionViewHelper.php
+++ b/src/Core/ViewHelper/AbstractConditionViewHelper.php
@@ -63,21 +63,6 @@ abstract class AbstractConditionViewHelper extends AbstractViewHelper {
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


NamelessCoder:
This is on hold for a short while still - has implications in CMS adapter package which currently does not replicate these methods and would break from this change.

NamelessCoder:
CMS adapter merge request is now fitted with this method. Confirmed existing in IfViewHelper. No other condition ViewHelpers exist at this time which depend on this.

PR #4, opened 2015-09-18
[TASK] Improve error message when unable to resolve template

NamelessCoder:
Example error message after change:

"Tried resolving a template file for controller action "MyController->myAction" in format ".html", but none of the paths contained the expected template file (Mycontroller/MyAction.html). The following paths were checked: /foo, /bar"


mneuhaus:
+1 :)

PR #12, opened 2015-09-23
[TASK] Remove type hint for evaluateCondition method

NamelessCoder:
Removing the type hint keeps this base class compatible with current condition ViewHelpers in CMS which is a pre-requisite before releasing Fluid in version 1.0.0.


mneuhaus:
:+1:

PR #16, opened 2015-09-24
[FEATURE] Include line, character no. and Fluid code in parse errors

NamelessCoder:
Like the subject says, Fluid parser errors now echo the template identifier (unfortunately not the filename since there is no file name in the parser), the line number and the character number when parsing the template source. The pointers can then be included in a re-thrown error message.

Example error:

```
TYPO3Fluid\Fluid\Core\Parser\Exception: Fluid parse error in template Default_action_Default_1cb8dc11e29962882f629f79c0b9113ff33d6219, line 11 at character 3. Error: The ViewHelper "<f:serender>" could not be resolved.
Based on your spelling, the system would load the class "TYPO3Fluid\Fluid\ViewHelpers\SerenderViewHelper", however this class does not exist. (1407060572). Template code: <f:serender section="Foo" optional="1">
```

### DIFF
--- a/src/Core/Parser/TemplateParser.php
+++ b/src/Core/Parser/TemplateParser.php
@@ -63,6 +63,21 @@ class TemplateParser {
 	protected $templateProcessors = array();
 
 	/**
+	 * @var integer
+	 */
+	protected $pointerLineNumber = 1;
+
+	/**
+	 * @var integer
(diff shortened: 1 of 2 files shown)


mneuhaus:
:dancers: :+1: :dancers:

PR #17, opened 2015-09-24
[FEATURE] Support Legacy Namespaces and throw Exception on unknown namespace use

### LINKED ISSUE #8: Throw exception for unresolved namespaces
Unknown namespaces used to be ignored in Fluid. But that lead to serious issues, e.g. when extracting template code into a partial while forgetting the namespace imports.
S.th. like
```html
<media:image image="{image}" />
```
Would try to render '{image}` directly if the `media:image` ViewHelper was ignored, leading to various errors that are very hard to debug.

With Flow 3.0 we therefore introduced a stricter handling of unknown namespaces, see: https://jira.neos.io/browse/FLOW-150

```html
<unknownNamespace:foo />
```

## Expected output

```
Error while rendering a ViewHelper
The namespace of ViewHelper notation "<unknownNamespace:foo.../>" could not be resolved.

Possible reasons are:
* you have a spelling error in the viewHelper namespace
* you forgot to import the namespace using "{namespace unknownNamespace=Some\Package\ViewHelpers}"
* you're trying to use a non-fluid xml namespace, in which case you can use "{namespace unknownNamespace}" to ignore this namespace for fluid rendering
```

## Actual output

```
<unknownNamespace:foo />
```

mneuhaus:
This commit adds back the "old" namespace syntax:

{namespace foo=Foo\Bar\ViewHelpers}

and it checks if you're using any viewHelpers in namespaces you didn't
register to reduce common problems where you move stuff in templates,
spelling errors and similar which result in not-rendered viewHelpers
or even exceptions because as a side effect arguments are cast to
string, but fail.

fixes  #8

### DIFF
--- a/src/Core/Parser/Patterns.php
+++ b/src/Core/Parser/Patterns.php
@@ -20,6 +20,11 @@ abstract class Patterns {
 	static public $SPLIT_PATTERN_TEMPLATE_OPEN_NAMESPACETAG = '/xmlns:([a-z0-9\.]+)=("[^"]+"|\'[^\']+\')*/xi';
 
 	/**
+	 * @var string
+	 */
+	static public $NAMESPACE_DECLARATION = '/(?<!\\\\){namespace\s*(?P<identifier>[a-zA-Z\*]+[a-zA-Z0-9\.\*]*)\s*(=\s*(?P<phpNamespace>(?:[A-Za-z0-9\.]+|Tx)(?:\\\\\w+)+)\s*)?}/m';
+
+	/**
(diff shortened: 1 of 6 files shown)


bwaidelich:
This does not ignore template code within CDATA tags!

bwaidelich:
i.e.
```
<![CDATA[{nonExistingNamespace:basr()}]]>
```

## Expected:

```
{nonExistingNamespace:bar()}
```

## Actual:

```
Error: Unkown Namespace: nonExistingNamespace:bar() (0). Template code: 
</body>
</html>
```

PR #18, opened 2015-09-26
[TASK] Throw same Exception type as caught in Parser

NamelessCoder:
Exceptions converted by the Parser into pretty errors with template line, char and snippet are now thrown as the same type of Exception that was originally caught. This allows external users to check by class which type of Exception was thrown without having to read/parse the error message.


NamelessCoder:
Thanks Marc - hope this helps with the other open PR ;)

mneuhaus:
yes it does, pr-update coming up ;)

PR #19, opened 2015-10-02
[FEATURE] SpacelessViewHelper to trim whitespace outside tags

NamelessCoder:
```xml
<f:spaceless>
<div>
    <div>
        <div>text

text</div>
    </div>
</div>
</f:spaceless>
```

Outputs:

```html
<div><div><div>text

text</div></div></div>
```


NamelessCoder:
See also https://review.typo3.org/#/c/43747/

PR #21, opened 2015-10-09
[FEATURE] Passing of tag content from f:render to partial/section

NamelessCoder:
Allows a completely new approach to structuring Fluid template rendering, allowing partials and sections to be used as "wrappers" for an arbitrary piece of template code. Consider the following example:

```xml
<f:section name="MyWrap">
    <div>
        <!-- And LOTS more HTML, using variables if desired -->
        <!-- Tag content of f:render output: -->
        {contentVariable -> f:format.raw()}
    </div>
</f:section>

<f:render section="MyWrap" contentAs="contentVariable">
    This content will be wrapped. Any Fluid code can go here.
</f:render>
```

The result is that the tag content of `f:render` is assigned as the template variable `{contentVariable}` which can then be used in the template code that is in the section/partial. It's the same as passing a big chunk of Fluid to the `arguments` array as a variable there, and using it the same way - but it is a lot easier on the syntax.

The new structure approach is enabled because `f:render` can then be used recursively to wrap your template code in any number of wraps (imagine the `f:section` tags with content similar to above):

```xml
<f:render partial="OuterWrap" contentAs="contentVariable">
    <f:render partial="InnerWrap" contentAs="contentVariable">
        <f:render partial="ItemWrap" contentAs="contentVariable">
            <h2>My item, nicely wrapped</h2>
            <p>Nice.</p>
        </f:render>
   </f:render>
</f:render>
```

Which would render as follows:

1. Outermost wrap renders tag content and assigns as variable for second level
2. Second level wrap renders tag content and assigns as variable for innermost level
3. Innermost level renders the actual content you wish to wrap and assigns it as variable
4. The innermost "wrap" partial/section is rendered (because no more `f:render` children exist)
5. The second level "wrap" partial/section is rendered with the HTML from step 4 as content variable.
6. The outer level "wrap" partial/section is rendered with the HTML from step 5 as content variable.
7. The entire thing is output.

### DIFF
--- a/src/ViewHelpers/RenderViewHelper.php
+++ b/src/ViewHelpers/RenderViewHelper.php
@@ -83,6 +83,7 @@ class RenderViewHelper extends AbstractViewHelper {
 		$this->registerArgument('arguments', 'array', 'Array of variables to be transferred. Use {_all} for all variables', FALSE, array());
 		$this->registerArgument('optional', 'boolean', 'If TRUE, considers the *section* optional. Partial never is.', FALSE, FALSE);
 		$this->registerArgument('default', 'mixed', 'Value (usually string) to be displayed if the section or partial does not exist', FALSE, NULL);
+		$this->registerArgument('contentAs', 'string', 'If used, renders the child content and adds it as a template variable with this name for use in the partial/section', FALSE, NULL);
 	}
 
 	/**
@@ -96,6 +97,13 @@ class RenderViewHelper extends AbstractViewHelper {
 		$partial = $this->arguments['partial'];
 		$arguments = (array) $this->arguments['arguments'];
 		$optional = (boolean) $this->arguments['optional'];
+		$contentAs = $this->arguments['contentAs'];
+		$tagContent = $this->renderChildren();
+
+		if ($contentAs !== NULL) {
+			$arguments[$contentAs] = $tagContent;
+		}
+
 		$content = '';
 		if ($partial !== NULL) {
 			$content = $this->viewHelperVariableContainer->getView()->renderPartial($partial, $section, $arguments, $optional);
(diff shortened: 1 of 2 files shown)


NamelessCoder:
Another use case just for the heck of it:

```xml
<f:render partial="{outerWrapType}" contentAs="contentVariable" optional="1">
    <f:render partial="{innerWrapType}" contentAs="contentVariable" optional="1">
        <f:render partial="{ItemWrapType}" contentAs="contentVariable" optional="1">
            <h2>My item, nicely wrapped</h2>
            <p>Nice.</p>
        </f:render>
   </f:render>
</f:render>
```

Making every wrapping 1) dynamic and selectable with a template variable, 2) optional and simply not rendered if not found or type not set. Secondarily, allows disabling a wrapping by setting the template variable to an empty or purposely invalid value.

mneuhaus:
dude, not cool! 
You totally stole this idea from me! (https://gist.github.com/mneuhaus/26b54e0698d0f8c644a6)
i'm outraged!
i'll send my lawyers after you'r stuff, especially your Monitor!!!

just kidding of course ;) +1 from me, works as a seamless replacement for my BlockViewHelper

mneuhaus:
here's another example to massively simplify form rendering with this little magic:

**Partials/FormFieldWrap.html**

```html
<div class="form-group">
	<label for="{id}" class="col-sm-3 control-label">
		{label}
	</label>
	<div class="col-sm-9">
		{control -> f:format.raw()}

		<f:if condition="{infotext}"><p class="help-block">{infotext}</p></f:if>
	</div>
</div>
```

**Using the FormFieldWrap Partial in a template**

```html
<f:render partial="FormFieldWrap" arguments="{id: 'foo', label: 'Foo!', infotext: 'lorem ipsum dolor set amet'}" contentAs="control">
	<textarea name="foo" id="foo"></textarea>
</f:render>
```

**Output**

```html
<div class="form-group">
	<label for="foo" class="col-sm-3 control-label">
		Foo!
	</label>
	<div class="col-sm-9">
		<textarea name="foo" id="foo"></textarea>

		<p class="help-block">lorem ipsum dolor set amet</p>
	</div>
</div>
```

Issue #7, opened 2015-09-23
Boolean ViewHelper arguments comparing numeric values behave differently


bwaidelich:
Not to argue what makes more sense, but to report the *different* behavior

# "==" operator, comparing strings with integers:

```html
<f:if condition="'foo' == 0">
  <f:then>then</f:then>
  <f:else>else</f:else>
</f:if>
```

## Expected output

```
then
```

## Actual output

```
else
```

# ">=" operator, comparing strings with floats:

```html
<f:if condition="'1.1 >= 'foo'">
  <f:then>then</f:then>
  <f:else>else</f:else>
</f:if>
```

## Expected output

```
then
```

## Actual output

```
else
```

# ">" operator, comparing strings with integer:

```html
<f:if condition="'foo' > 0">
  <f:then>then</f:then>
  <f:else>else</f:else>
</f:if>
```

## Expected output

```
else
```

## Actual output

```
then
```

The reason for this new behavior is that now both $left and $right are *strings* in `BooleanNode::evaluateComparator()`. Before numeric values were converted to `NumericNode` instances with

```php
if (is_numeric($value)) {
	$this->leftSide->addChildNode(new NumericNode($value));
} else {
	$this->leftSide->addChildNode(new TextNode(preg_replace('/(^[\'"]|[\'"]$)/', '', $value)));
}
```

in the constructor.

albe:
See also https://forge.typo3.org/issues/6757#note-17 on background explanation why that was needed

NamelessCoder:
Can't help but chuckle a bit over these comparisons :)

I guess we have to support each one explicitly?

bwaidelich:
> Can't help but chuckle a bit over these comparisons :)

As said, this is mainly a record of the *changed* behavior. But as Alexander wrote, it makes sense if you compare it with PHPs behavior. And for strict comparisons we'll have the `===` operator AFAIS.

Personally I find `'foo' == 0` => TRUE also weird and one should use such code in Fluid to begin with, but I could bet that people out there already rely on that behavior.

> I guess we have to support each one explicitly

This was just an extract, here's a more complete list: http://134.119.11.221/flowpack.viewhelpertest/standard/index?selectedPartial=if

We don't have to support them explicitly, just make sure that numeric values won't get casted to strings .

Issue #6, opened 2015-09-21
Boolean ViewHelper arguments containing "false" are considered FALSE


bwaidelich:
```html
<f:if condition="anArbitary String containing FaLsE">
  <f:then>then</f:then>
  <f:else>else</f:else>
</f:if>
```
## Expected output

```
then
```

## Actual output

```
else
```

Also:
```html
<f:if condition="   FALSE   ">
  <f:then>then</f:then>
  <f:else>else</f:else>
</f:if>
```
## Expected output

```
then
````

## Actual output

```
else
```

**Comment by NamelessCoder** [comment]
Interesting! The trimmed-to-`FALSE` case makes sense (imho this whitespace-surrounded condition is an edge case; we could support it with an added check) but the other one is quite strange, having any casing of `false` anywhere in the string... broken regular expression?

**Comment by NamelessCoder** [comment]
Hang on a second here. This syntax will never, ever happen - not even with an unrolled ObjectAccessorNode returning a string that is then fed to the BooleanNode (e.g. `<f:if condition="{variableContainingStringContainingFalse}" then="yes" else="no" />` still correctly returns "yes".

Correctly quoting the hardcoded string has a completely different and correct result. Un-quoted strings *are intended to be parsed as possible expression parts separated by whitespace* so your node actually results in this stack:

```php
$parts = array('anArbitary', 'String', 'containing', 'FaLsE');
```

...which would be fine if each of those words instead were a number, a comparator, a Fluid variable reference etc.

I've added some new test cases in the linked commit which test for the proper syntax variation of this first reported issue and they return TRUE as they should. However: the current string trimming behavior is indeed unpredictable; it would forcibly trim both quotes **and** any whitespace in the string that was quoted, whether or not the string came from a hardcoded TextNode or an ObjectAccessorNode. The linked patch fixes that situation by first checking if an encountered string is quoted and if it is, carefully trimming only quotes from it. It is done by trimming only the first character of the string so that any nested quotes like `'"nested"'` becomes `"nested"` rather than `nested`. In addition, the constructor of BooleanNode is made responsible for the initial trimming of the expression when the expression is a plain old string (meaning it comes hardcoded from Fluid template).

bwaidelich:
Hi,

> Hang on a second here. This syntax will never, ever happen

But doesn't this example proof the opposite: http://134.119.11.221/flowpack.viewhelpertest/standard/index?selectedPartial=if ?

NamelessCoder:
[...]

PR #22, opened 2015-10-11
[BUGFIX] Match PHP behavior in deciding comparability of variables

NamelessCoder:
This change:

* Uses PHP math workaround ($value + 0) to ensure numeric-ness of unqouted values.
* Changes the analysis of left/right sides to use is_scalar as basic decision; allowing scalar values to be compared readily with null and resource pointers.
* Changes test expectations that were previously off PHP standard.
* Adds new test cases for the PHP standard comparison expectations.

Close: #7


mneuhaus:
:+1:

bwaidelich:
I can confirm that this fixes the condition tests in http://134.119.11.221/flowpack.viewhelpertest/standard/index?selectedPartial=if (#6 still applies).

Well done!

Issue #5, opened 2015-09-21
Reuse ViewHelper instances


bwaidelich:
The current implementation contains a change in behavior that prevents some ViewHelpers to work as before: Previously a ViewHelper instance was reused for the same node in the structure. Now a new instance is created every time.
Aside from a potential increase in memory consumption this leads to the following problem:

```html
<f:for each="{0:1, 1:2, 2:3, 3:4}" as="foo">
	<f:cycle values="{0: 'foo', 1: 'bar', 2: 'baz'}" as="cycle">{cycle}</f:cycle>
</f:for>
```

## Expected output

```
foo
bar
baz
foo
```

## Actual output

```
foo
foo
foo
foo
```

Although we could maybe adjust the ``cycle`` ViewHelper to cater for that new behavior, it will probably break some 3rd party ViewHelpers, too.

NamelessCoder:
Current behavior is to simply create a plain, new instance every time ViewHelperResolver is asked to create an instance - the goal being complete cleanliness and lifecycle predictability. That being said, if we can document a performance problem (in ViewHelpers *not belonging to TYPO3 or Flow and thus depending on complex DI*) I'd be happy to discuss if instance reuse should be introduced in the base package.

But for now, I would prefer to solve this one by doing two things:

1. Make the core ViewHelper(s) not depend on instance reuse but instead a ViewHelperVariableContainer storage.
2. Introduce forced instance reuse in each adapter package, in ViewHelperResolver, based simply on the class name requested. And call `resetState()` again which I think currently isn't happening.

Any objections to this strategy?

bwaidelich:
> Make the core ViewHelper(s) not depend on instance reuse but instead a ViewHelperVariableContainer storage.

That won't work because only the ViewHelper at the *same syntax tree* location should keep it's state.
Otherwise

```html
<f:for each="{array1}" as="foo">
    <f:cycle values="{0: 'foo', 1: 'bar', 2: 'baz'}" as="cycle">{cycle}</f:cycle>
    <f:for each="{array2}" as="subfoo">
      <f:cycle values="{0: 'subfoo', 1: 'subbar', 2: 'subbaz'}" as="cycle">{cycle}</f:cycle>
    </f:for>
    <f:cycle values="{0: 'bar', 1: 'foos'}" as="cycle">{cycle}</f:cycle>
</f:for>
```

would lead to weird results.

For the same reason reusing the instance based on the class name probably won't work

[...]

PR #25, opened 2015-10-12
[BUGFIX] Let f:cycle use "as" argument as scope of cycled variable

NamelessCoder:
This change allows f:cycle to be nested (see linked issue) - to work with multiple cycled variables in the same loop. In essence it identifies the instance of "values" using the value of the "as" argument and uses ViewHelperVariableContainer for storing the index.

Also changes the ViewHelper to not store the values but rather just store the index and read out the appropriate value based on that, from the provided values in arguments.

References: #5

### DIFF
--- a/src/ViewHelpers/CycleViewHelper.php
+++ b/src/ViewHelpers/CycleViewHelper.php
@@ -53,20 +53,6 @@ class CycleViewHelper extends AbstractViewHelper {
 	protected $escapeOutput = FALSE;
 
 	/**
-	 * The values to be iterated through
-	 *
-	 * @var array|\SplObjectStorage
-	 */
-	protected $values = NULL;
-
-	/**
-	 * Current values index
-	 *
-	 * @var integer
-	 */
-	protected $currentCycleIndex = NULL;
-
-	/**
 	 * @return void
 	 */
 	public function initializeArguments() {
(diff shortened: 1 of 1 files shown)


NamelessCoder:
I think this can be abandoned (underlying issue of reusing instances is solved). @bwaidelich and @mneuhaus your two cents please? If both agree, last one of you can just go ahead and close it.

mneuhaus:
fine with me i guess :)

mneuhaus:
@bwaidelich could you post your 2 cents about this? we have a "related" issue which would break the old behavior of the cycle viewHelper because of a Bug we need to fix, which the cycle kinda depends on to work. see discussion in https://github.com/TYPO3Fluid/Fluid/pull/77#issuecomment-195309441

NamelessCoder:
Issue in #77 appears to be resolved - which means once again, this pull request should probably be abandoned. I still lean towards not reusing instances (this recent issue is just an example of the complication it adds) but it is in no way urgent and there are probably plenty other (better) ways to solve this particular `f:cycle` issue.

bwaidelich:
Hi guys,

sorry for the lack of feedback, this one escaped me.

But sure, if #77 is resolved and the following template code
```
<f:for each="{0: 'a', 1: 'b', 2: 'c'}" as="x">
    <f:cycle values="{0: 'foo', 1: 'bar', 2: 'baz'}" as="cycle">{cycle}</f:cycle>
    <f:for each="{0: 'a', 1: 'b', 2: 'c'}" as="x2">
      <f:cycle values="{0: 'subfoo', 1: 'subbar', 2: 'subbaz'}" as="cycle">{cycle}</f:cycle>
    </f:for>
    <f:cycle values="{0: 'foo2', 1: 'bar2'}" as="cycle">{cycle}</f:cycle>
</f:for>
```

renders as

```
    foo
      subfoo
      subbar
      subbaz
    foo2

    bar
      subfoo
      subbar
      subbaz
    bar2

    baz
      subfoo
      subbar
      subbaz
    foo2
```

then please abandon this.
(Unfortunately I can't test this myself right now because I didn't get the latest version running with Flow yet).


PS: Every ViewHelper that adds something to the `ViewHelperVariableContainer` should *always* remove it again in order to prevent side-effects. That's not done in this change

mneuhaus:
i just confirmed, that in the current version of Fluid Standalone the example of @bwaidelich renders as: 

```
foo
	subfoo
	subbar
	subbaz
foo2
bar
	subfoo
	subbar		
	subbaz
bar2

baz
	subfoo
	subbar
	subbaz
foo2
```

so i'm closing this :)

PR #26, opened 2015-10-16
[TASK] Cleanup in ViewHelperNode

NamelessCoder:
* Avoids setting the rendering context (this gets done by the Invoker already; as is proper)
* Do not clone the "uninitialized" ViewHelper instance upon evaluating the rendering context (can *still* be done by Invoker if required/desired but is no longer default. Fixes: #5)
* Use `get_class` to read the obvious class name of already resolved ViewHelper, bypassing an already performed lookup in Resolver
* Change `getArgumentDefinitions` to read already retrieved argument definitions instead of consulting the Resolver (which would cache them, but still...)
* Changed order of logic in constructor to put simple assignments on top and rewriting logic on bottom

### DIFF
--- a/src/Core/Parser/SyntaxTree/ViewHelperNode.php
+++ b/src/Core/Parser/SyntaxTree/ViewHelperNode.php
@@ -60,9 +60,9 @@ class ViewHelperNode extends AbstractNode {
 	 */
 	public function __construct(ViewHelperResolver $resolver, $namespace, $identifier, array $arguments, ParsingState $state) {
 		$this->viewHelperResolver = $resolver;
-		$this->viewHelperClassName = $resolver->resolveViewHelperClassName($namespace, $identifier);
-		$this->uninitializedViewHelper = $resolver->createViewHelperInstance($namespace, $identifier);
 		$this->arguments = $arguments;
+		$this->viewHelperClassName = $resolver->resolveViewHelperClassName($namespace, $identifier);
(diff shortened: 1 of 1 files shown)


NamelessCoder:
Small note about the historical perspective:

In the former version of Fluid, the `setViewHelperNode` and `setChildNodes` methods would only be called on ViewHelpers which implemented a certain interface. As part of the general interface cleanup (which removed almost all ViewHelper interfaces, consolidating them into one) those interfaces no longer exist and the methods are simply called on *every* ViewHelper instance and is now implemented as NOOP-style methods in the base classes.

Those particular methods *are only called on uncompiled templates* because the entire Node system isn't used in compiled templates. This enables a new feature: compiling ViewHelpers based on for example the child nodes they contain. However, it also means that some basic internal stuff is changed:

* ViewHelpers are now compilable by default and the only way to forcibly disable such compiling is to call `$compiler->disable()` from the `compile()` method of a ViewHelper.
* The methods to set nodes are not called once the template is compiled, which means that obviously those methods should only be used if the compiler also gets disabled or custom compiling is implemented, handling those nodes during compiling so the compiled execution doesn't depend on them being present.

That's just to explain why there is *some* logic to rewrite/set nodes in the ViewHelperNode itself.

But please test this. There is some risk of breaking ViewHelpers that assume the RenderingContext is already present when receiving the child nodes - which given the explanation above should be happening in the `compile` and/or `render` method instead (which are non-static methods that can consume `$this->renderingContext`, `$this->viewHelperNode` and `$this->childNodes`). We should carefully consider the impact.

PR #27, opened 2015-10-16
[BUGFIX] Do not register Namespaces inside CDATA

### LINKED ISSUE #8: Throw exception for unresolved namespaces

mneuhaus:
closes #8

### DIFF
--- a/src/Core/Parser/Patterns.php
+++ b/src/Core/Parser/Patterns.php
@@ -106,6 +106,14 @@ abstract class Patterns {
 	static public $SCAN_PATTERN_CDATA = '/^<!\[CDATA\[(.*?)\]\]>$/s';
 
 	/**
+	 * This pattern detects any CDATA fields inside a string wihtout the limitation
+	 * that it has to be the start and end of that string. This is primarily
+	 * used by the TemplateParser to strip the TemplateSource from CDATA parts
+	 * to ignore namespace registrations inside a cdata tag
+	 */
+	static public $STRIP_PATTERN_CDATA = '/<!\[CDATA\[(.*?)\]\]>/s';
+
+	/**
 	 * Pattern which splits the shorthand syntax into different tokens. The
 	 * "shorthand syntax" is everything like {...}
 	 */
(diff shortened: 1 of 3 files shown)


NamelessCoder:
Commit message CGL (lowercase letter after prefix) - other than that, this looks good to me. If I'm not mistaken it is very much expected that *nothing whatsoever* inside CDATA gets processed so completely stripping it out seems like the right way.

mneuhaus:
Commit message updated :)

bwaidelich:
This doesn't solve the issue completely unfortunately..

```
some line
<![CDATA[{nonExistingNamespace:bar()}]]>
```

## Expected:

```
some line
{nonExistingNamespace:bar()}
```

## Actual:

```
Error: Unknown Namespace: nonExistingNamespace:bar() (0). Template code: 
</body>
</html>
```

bwaidelich:
..it might work to add the ``m`` modifier to ``$SCAN_PATTERN_CDATA`` - but I wonder why this new issue occurred suddenly, maybe we should investigate..

NamelessCoder:
> but I wonder why this new issue occurred suddenly, maybe we should investigate..

I don't think it did occur suddenly. It was silent until unregistered namespaces began throwing exception - and the decision to *not* do that lies quite a while back and has been discussed and undone since then ;)

NamelessCoder:
Replaced by #27

PR #28, opened 2015-10-18
[TASK] Use ENT_QUOTES rather than ENT_COMPAT

NamelessCoder:
Ensures escaping of single quotes. See: https://forge.typo3.org/issues/59540


NamelessCoder:
Test breaks for an unrelated reason (PHP7 decided it should add line numbers in a way that output buffering doesn't grab). Will be fixing that in a coming PR.

PR #36, opened 2015-11-15
[TASK] Improve TemplateParser logic

NamelessCoder:
This change:

* Improves handling of errors with template examples
* Moves such template code errors to TemplateParser entirely
* Implements validation of required arguments (by presence alone) in ViewHelperNode constructor, causing it to fail on bad syntax
* Introduces a secondary validateAdditionalArguments method that allows validating the presences of additional, unregistered arguments at parse time
* Cleans up the way ViewHelperNode is handled in TemplateParser
* Makes sure the TemplatePaths checks for global (not controller specific) templates after controller specific ones

### DIFF
--- a/src/Core/Parser/SyntaxTree/ViewHelperNode.php
+++ b/src/Core/Parser/SyntaxTree/ViewHelperNode.php
@@ -6,7 +6,9 @@ namespace TYPO3Fluid\Fluid\Core\Parser\SyntaxTree;
  * See LICENSE.txt that was shipped with this package.
  */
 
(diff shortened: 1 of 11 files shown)


NamelessCoder:
This one spiraled a bit - still needs to be updated for commit message and pull request description, but in short: changes the way the template parser and viewhelpernode reports errors and clean up the templateparser's handling of viewhelpernodes.

NamelessCoder:
And it looks like the PHP7 building is currently failing for unrelated reasons. The code is still PHP7-safe.

PR #46, opened 2015-12-30
[FEATURE] RenderingContext as implementation API

NamelessCoder:
#### Status: done
This change:

* Wraps all essential context-like classes into RenderingContext
* Removes those same context-like classes from various instances which now receive a RenderingContext

Using the RenderingContext is ideal since it is available both during parsing and once templates are compiled, and can be replaced at any time without invalidating compiled templates.

The change breaks code which depends on the following usages:

* getViewHelperInvoker() on ViewHelperResolver (moved to RenderingContext)
* getExpressionNodeTypes on ViewHelperResolver (moved to RenderingContext)
* resolveViewHelperInvoker() by ViewHelper class name (moved to RenderingContext, no class name gets passed)
* explicit calls to injectViewHelperVariableContainer on RenderingContext (renamed to setViewHelperVariableContainer for consistency in interface)

### DIFF
--- a/examples/example_conditions.php
+++ b/examples/example_conditions.php
@@ -23,7 +23,7 @@ $view->assign('ternaryFalse', 'The ternary expression is FALSE');
 
 // Assigning the template path and filename to be rendered. Doing this overrides
 // resolving normally done by the TemplatePaths and directly renders this file.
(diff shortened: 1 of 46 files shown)


mneuhaus:
cleaned up various places quite nicely :)

PR #53, opened 2016-01-10
[FEATURE] Allow f:alias to create VariableProviders using "src" argument

NamelessCoder:
This adds three new arguments to f:alias:

* `src` which can contain data which can be read by a VariableProvider
* `name` which can contain the name of a "container" variable in which values from the Provider will be placed
* `type` which allows manually defining the type

Currently, the one VariableProvider type which makes sense to use here is the `JsonVariableProvider`.

Examples:

```xml
<!-- given "file.json" contains {"name": "John", "phone": "1-800-FLUID"} -->
<f:alias src="/path/to/file.json" as="json">
  {json.name} at {json.phone}
  <!-- outputs: "John at 1-800-FLUID" -->
</f:alias>
```

The "src" argument can any stream wrapper compatible URI.

In addition the "JSONVariableProvider" has been renamed to "JsonVariableProvider" to be able to correctly assume the name of this and future VariableProviders to be supported.

### DIFF
--- a/src/Core/Variables/JSONVariableProvider.php
+++ b/src/Core/Variables/JSONVariableProvider.php
@@ -7,12 +7,12 @@ namespace TYPO3Fluid\Fluid\Core\Variables;
  */
 
 /**
- * Class JSONVariableProvider
+ * Class JsonVariableProvider
  *
  * VariableProvider capable of using JSON files
  * and streams as data source.
  */
-class JSONVariableProvider extends StandardVariableProvider implements VariableProviderInterface {
+class JsonVariableProvider extends StandardVariableProvider implements VariableProviderInterface {
 
 	/**
 	 * @var integer
--- a/src/ViewHelpers/AliasViewHelper.php
+++ b/src/ViewHelpers/AliasViewHelper.php
@@ -6,14 +6,22 @@ namespace TYPO3Fluid\Fluid\ViewHelpers;
  * See LICENSE.txt that was shipped with this package.
  */
 
+use TYPO3Fluid\Fluid\Core\Parser\Exception as ParserException;
+use TYPO3Fluid\Fluid\Core\ViewHelper\Exception as ViewHelperException;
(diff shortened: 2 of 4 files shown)


mneuhaus:
+1

NamelessCoder:
+1

mneuhaus:
+1

NamelessCoder:
+1

mneuhaus:
+1

mneuhaus:
+1

mneuhaus:
looks good

mneuhaus:
:+1:

mneuhaus:
the more i think about this, the less i think of this as "core" feature :/

NamelessCoder:
Leave it open for a while - we don't need/want to make a decision about this yet since it's a feature.

NamelessCoder:
Would any sub-set of this be more acceptable? It could be seen as three adaptations:

* The support for VariableProvider as `src` in general (easily achieved if we allow instances, little less easy if we allow passing a class name which I think we should allow)
* Support for any variable type as `src` combined with `name` argument to assign a single variable (as a key/value property based substitute for defining an array to alias map which can be a bit hard to use)
* Support for `type` which would attempt to cast the array (or key/value pair if second adaptation is applied) or use it as constructor argument for the class passed as `type` (e.g. ArrayIterator, SeekableIterator and whatnot)

I think any one of them could make sense but perhaps you are both right that it is a bit much when combined like this... let me know what you think! (also if you're a user reading this and have an opinion on the matter)

mneuhaus:
Still leaning towards not putting that into fluid core. It could easily be put into a separate package with a clearer viewHelper name like "externalData", "data", etc. 
This feature would "tempt" users to "sideload" data into fluid templates instead of the normal view->assign way. The idea is neat, but i can't think of a common usecase where this would be needed and couldn't be solved by view->assign or a custom VariableProvider

NamelessCoder:
Discarding this idea - such VariableProviders are indeed best created externally and assigned.

PR #69, opened 2016-01-29
[BUGFIX] Support bin/fluid execution as dependency

mbrodala:
If typo3fluid/fluid is installed as dependency the bin/fluid script fails with an error:

> PHP Fatal error:  Class 'TYPO3Fluid\Fluid\Core\Variables\StandardVariableProvider' not found in
> /var/www/vendor/typo3fluid/fluid/examples/include/class_customvariableprovider.php on line 19

Fix this by also looking for autoload.php in the vendor-dir this package is currently
installed to.


NamelessCoder:
Thanks!

mbrodala:
:-)

PR #81, opened 2016-03-05
[BUGFIX] Ensure correct line endings in Unit Tests on Windows systems

IchHabRecht:
This patch ensures correct line endings for Unit Test when running on Windows systems (and checked out as Unix files).


mneuhaus:
do i understand this correctly, PHP_EOL contains "incorrect" data on windows systems?

IchHabRecht:
On windws PHP_EOL == "\n\r"

PR #87, opened 2016-03-11
[BUGFIX] Avoid creating invalid class names for compiled template sources

NamelessCoder:
When passing a template source (not file reference) the resulting compiled class name was potentially invalid due to using sha1() as prefix. Adding a `source_` prefix avoids the problem.


helhum:
works, thanks!

PR #88, opened 2016-03-11
[BUGFIX] Ensure sanitized paths on Windows

IchHabRecht:
This patch ensures sanitized paths when using realpath(). Some
improvements for the tests are added as well.


IchHabRecht:
Sorry for not thinking about the tests before.

PR #89, opened 2016-03-14
[TASK] Increase code coverage percentage

NamelessCoder:
This change increases the percentage of covered code by adding additional test cases and refactoring the recursive array handling method in TemplateParser. No behaviors changed.


NamelessCoder:
Nothing different except depending on return value from `preg_match_all` to enter condition rather than checking matches array and early-returning.

mneuhaus:
-0.5 for turning away from "return early" but i'll let it pass in this instance ;P

PR #93, opened 2016-03-15
[BUGFIX] Assign RenderingContext during parsing

NamelessCoder:
Some third party ViewHelpers may depend on the presence of a rendering context while parsing - although it technically is an incorrect expectation to have the context available at this point, making it available allows third parties to access various secondary aspects of the rendering context.

Specifically, the lack of this parse-time rendering context setting affects Widgets used in TYPO3 CMS when the template is uncompilable.

Followup for https://github.com/TYPO3Fluid/Fluid/pull/77 which removed this line as it was evidently redundant. Line restored now along with a comment linking to this pull request.

### DIFF
--- a/src/Core/Parser/SyntaxTree/ViewHelperNode.php
+++ b/src/Core/Parser/SyntaxTree/ViewHelperNode.php
@@ -57,6 +57,8 @@ class ViewHelperNode extends AbstractNode {
 		$this->viewHelperClassName = $resolver->resolveViewHelperClassName($namespace, $identifier);
 		$this->uninitializedViewHelper = $resolver->createViewHelperInstanceFromClassName($this->viewHelperClassName);
 		$this->uninitializedViewHelper->setViewHelperNode($this);
+		// Note: RenderingContext required here though replaced later. See https://github.com/TYPO3Fluid/Fluid/pull/93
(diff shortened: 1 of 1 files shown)


kitsunet:
Just stumbling here from the comment in the code. Obviously Flow widgets need the  same I guess. But how do I create a VH node from an interceptor (which doesn't have access to the context currently)?

NamelessCoder:
Hi Christian,

This will have to be the very short version and I apologise for that (post-moving chaos). Historical reason for dropping the context from interception areas was that after changing the escaping behavior so it no longer creates VH nodes (wraps with an escapingnode instead) the interception area no longer had any immediate need for accessing the context. And so it was dropped, saving many setter function calls (more, the bigger the template gets).

However, since every interceptor is now created via the rendering context you could in theory prepare every interceptor you provide, with a link back to the rendering context instance (which would be a reference to the original instance with everything that this implies; caveat in particular when the context gets cloned for sub-request rendering, e.g. widgets). With just the buildParserConfiguration method override you *should* be able to get there - if necessary, look into somehow reacting to when the rendering context gets cloned (to re-assign that cloned instance in interceptors the cloned context returns from buildParserConfiguration).

Hope that makes sense - if you need more info on this I'm happy to discuss.

PR #94, opened 2016-03-15
[BUGFIX] Avoid double HTML encoding on chained view helpers

ohader:
If view helpers are chained, already encoded HTML entities will
be encoded again and again. To circumvent that, the handing in
the escaping interceptor is adjusted to avoid double encoding.

Related: https://forge.typo3.org/issues/75133
Releases: master

### DIFF
--- a/src/Core/Parser/SyntaxTree/EscapingNode.php
+++ b/src/Core/Parser/SyntaxTree/EscapingNode.php
@@ -38,7 +38,7 @@ class EscapingNode extends AbstractNode {
 	 * @return number the value stored in this node/subtree.
 	 */
 	public function evaluate(RenderingContextInterface $renderingContext) {
-		return htmlspecialchars($this->node->evaluate($renderingContext), ENT_QUOTES);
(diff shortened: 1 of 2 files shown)


mneuhaus:
good catch! :+1:

helhum:
Imho this fix is wrong. 

Given
* a myVh has escapeChildren **and** escapeOutput set to true
* template variable foo is `<h1>foo</h1>`
* a template  `<ns:myVh>{foo}</ns:myVh>`

I expect the rendered result to be `&amp;gt;h1&amp;lt;foo&amp;gt;/h1&amp;lt;`

helhum:
Instead, we need to change the CropViewHelper (in TYPO3 CMS adapter) to disable children escaping and allow only output escaping

See flow fluid, where this is done correctly:

https://github.com/neos/fluid/blob/master/Classes/TYPO3/Fluid/ViewHelpers/Format/CropViewHelper.php#L55

ohader:
You're right that this has to be done in the view helpers... otherwise double encoding on purpose won't work anymore...

see https://github.com/TYPO3Fluid/Fluid/pull/98

PR #98, opened 2016-04-11
Revert "[BUGFIX] Avoid double HTML encoding on chained view helpers"

ohader:
This reverts commit d2e36f5ff2a9d3101130c66302695f207e5d3a50.

With the previous change it is not possible anymore to enforce double
encoding. Thus, the behavior has to be changed using the class member
variables escapeChildren and escapeOutput instead (which has been
done for the view helpers of the core already).

Related: https://forge.typo3.org/issues/75133


helhum:
Thanks!

PR #100, opened 2016-04-18
[FEATURE] Re-Introduce escaping modifiers functionality

helhum:


NamelessCoder:
Hi Helmut,

This wasn't explained during the decoupling process but it is actually on purpose that this feature was removed. The intention was that if a framework desires to allow users to disable escaping for an entire file like this, then that should be done using a TemplatePreProcessor (native to the framework implementing Fluid).

At the moment I don't agree that this should be a feature of standalone Fluid. I view it much like the legacy namespace extraction logic and think it should be implemented the same way.

helhum:
Hi!

We discussed this on Slack. Here is the result:

Since Fluid has support for different formats of templates (xml, html, json …), and escaping is enabled by default for **all** formats currently, one would need to wrap **any** view helper in `<f:format.raw>` when using a format, where HTML escaping is not desired.

To mitigate this, we add a way to enable/disable the escaping interception for templates.

Additionally there might be a change, where we would disable HTML escaping by default for other formats than HTML and XML

helhum:
@NamelessCoder any news here?

PR #101, opened 2016-04-20
[TASK] Simplifiy encoding behavior

helhum:
We now disable children escaping in case output escaping is already on.
This is a saner default so that view helper creators very seldomly need to
think about setting any of the escaping behavior properties.


helhum:
ping

mneuhaus:
pong :)

helhum:
Thanks!

PR #104, opened 2016-05-23
[BUGFIX] Respect template paths priority

froemken:


NamelessCoder:
Hi Stefan - the intention is that you fill TemplatePaths with already sorted paths. We definitely do not want to have the internal property become sorted efter time you *get* the template paths. Fluid itself should not do any sorting - all it should do is make sure it starts from the bottom of paths when detecting files.

helhum:
I agree with @NamelessCoder here. We should close this pr

NamelessCoder:
@froemken If this issue comes from TYPO3 do you have a core issue for that? We should be able to handle this in the CMS adapter, if for example it is a problem when setting paths from TS.

froemken:
Here is the Forge-Ticket: https://forge.typo3.org/issues/75862

froemken:
Quote: "all it should do is make sure it starts from the bottom of paths when detecting files". Just two lines above you told me: "the intention is that you fill TemplatePaths with already sorted paths". Funny: So, if TYPO3 calls Fluid with already sorted paths AND you/Fluid starts detecting files from the bottom we will have the some problem again.

helhum:
@froemken: don't get it, sorry. This thing needs to be fixed on TYPO3 side

NamelessCoder:
If you put those two quotes together it should say that you should provide TemplatePaths with sorted paths where the last path is the one you want checked first - and keys do not matter to Fluid.

NamelessCoder:
I'll make myself a cup of coffee and see if I can track this down in the CMS adapter!

helhum:
@NamelessCoder I have a fix almost ready. Can you clarify for me which path is preferred?

Given I have default templates in `/base` which should be overridden by templates in `/regular` which should be overridden by templates in `/override`

Is this ordering correct?

```
Array
(
    [0] => /override
    [1] => /regular
    [2] => /base
)
```

NamelessCoder:
Oh darn, me too! :)

In the described use case the order should be opposite. Fluid itself will receive arrays in the human logical order where you define base first, then override A, then override B etc. but then internally loops it in opposite order to detect in override B, then override A, then base.

My fix: in FLUIDTEMPLATE contentObject I am `ksort`ing the paths before passing them to the View. Controllers should already sort the paths correctly.

NamelessCoder:
I'll submit my suggested fix so we can discuss it in Gerrit - should be there in a minute.

NamelessCoder:
My suggested patch - https://review.typo3.org/#/c/48279/

Alternative strategy could be fill TemplatePaths with the full `$conf` but since we need stdWrap, I think this is the most transparent approach.

froemken:
@helhum My intention was to prevent you from fixing this bug in both system, which will result in the same problem again.

NamelessCoder:
@froemken Perhaps the confusion comes from the difference that TYPO3Fluid does not and should not care about the keys - it only cares about the order of the array you give TemplatePaths; and lets you create custom TemplatePaths if you wish them to be resolved differently. Whereas TYPO3 CMS cares about the keys because of the way you define TypoScript (not necessarily in order so keys as priority is required).

This is why the right solution is to solve this one in TYPO3 CMS without affecting how TYPO3Fluid itself behaves. The problem is, in other words, that TYPO3 CMS neglects to sort these paths in the use case you describe, and so TYPO3Fluid receives paths in the wrong order compared to what's required by the API.

But don't stop reporting/submitting pull requests! Some cases, like this one, is not immediately logical without a lot of knowledge about the decisions made when decoupling Fluid - and as you know I'm always happy to explain those ;)

Will close this one and continue on Gerrit. Interested to see if @helhum came up with the same solution as me :)

NamelessCoder:
Something is fishy... the linked patch for TYPO3 is still needed, but it looks like I mixed up TemplatePaths of TYPO3 and TemplatePaths of TYPO3Fluid in the description.

Situation now:

* TemplatePaths of TYPO3Fluid *does* indeed care about the array key but only when passing it an array that contains the `templateRootPaths` etc. keys.
* TemplatePaths of TYPO3Fluid iterates template paths top->bottom which means they do get processed in the order @helhum mentions, but works because of the krsort when receiving an array with paths.
* Example of template overrides in TYPO3Fluid works by coincidence (automatic index keys, array reversed, iterated top->bottom results in same order)

Intended situation:

* TemplatePaths of TYPO3Fluid *does not* care about array keys and *does not* sort the array.
* TemplatePaths of TYPO3Fluid *must* change iteration order bottom->top and avoid the krsort.
* TemplatePaths of TYPO3 CMS should implement this forced sorting - in any or all of the extractPathArrays, setTemplateRootPaths, setPartialRootPaths or setLayoutRootPaths methods.

It's pretty clear to see why you'd get confused. I even managed to confuse myself about this one :/ now to work on fixing this. Sorry about all the back and forth - my bad!

helhum:
Thanks @NamelessCoder plan sounds good.

PR #105, opened 2016-05-24
[BUGFIX] Do not sort by key when receiving paths arrays

NamelessCoder:
During the decoupling a decision was made to not use
array keys for the priority of template paths - instead,
the order of the array should determine the lookup order.

However, the TemplatePaths class of TYPO3Fluid had
received a sorting behavior that was intended for the
TemplatePaths class of TYPO3 CMS. This patch removes
that sorting behavior.

This fix must be followed up by a patch for TYPO3 CMS.
suggested strategy there is to override the setter methods
for paths and enforce the sorting there.

References: #104
References: https://forge.typo3.org/issues/75862
References: https://review.typo3.org/#/c/48279/

### DIFF
--- a/src/View/TemplatePaths.php
+++ b/src/View/TemplatePaths.php
@@ -241,17 +241,13 @@ class TemplatePaths {
 		$action = ucfirst($action);
 		$identifier = $controller . '/' . $action . '.' . $format;
 		if (!array_key_exists($identifier, self::$resolvedFiles['templates'])) {
-			foreach ($this->getTemplateRootPaths() as $templateRootPath) {
-				$candidate = $this->ensureAbsolutePath($templateRootPath . $action . '.' . $format);
-				if (file_exists($candidate)) {
-					return self::$resolvedFiles['templates'][$identifier] = $candidate;
-				}
-			}
-			foreach ($this->getTemplateRootPaths() as $templateRootPath) {
-				$candidate = $this->ensureAbsolutePath($templateRootPath . $identifier);
-				if (file_exists($candidate)) {
-					return self::$resolvedFiles['templates'][$identifier] = $candidate;
-				}
+			$templateRootPaths = $this->getTemplateRootPaths();
+			try {
+				return self::$resolvedFiles['templates'][$identifier] = $this->resolveFileInPaths($templateRootPaths, $controller . '/' . $action, $format);
+			} catch (InvalidTemplateResourceException $error) {
+				return self::$resolvedFiles['templates'][$identifier] = $this->resolveFileInPaths($templateRootPaths, $action, $format);
+			} catch (InvalidTemplateResourceException $error) {
+				return self::$resolvedFiles['templates'][$identifier] = NULL;
 			}
 		}
 		return isset(self::$resolvedFiles['templates'][$identifier]) ? self::$resolvedFiles['templates'][$identifier] : NULL;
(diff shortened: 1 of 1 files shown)


helhum:
no test change needed for this? 😧

src/View/TemplatePaths.php
```diff
@@ -666,7 +655,7 @@ protected function getPartialPathAndFilename($partialName) {
 	 */
 	protected function resolveFileInPaths(array $paths, $relativePathAndFilename, $format = self::DEFAULT_FORMAT) {
 		$tried = array();
-		foreach ($paths as $path) {
+		while ($path = array_pop($paths)) {
```
helhum: what is the point of this change?

src/View/TemplatePaths.php
```diff
@@ -241,17 +241,11 @@ public function resolveTemplateFileForControllerAndActionAndFormat($controller,
...
-				$candidate = $this->ensureAbsolutePath($templateRootPath . $identifier);
-				if (file_exists($candidate)) {
-					return self::$resolvedFiles['templates'][$identifier] = $candidate;
-				}
+			$templateRootPaths = $this->getTemplateRootPaths();
+			try {
+				return self::$resolvedFiles['templates'][$identifier] = $this->resolveFileInPaths($templateRootPaths, $controller . '/' . $action, $format);
+			} catch (InvalidTemplateResourceException $error) {
+				return self::$resolvedFiles['templates'][$identifier] = $this->resolveFileInPaths($templateRootPaths, $action, $format);
 			}
```
helhum: I do not see a connection between nor sorting paths and this change. What is this change good for?

src/View/TemplatePaths.php
NamelessCoder: Reversing the order of iteration

src/View/TemplatePaths.php
NamelessCoder: Refactoring to call existing method to resolve file - that method has now been fixed to reverse the iteration order (see above)

NamelessCoder:
No test change required. Test cases were already providing and expecting the right orders (since this krsort behavior was never intended to be in Fluid itself)

src/View/TemplatePaths.php
helhum: ah, of course. looks a bit non obvious, but I could live with that

NamelessCoder: You're right - I'll put this as a code comment for anyone reading there and not reading this pull request.

src/View/TemplatePaths.php
```diff
@@ -241,17 +241,11 @@ public function resolveTemplateFileForControllerAndActionAndFormat($controller,
...
-			foreach ($this->getTemplateRootPaths() as $templateRootPath) {
-				$candidate = $this->ensureAbsolutePath($templateRootPath . $identifier);
-				if (file_exists($candidate)) {
-					return self::$resolvedFiles['templates'][$identifier] = $candidate;
-				}
+			$templateRootPaths = $this->getTemplateRootPaths();
+			try {
+				return self::$resolvedFiles['templates'][$identifier] = $this->resolveFileInPaths($templateRootPaths, $controller . '/' . $action, $format);
+			} catch (InvalidTemplateResourceException $error) {
+				return self::$resolvedFiles['templates'][$identifier] = $this->resolveFileInPaths($templateRootPaths, $action, $format);
```
helhum: not sure, but don't we potentially then throw an exception here which was not thrown before?

NamelessCoder: Ah, yes and no... `getTemplateSource` will throw the same exception if this method returns NULL or the file does not exist - but the exception thrown from that method is a LOT more descriptive. Will refactor!

PR #107, opened 2016-05-31
[BUGFIX] Support dynamic nodes in Layout name attribute

NamelessCoder:
This enables use of any dynamic expressions, ViewHelpers
or variable references as part of or only value of the
"name" attribute of `f:layout` - see included example
`example_dynamiclayout.php`.

Marked as BUGFIX since this support was advertised
and intended to be included right from the start.

See also: https://forge.typo3.org/issues/70285
See also: https://forge.typo3.org/issues/76314


mneuhaus:
looks ok to me 👍

PR #110, opened 2016-06-18
[TASK] Clean up resolved file caching

NamelessCoder:
This change adds an internal ability to clean resolved filenames and identifiers by type of cache and uses this to flush only the necessary type from each of the set***Paths methods.

Also cleans up how flushing happens: now done for all types when fresh instances are created and done for specific types when individual template type paths are set.

Implements constants for the cache type names.


NamelessCoder:
NB: if/when this one gets merged then I suggest a follow-up refactoring which makes the internal resolving cache a non-static property and adjusts code accordingly (but preserves the flushing mechanics, of course).

PR #111, opened 2016-06-21
[BUGFIX] Resolve Layout correctly if layout name is a TextNode

NamelessCoder:
Resolves an issue where once compiled, if a template used a standard TextNode as name of Layout, the compiled template would fail to render with an exception about failing to resolve a Layout filename due to looking for an unexpected file (missing basename).


NamelessCoder:
Replaced by https://github.com/TYPO3Fluid/Fluid/pull/124 which causes casting of TextNodes to string which returns value from TextNode->getText(), in addition to supporting standard string values.

PR #112, opened 2016-06-21
[FEATURE] Improve the Debug ViewHelper's capabilities

NamelessCoder:
This change adds the following capabilities to `f:debug`:

* Option to produce HTML output (off by default)
* Option to limit recursion levels (5 by default)
* Ability to render ArrayObject, Iterator as key/value pairs
* Ability to render recursively
* Ability to render stream metadata

### DIFF
--- a/src/ViewHelpers/DebugViewHelper.php
+++ b/src/ViewHelpers/DebugViewHelper.php
@@ -6,19 +6,11 @@ namespace TYPO3Fluid\Fluid\ViewHelpers;
  * See LICENSE.txt that was shipped with this package.
  */
 
+use TYPO3Fluid\Fluid\Core\Variables\VariableExtractor;
 use TYPO3Fluid\Fluid\Core\ViewHelper\AbstractViewHelper;
 
 /**
- * View helper that outputs its child nodes with \TYPO3Fluid\Flow\var_dump()
  *
- * = Examples =
- *
- * <code>
- * <f:debug>{object}</f:debug>
- * </code>
- * <output>
- * all properties of {object} nicely highlighted
- * </output>
  *
  * <code title="inline notation and custom title">
  * {object -> f:debug(title: 'Custom title')}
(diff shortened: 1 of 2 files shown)


src/ViewHelpers/DebugViewHelper.php
```diff
@@ -70,6 +64,83 @@ public function render() {
...
+							str_repeat('  ', $level),
+							$property,
+							static::dumpVariable($value, $html, $level + 1, $levels)
+						);
+					}
+				}
+			}
+		} else {
+			if (is_scalar($variable)) {
+				$string = sprintf('<code>%s = %s</code>', $typeLabel, var_export($variable));
```
helhum: Everything here that it put into HTML must be HTML encoded with htmlspecialchars!

src/ViewHelpers/DebugViewHelper.php
```diff
@@ -70,6 +64,83 @@ public function render() {
 			return (is_object($expressionToExamine) ? get_class($expressionToExamine) : gettype($expressionToExamine));
 		}
 
-		return var_export($expressionToExamine, TRUE);
+		$html = $this->arguments['html'];
```
helhum: This is really tricky here. We either would need to disable output escaping (we encode for HTML ourselves) or let it enabled, so that the interceptor can do so where required. However having both in one view helper is impossible ind imho not optimal design.

Output is either too much encoded or to less

NamelessCoder: It should be possible to avoid double encoding HTML entities - you are of course correct about the escaping concerns, I'll update the patch.

NamelessCoder:
Added escaping of special chars in scalar values when in HTML mode (and prevented double escaping), implemented reflection to read property names so VariableExtractor can read values using getters etc. - and added some unit test cases to cover the newly added capabilities.

jkphl:
@NamelessCoder I tested this and found a couple of problems with the current implementation (list might not be complete):

1. `getValuesOfNonScalarVariable()` fails with `NULL` values (which aren't scalar, obviously)
2. `\DateTimeInterface` objects don't yield a reasonable result. The output is something like

  ```
  2016-06-21 00:40:35.000000: NULL
  1: NULL
  +02:00: NULL
  ```
3. Casting an `ArrayObject` to an array by using `(array)` seems to bypass the object's `getArrayCopy()` method and simply return the array or object `input` of the `ArrayObject`. So if you're doing custom things on property retrieval (which I do), these won't survive the `(array)`. Calling `getArrayCopy()` instead would be more suitable I think.

As I said, there might be other glitches, but I can't effectively test further until the aforementioned ones are solved.

NamelessCoder:
Thanks for the feedback Joschi - I'll be updating the patch with the additions you mention and any others. If you think of other objects or implementations that aren't supported, keep 'em coming :)

PR #113, opened 2016-06-25
[FEATURE] Provide dedicated ViewHelperCompiler and Traits

NamelessCoder:
This change adds a dedicated ViewHelperCompiler to cover frequent use cases in ViewHelpers:

* Compiling directly with static function call (to renderStatic - or other methods on other classes)
* Compiling with static function call and using one of the provided arguments as "content" or calling renderChildren if that argument is not passed.

Both use cases support calling a static method on *any* class - not just ViewHelpers, or on another ViewHelper - if that method conforms to the signature of the renderStatic method defined in ViewHelperInterface.

Both use cases are then implemented in ViewHelpers of Fluid where possible, which improves performance of the following ViewHelpers:

* f:format.cdata
* f:format.printf
* f:alias
* f:comment
* f:count
* f:for
* f:groupedFor
* f:or
* f:spaceless

Other ViewHelpers are, with one exception (f:cycle) already compilable in the most efficient way. The cycle ViewHelper still requires a custom compile method that will probably be so adapted it cannot be made generic enough to merit putting it in a Trait and the ViewHelperCompiler.

### DIFF
--- a/src/Core/Compiler/ViewHelperCompiler.php
+++ b/src/Core/Compiler/ViewHelperCompiler.php
@@ -1,0 +1,144 @@
+<?php
+namespace TYPO3Fluid\Fluid\Core\Compiler;
+
+use TYPO3Fluid\Fluid\Core\ViewHelper\ViewHelperInterface;
+
+/**
+ * Class ViewHelperCompiler
+ *
+ * Assistant class designed to exclusively compile ViewHelpers
+ * of common types. Each ViewHelper that wants to compile itself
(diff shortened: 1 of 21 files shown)


NamelessCoder:
NB! This is WIP but only because it contains https://github.com/TYPO3Fluid/Fluid/pull/112 until that one gets merged; the new compiling logic itself should be ready to go.

NamelessCoder:
Ready for your consideration!

PR #117, opened 2016-07-12
[TASK] Ensure testing the compiled/ cached templates

helhum:


helhum:
This PR is to demonstrate how hard things are broken with active caches

PR #118, opened 2016-07-12
Revert "[BUGFIX] Support dynamic nodes in Layout name attribute (#107)"

helhum:
This reverts commit 1cc3600097b74420c3fc857977b0af6bfbeafed1.
Reverted for now, as it completely breaks compiled templates.


helhum:
Ok, so obviously the revert does not fix everything and there is still one remaining broken case

NamelessCoder:
This one is demanded by TYPO3 CMS so we cannot revert. See https://github.com/TYPO3Fluid/Fluid/pull/124 for a fix that solves the issues you encounter with layouts in compiled templates.

Issue #127, opened 2016-07-13
Remove reserved variable names except "_all"?


NamelessCoder:
This is an issue to discuss whether or not we should remove the `on`, `yes`, `true`, `off`, `no` and `false` variable names (case insensitive!) which are currently reserved.

### Background

In earlier versions of Fluid, using a reserved variable name would trigger an error. In Fluid standalone however, such variables are given a default value so for example using `{on}` in a template will result in a boolean `TRUE` being returned from the ObjectAccessorNode.

This feature is as of this time undocumented but considered a hardcoded feature of Fluid.

### Improvement

Dropping the reserved variable names would result in two things:

1. Performance of every ObjectAccessorNode can be improved by avoiding the need to check if a variable is reserved (which happens for *every* variable you try to access).
2. Reserved variable names can now be used in the template (previously, you could declare but not access the declared value - it would be hidden away due to the name).

The method could then be compressed to a single decision: is the variable name `_all`; if so, return `->getAll()` from VariableProvider. Furthermore the special `_all` variable name could be made case sensitive and only the lowercase variant trigger getting all - which would save an additional `strtolower` call for every variable accessed.

See https://github.com/TYPO3Fluid/Fluid/blob/master/src/Core/Parser/SyntaxTree/ObjectAccessorNode.php#L81

mneuhaus:
To be honest, i didn't even know those existed. Seems "counter intuitive" so i'd say lets scrap them.
But only if it's not to widely used, has anyone an idea if/how much that's used?

NamelessCoder:
> But only if it's not to widely used, has anyone an idea if/how much that's used?

I have a guesstimate - before Fluid went standalone, these variables were impossible to define and accessing them would result in NULL. It is only after standalone that they have constant values and the fact is undocumented, so I would guess that approximately zero people currently use it ;)

PR #128, opened 2016-07-13
[TASK] Micro-optimisation in ObjectAccessorNode->evaluate

NamelessCoder:
The method is refactored to clearly declare as class property the list of reserved variable names and their values. The switch is (although nicely performant) replaced by two if-clauses, which avoid calling ->getVariableProvider on RenderingContext if retrieving default constants.

See also: https://github.com/TYPO3Fluid/Fluid/issues/127

### DIFF
--- a/src/Core/Parser/SyntaxTree/ObjectAccessorNode.php
+++ b/src/Core/Parser/SyntaxTree/ObjectAccessorNode.php
@@ -28,6 +28,20 @@ class ObjectAccessorNode extends AbstractNode {
 	 */
 	protected $accessors = array();
 
+    /**
+     * Reserved variables which have default, static values
+     *
+     * @var array
+     */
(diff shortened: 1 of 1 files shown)


helhum:
Any numbers how much time is spent on that?

Other ideas: isn't it possible to compile the static values as static values in the PHP cache?

other than that, I have no objections

NamelessCoder:
> Any numbers how much time is spent on that?

The impact without also doing #127 is near nothing (both switch and array are compiled to hash maps which would perform equally well). Basically, it will only improve performance in the case where you are actually using these hardcoded constants (and there are, as you also touch on, much better ways to support such things if wanted!).

With #127 in mind we can view this one as a preparation that makes it easier to drop or refactor these constant/reserved variables in a coming PR when we agree how to solve it.

Issue #129, opened 2016-07-15
"-1 != -1" wrongly converted to boolean.


kitsunet:
The `BooleanParser` seems to struggle with the mentioned expression.

Running

```
<f:if condition="-1 != -1">
  <f:then>
    Then child!
  </f:then>
  <f:else>
    Else child!
  </f:else>
</f:if>
```

Will result in `Then child!` and the compiled code looks like this:

```
$array2 = array();
$array2['0'] = '-1 != -1';

$expression3 = function($context) {return "-";};
$arguments1['condition'] = TYPO3Fluid\Fluid\Core\Parser\SyntaxTree\BooleanNode::convertToBoolean(
					$expression3(
						TYPO3Fluid\Fluid\Core\Parser\SyntaxTree\BooleanNode::gatherContext($renderingContext, $array2)
					),
					$renderingContext
				);
```

From what I see the BooleanParser does not take `-` into account. I will try to find a fix for it.

NamelessCoder:
If he'll forgive the notification, let's mention @mneuhaus here - he is the resident expert on the BooleanParser and maybe has a hint that can save you some time :)

mneuhaus:
lookin into it after lunch :)

mneuhaus:
Pullrequest waiting for review :)

kitsunet:
Don't know the policy, but this could be closed now?

mneuhaus:
sure it can :)

PR #130, opened 2016-07-15
[BUGFIX] Fix bug with negative numbers in BooleanParser

mneuhaus:
Comparisons with negative Numbers got parsed falsely because the - was missing from the
tokenregex.

related issue: #129


kitsunet:
Wonderful, thanks! Code review 👍

kitsunet:
And just verified the solution :+1:

PR #131, opened 2016-07-30
[BUGFIX] Detect/allow number-prefixed variables in array parts

NamelessCoder:
This change fixes an issue where a template variable named for example "123variable" would render correctly when referenced as `{123variable}` but if used in an array part like for example `<f:alias map="{newVar: 123variable}">` the variable would not be resolved. Instead, a numeric value of `123` would be cast by the parser. Loosening up the expression which required array part variable references to start with a letter solves the issue.

See also: https://forge.typo3.org/issues/76276


NamelessCoder:
NB: not considered breaking since it only "breaks" cases where a template author intentionally writes numbers with letter suffixes as array parts and (incorrectly) expects a number as a result (rather than NULL which is what such a case results in now, if no such named variable exists).

PR #132, opened 2016-07-31
[FEATURE] Remove all reserved variables except _all

NamelessCoder:
Close: #127


NamelessCoder:
Reworded to a feature since it changes behavior to allow new usages. Also, should not be backported to the 1.0 branch.

Issue #135, opened 2016-08-30
resolveFileInPaths stops after empty path


vertexvaar:
With 61b1fb95d69b4255ef1bcf7cce0306deecfbc1ba introduces a bug where resolving possible file paths will be stopped when a path is considered empty
https://github.com/TYPO3Fluid/Fluid/blob/master/src/View/TemplatePaths.php#L662 does not explicitly check for check for `=== null`, which is returned by array_pop at the end of an array.
So if the first pathin templateRootPaths is an empty string, no path will be checked and an exception will be thrown.
I'll provide a PR shortly.

PR #136, opened 2016-08-30
[BUGFIX] Explicitly check for end of array (#135)

vertexvaar:
Explicitly check for null when testing for the end of the paths
array when searching for file candidates. This fixes an issue with empty
strings as root path, which would immediately stop the iteration and
trigger an exception.

Resolves: #135


NamelessCoder:
Thanks for a great description! I took the liberty of adjusting the commit message slightly so it is more clear what the intention is.

PR #138, opened 2016-09-01
[BUGFIX] Support cached OrViewHelper

sirdiego:
When the OrViewHelper gets cached, the $argumentsX variable
gets filled with NULL and the ViewHelperInvoker tests this
with isset(). If this isset() returns false, the Invoker then
fills the argument with its default value. In this case an empty
string is filled and not NULL, like expected.

You can replicate this by using the OrViewHelper once without cache
and then with cache. The cached version will return an empty string
instead of the provided alternative.


NamelessCoder:
Thanks Diego - this and a couple of other ViewHelpers could use a round of cleaning!

PR #139, opened 2016-09-03
Require minimum pcre lib 8.36

ksjogo:


NamelessCoder:
Recreated as https://github.com/TYPO3/Fluid/pull/208 so the build will assign our required status checks!

PR #143, opened 2016-09-15
Allow upper camel case letters as namespace identifier

froemken:
In earlier versions it was allowed to use lower and upper case letters for namespace identifier. Further upper case letters are still valid via your Pattern: $SCAN_PATTERN_TEMPLATE_VIEWHELPERTAG

So would be cool to see allowed upper case letters in namespace identifiers again in your code.

Stefan


NamelessCoder:
Meta note: if someone merges this then *do not forget to reword the commit subject!*. @froemken you don't have to do this as it's a very tiny effort for us, but it helps if you use the standard TYPO3 commit message style ;)

PR #147, opened 2016-09-16
[TASK] Use strpos instead of stristr

sascha-egerer:
`stristr` creates a copy of the string which is not required if we
just want to check if a string contains a string.
The case insensitive check is also not requird if we check for characters
that do not have an uppercase representation.

### DIFF
--- a/src/Core/Variables/VariableExtractor.php
+++ b/src/Core/Variables/VariableExtractor.php
@@ -111,7 +111,7 @@ class VariableExtractor {
 	 * @return array
 	 */
 	protected function resolveSubVariableReferences($subject, $propertyPath) {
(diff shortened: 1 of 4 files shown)


src/Core/Variables/VariableExtractor.php
```diff
@@ -111,7 +111,7 @@ public function getAccessorsForPath($subject, $propertyPath) {
 	 * @return array
 	 */
 	protected function resolveSubVariableReferences($subject, $propertyPath) {
-		if (stristr($propertyPath, '{')) {
+		if (strpos($propertyPath, '{')) {
```
helhum: please add a `!== FALSE` here as well

NamelessCoder:
Thanks :)

PR #149, opened 2016-09-16
Do not override custom template paths with defaults

sascha-egerer:
`TemplatePaths::fillDefaultsByPackageName()` does set the default path
for templates based on the package key. If pathes are already defined
they are overriden but should not.
Defaults path is now merged with existing paths configurations.

### DIFF
--- a/src/View/TemplatePaths.php
+++ b/src/View/TemplatePaths.php
@@ -365,9 +365,18 @@ class TemplatePaths {
 	 */
 	public function fillDefaultsByPackageName($packageName) {
 		$path = $this->getPackagePath($packageName);
-		$this->setTemplateRootPaths(array($path . self::DEFAULT_TEMPLATES_DIRECTORY));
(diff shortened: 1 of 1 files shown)


IchHabRecht:
@sascha-egerer Are you sure the configuration for your paths is applied before default paths are added in Fluid Standalone?

sascha-egerer:
@IchHabRecht  Looks like that can happen. See https://review.typo3.org/#/c/48279

IchHabRecht:
@sascha-egerer the linked issue is due to internal TYPO3 Fluid magic. IMHO this is nothing we have to care about in Standalone Fluid.

NamelessCoder:
Can't approve, sorry! The fill method are designed to override all existing paths - and just like Nicole writes, the template paths issue you face is specific to TYPO3 CMS and should be fixed there.

Thanks nonetheless for your efforts to solve this (and the +1 on forge) ;)

PR #159, opened 2016-09-27
[FEATURE] Enable Stringlist to array casting

mneuhaus:
This allows the casting of a list of strings to a simple indexed array:
{'foo,bar, baz' as array}


mneuhaus:
putting this down for now, because we might introduce issues that'll shoot us in the foot in the future (special characters, encodings, etc)

PR #173, opened 2016-09-28
[FEATURE] Allow controller-less resolving in templateRootPaths

NamelessCoder:
This change allows resolveAvailableTemplateFiles on
TemplatePaths to be called with a controller name of
NULL, which causes Fluid to look for the action’s
template file in the templateRootPaths directly, no
controller sub-directory being checked.

### DIFF
--- a/src/View/TemplatePaths.php
+++ b/src/View/TemplatePaths.php
@@ -258,14 +258,14 @@ class TemplatePaths {
 	}
 
 	/**
-	 * @param string $controllerName
+	 * @param string|NULL $controllerName
 	 * @param string $format
 	 * @return array
 	 */
 	public function resolveAvailableTemplateFiles($controllerName, $format = self::DEFAULT_FORMAT) {
 		$paths = $this->getTemplateRootPaths();
 		foreach ($paths as $index => $path) {
-			$paths[$index] = $path . $controllerName . '/';
+			$paths[$index] = rtrim($path . $controllerName, '/') . '/';
 		}
 		return $this->resolveFilesInFolders($paths, $format);
 	}


src/View/TemplatePaths.php
```diff
@@ -258,14 +258,14 @@ public function resolveTemplateFileForControllerAndActionAndFormat($controller,
 	}
 
 	/**
-	 * @param string $controllerName
+	 * @param string|NULL $controllerName
 	 * @param string $format
 	 * @return array
 	 */
 	public function resolveAvailableTemplateFiles($controllerName, $format = self::DEFAULT_FORMAT) {
```
mneuhaus: if it's optional anyway, lets make it optional in the signature as well :) ```resolveAvailableTemplateFiles(NULL)``` feels stupid

NamelessCoder: I hesitate to introduce that change just now, as it would be a breaking change for PHP7 on platforms providing a custom TemplatePaths implementation - but you are correct and we should do this at some point (and then bump to version 2.0). But let's accumulate some more breaking changes before we do that ;)

mneuhaus: how about tracking future deprecation "todos" in the doc comment?

```
the method signature will be changed to make $controllerName optional in version 2.0...
```

mneuhaus: better yet, i created a Milestone for 2.0 and added an issue for that change to that milestone :)
https://github.com/TYPO3Fluid/Fluid/issues/177

NamelessCoder: Milestone was the perfect choice :)

PR #176, opened 2016-09-28
[FEATURE] Introduce StopCompilingChildrenException

NamelessCoder:
This change adds a new type of Exception and handling
thereof. The new Exception can be thrown from within
ViewHelpers’ `compile()` method and causes the compiler
to skip compiling:

* Further child nodes down the stack from current node
* Argument values and closures
* Render children closures

Instead, the compiler replaces the entire compiled state
of the ViewHelper that threw the Exception, turning it into
a static string provided in the Exception.

Planned usage is the coming cache improvements which
include a new ViewHelper which uses this to create a
completely static value for the compiled template.

### DIFF
--- a/src/Core/Compiler/NodeConverter.php
+++ b/src/Core/Compiler/NodeConverter.php
@@ -143,57 +143,67 @@ class NodeConverter {
 
 		// Build up $arguments array
 		$argumentsVariableName = $this->variableName('arguments');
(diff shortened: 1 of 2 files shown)


NamelessCoder:
Probably needs a rebase after the others. Diff shows indentation issue that git diff does not show locally.

PR #180, opened 2016-10-02
[FEATURE] Support delegate ParsedTemplateInterface in f:render

NamelessCoder:
This change allows passing the class name of a custom
ParsedTemplateInterface implementation as `delegate`
argument for `f:render`.

This causes Fluid to load and verify the type of the class
you name and call the `render()` method with a cloned
rendering context and assigned `arguments` as input.

Allows fully bypassing any resolving, parsing and caching
that is normally done when you invoke `f:render`. The
delegate ParsedTemplateInterface then completely takes
over the task of rendering the output (and can use any
method you like).

Bonus: powerful enough to allow things like plugging in
any other template engine as a rendering delegate and
combine it with recursive rendering, content argument,
default value, optional output etc.


src/ViewHelpers/RenderViewHelper.php
```diff
@@ -98,17 +115,27 @@ public function render() {
...
 		$tagContent = $this->renderChildren();
 
 		if ($contentAs !== NULL) {
 			$arguments[$contentAs] = $tagContent;
 		}
 
 		$content = '';
-		if ($partial !== NULL) {
+		if ($delegate !== NULL) {
+			if (!is_a($delegate, ParsedTemplateInterface::class, TRUE)) {
```
mneuhaus: ```=== FALSE``` instead of ``!`` ?

src/ViewHelpers/RenderViewHelper.php
```diff
@@ -98,17 +115,27 @@ public function render() {
...
+				throw new \InvalidArgumentException(sprintf('Cannot render %s - must implement ParsedTemplateInterface!', $delegate));
+			}
+			$renderingContext = clone $this->renderingContext;
+			$renderingContext->getVariableProvider()->setSource($arguments);
+			$content = (new $delegate())->render($renderingContext);
+		} elseif ($partial !== NULL) {
 			$content = $this->viewHelperVariableContainer->getView()->renderPartial($partial, $section, $arguments, $optional);
 		} elseif ($section !== NULL) {
 			$content = $this->viewHelperVariableContainer->getView()->renderSection($section, $arguments, $optional);
+		} elseif (!$optional) {
```
mneuhaus: ```=== FALSE``` instead of ```!``` please :)

NamelessCoder: We're trying to move away from that explicitness unless the function potentially returns something not-boolean - which `is_a` does not. The `!== NULL` stuff is a little different since it checks (uncritically) if a VH argument was not provided and still has the default value. Oh and for `$optional` specifically it should already be cast to boolean above ;)

PR #181, opened 2016-10-03
[FEATURE] Improve multiple aspects of caching/compiing

NamelessCoder:
This is a rather extensive change. It implements:

* The ability to specify a RenderingContext class name
  for the CLI command.
* A new `warmup` action for the CLI command.
* FluidCacheWarmerInterface and StandardCacheWarmer.
* Method on FluidCacheInterface to return Warmer.
* Methods on TemplateCompiler to govern warmup mode.
* ViewHelper `f:cache.disable` to prevent compiling.
* ViewHelper `f:cache.warmup` to assign warmup-time
  default variables.
* ViewHelper `f:cache.static` to force converting all child
  nodes of ViewHelper to a static string in compiled result.
* FluidCacheWarmupResult assistant class to report
  warmup results to user.

The CLI command’s new action allows calling for example:

```sh
./bin/fluid --warmup --renderingContext My\Special\Context \
  --templateRootPaths "/path/one" "/path/two" \
  --partialRootPaths "/partials/one" "/partials/two" \
  --layoutRootPaths "/path/to/layouts" \
  --arguments "/path/to/jsonfile.json"
```

Which then performs a complete compiling of all template
files located in the specified paths, with TemplateCompiler
in warmup mode to allow `f:cache.warmup` to apply its
variables while compiling. If/when a piece of template code
cannot be compiled the CLI reports possible mitigations,
small advise which is handled by the CacheWarmer.

The CacheWarmer pattern is also possible to implement
in other ways - the API of this class requires just a
RenderingContext as input (which then returns configured
paths, variables and other custom context) and returns a
FluidCacheWarmupResult which allows analysis of the
result of compiling each individual file and reporting those
results to the user in new ways, CLI or not.

Includes examples for warmup-only variables and static
caching in example files.


NamelessCoder:
NOTE: **THIS IS STILL WIP!**

I plan to add additional unit tests and clean up some areas (also including adding more phpdoc for the public parts) - the collection of features is not expected to change though, so please feel free to begin reviewing and testing the active code ;)

NamelessCoder:
Ready for review.

bin/fluid
```diff
@@ -108,6 +118,42 @@ class FluidCommandLine {
...
+	protected function renderWarmupResult(\TYPO3Fluid\Fluid\Core\Cache\FluidCacheWarmupResult $result) {
+		$string = PHP_EOL . 'Template cache warmup results' . PHP_EOL . PHP_EOL;
+		foreach ($result->getResults() as $templatePathAndFilename => $aspects) {
+			$string .= sprintf(
+				"%s\n    Compiled? %s\n    Has Layout? %s\n",
+				$templatePathAndFilename,
+				$aspects[\TYPO3Fluid\Fluid\Core\Cache\FluidCacheWarmupResult::RESULT_COMPILABLE] ? 'YES' : 'NO',
+				$aspects[\TYPO3Fluid\Fluid\Core\Cache\FluidCacheWarmupResult::RESULT_HASLAYOUT] ? 'YES' : 'NO'
+			);
+			if ($aspects[\TYPO3Fluid\Fluid\Core\Cache\FluidCacheWarmupResult::RESULT_COMPILABLE]) {
```
mneuhaus: did you mean ``isset(...)``?

bin/fluid
```diff
@@ -108,6 +118,42 @@ class FluidCommandLine {
...
+				$string .= sprintf(
+					"    Compiled as: %s\n",
+					$aspects[\TYPO3Fluid\Fluid\Core\Cache\FluidCacheWarmupResult::RESULT_COMPILEDCLASS]
```
mneuhaus: this is not the same constant as the if, is this correct? maybe add that as isset to the if as well?

bin/fluid
```diff
@@ -108,6 +118,42 @@ class FluidCommandLine {
...
+				);
+			}
+			if (isset($aspects[\TYPO3Fluid\Fluid\Core\Cache\FluidCacheWarmupResult::RESULT_FAILURE])) {
+				$string .= sprintf(
+					"    Compilation failure reason: %s\n",
+					$aspects[\TYPO3Fluid\Fluid\Core\Cache\FluidCacheWarmupResult::RESULT_FAILURE]
+				);
+			}
+			if (isset($aspects[\TYPO3Fluid\Fluid\Core\Cache\FluidCacheWarmupResult::RESULT_MITIGATIONS])) {
```
mneuhaus: lets make sure, it's not only present, but an array as well?

src/Core/Cache/StandardCacheWarmer.php
```diff
@@ -0,0 +1,297 @@
...
+	 * about all detected template files and the compiling of
+	 * those files. If a template fails to compile or throws an
+	 * error, a mitigation suggestion is included for that file.
+	 *
+	 * @param RenderingContextInterface $renderingContext
+	 * @return FluidCacheWarmupResult
+	 */
+	public function warm(RenderingContextInterface $renderingContext) {
+		$renderingContext->getTemplateCompiler()->enterWarmupMode();
+		$formats = array('html', 'xml', 'txt', 'json', 'rtf', 'atom', 'rss');
```
mneuhaus: does it make sense, to make this a protected property, so you could extend and override this easily?

src/Core/Cache/StandardCacheWarmer.php
```diff
@@ -0,0 +1,297 @@
...
+	 * @param array $formats
+	 * @return FluidCacheWarmupResult
+	 */
+	protected function warmupTemplateRootPaths(RenderingContextInterface $renderingContext, array $formats) {
+		$result = new FluidCacheWarmupResult();
+		$paths = $renderingContext->getTemplatePaths();
+		foreach ($formats as $format) {
+			$paths->setFormat($format);
+			foreach ($this->detectControllerNamesInTemplateRootPaths($paths->getTemplateRootPaths()) as $controllerName) {
+				foreach ($paths->resolveAvailableTemplateFiles($controllerName, $format) as $templateFile) {
```
mneuhaus: we might want to refactor to reduce nesting here, although it's still "ok" i guess

src/Core/Cache/StandardCacheWarmer.php
```diff
@@ -0,0 +1,297 @@
...
+	 * Detect all available controller names in provided TemplateRootPaths
+	 * array, returning the "basename" components of controller-template
+	 * directories encountered, as an array.
+	 *
+	 * @param string $templateRootPaths
+	 * @return \Generator
+	 */
+	protected function detectControllerNamesInTemplateRootPaths(array $templateRootPaths) {
+		foreach ($templateRootPaths as $templateRootPath) {
+			foreach ((array) glob(rtrim($templateRootPath, '/') . '/*') as $pathName) {
```
mneuhaus: glob was pretty sketchy for me on windows in the past, do we care about that?

mneuhaus: and maybe we should use ```PATH_SEPARATOR``` instead of hard-coding '/'

src/Core/Compiler/TemplateCompiler.php
```diff
@@ -37,13 +39,49 @@ class TemplateCompiler {
 	protected $renderingContext;
 
 	/**
+	 * @var string
+	 */
+	protected $mode = self::MODE_NORMAL;
```
mneuhaus: constant to property assignment, wasn't there this impossible until "recently" in php?
do you know, since when this is possible?

mneuhaus: maybe move it to the constructor to be safe?

src/ViewHelpers/Cache/WarmupViewHelper.php
```diff
@@ -0,0 +1,163 @@
...
+ * then you can consider using `f:cache.disable` to prevent
+ * the template compiler from even attempting to compile it.
+ *
+ * USE WITH CARE! SOME EDGE CASES OF FOR EXAMPLE VIEWHELPERS
+ * WHICH REQUIRE SPECIAL VARIABLE TYPES MAY NOT BE SUPPORTED
+ * HERE DUE TO THE RUDIMENTARY NATURE OF VARIABLES YOU DEFINE.
+ *
+ * = Examples =
+ *
+ * <code title="Usage and effect">
```
mneuhaus: i think you forgot to update this documentation :)

src/ViewHelpers/Cache/WarmupViewHelper.php
```diff
@@ -0,0 +1,163 @@
...
+	/**
+	 * @var boolean
+	 */
+	protected $escapeOutput = FALSE;
+
+	/**
+	 * @return void
+	 */
+	public function initializeArguments() {
+		$this->registerArgument(
```
mneuhaus: #rant: man, i hate the registerArgument... it's so "unreadable"...

tests/Unit/ViewHelpers/Cache/WarmupViewHelperTest.php
```diff
@@ -0,0 +1,77 @@
...
+			'argumentName',
+			'closureName',
+			$initializationCode,
+			$viewHelperNode,
+			new TemplateCompiler()
+		);
+	}
+
+	/**
+	 * @tet
```
mneuhaus: ``@test`` ;)

tests/Unit/ViewHelpers/Cache/WarmupViewHelperTest.php
```diff
@@ -0,0 +1,77 @@
...
+		$renderingContext->expects($this->never())->method('setVariableProvider');
+		$renderingContext->setTemplateCompiler($compiler);
+		$subject = new WarmupViewHelper();
+		$subject->setRenderChildrenClosure(function() { return 'test'; });
+		$subject->setRenderingContext($renderingContext);
+		$this->assertEquals('test', $subject->render());
+	}
+
+	/**
+	 * @tet
```
mneuhaus: ``@test`` ;)

bin/fluid
NamelessCoder: It is correct; the compiled class name is only output if the template is compilable, but it is set (to the template identifier) also if the template is not compilable.

src/Core/Cache/StandardCacheWarmer.php
NamelessCoder: It does indeed!

src/Core/Cache/StandardCacheWarmer.php
NamelessCoder: I think we can defend two nested loops although one would be best... I am considering how/whether to implement Generators here (but there will also be a coming PR to do Generators explicitly).

mneuhaus: ok, it just kinda poked my eye ;)

src/Core/Cache/StandardCacheWarmer.php
NamelessCoder: `DIRECTORY_SEPARATOR` substituted - I'm not sure about `glob` on Windows but since this is both wrapped in a CLI command, intended for deployment process, and possible to override in custom warmers, I tend to not really care too much about Windows here ;)

src/Core/Compiler/TemplateCompiler.php
NamelessCoder: Hmm, if I remember correctly this was first supported in PHP 4. But must admit my memory isn't that good. It is *certainly* supported on all our supported versions ;)

mneuhaus: ok, maybe i'm confusing this with a ``new ClassName()`` something wasn't possible there well beyond 5.0, afaik

NamelessCoder: The only requirement is those assignments must not call any active code, e.g. no functions, no constructors, no `constant('name')`, `isset()` or anything of that sort. But constants work regardless of how. Maybe you're thinking about the `\Vendor\Class::class` thing which is fairly recent in general?

Issue #184, opened 2016-10-07
Properties of \ArrayAccess objects may become inaccessible.


kitsunet:
The following scenario is currently not solvable with the Accessor handling as is, but was before standalone:

Given a class roughly like this (skipped some implementation details):
```
class Foo implements \ArrayAccess
{
  // ArrayAccess implementation

  getSomePropertyNotAccessibleByArrayAccess()
  {
    return $this->somePropertyNotAccessibleByArrayAccess;
  }
}
```

Instance available in the template as `foo`:

```
{foo.somePropertyNotAccessibleByArrayAccess}
```

Will result in `NULL` because the getter is not used as soon as the object was detected as implementing `\ArrayAccess`.

IMHO correct would be to check if `getSomePropertyNotAccessibleByArrayAccess` (or any of the other allowed "getter" methods) is available or if it is a public property, before using ArrayAccess.

bwaidelich:
Just for reference:
This is the logic that currently applies in the Flow Package:
For `{someThing.foo}`:
* `$someThing->getFoo()`
* `$someThing->isFoo()`
* `$someThing->hasFoo()`
* `$someThing->foo` (if `\stdClass`)
* `$someThing['foo']` (if `\ArrayAccess` or `\SplObjectStorage`

in that order

NamelessCoder:
> This is the logic that currently applies in the Flow Package:

Fluid's logic has been optimised for performance and clarity - and this is what now prevents this use case. The particular behavior from Flow comes from the ObjectAccess (correct me if wrong please!) and can be implemented if desired, using a custom VariableProvider (see https://github.com/NamelessCoder/TYPO3.CMS/blob/master/typo3/sysext/fluid/Classes/Core/Variables/CmsVariableProvider.php for reference; this provider is solely designed to delegate object accessing to ObjectAccess).

I guess my actual question is: how prevalent is this use case of mixing ArrayAccess with getter methods and does that prevalence merit changing the intentionally optimised accessing in Fluid?

kitsunet:
First of all this can not be fixed in the VariableProvider, I tried but due to predetermination of Accessors you can end up with a compiled template directly accessing via:
`...->getVariableProvider()['foo']['bar']` Where `foo` is the array access object and `bar` is a gettable property.

[...]

PR #185, opened 2016-10-07
[BUGFIX] Makes object properties accessible in ArrayAccess (#184)

kitsunet:
Due to the precedence for ArrayAccess accessors a property only
accessible via getter method in an ArrayAccess object cannot be
retrieved via Fluid currently.

This change adjusts the precedence so that for objects the
ArrayAccess is used as fallback after checking getter methods
for the given propertyName.

### DIFF
--- a/src/Core/Variables/VariableExtractor.php
+++ b/src/Core/Variables/VariableExtractor.php
@@ -191,7 +191,7 @@ class VariableExtractor {
 	 * @return string|NULL
 	 */
 	protected function detectAccessor($subject, $propertyName) {
-		if (is_array($subject)) {
+		if (is_array($subject) || ($subject instanceof \ArrayAccess && $subject->offsetExists($propertyName)) {
 			return self::ACCESSOR_ARRAY;
 		}
 		if (is_object($subject)) {
@@ -206,9 +206,6 @@ class VariableExtractor {
 			}
 			if (property_exists($subject, $propertyName)) {
 				return self::ACCESSOR_PUBLICPROPERTY;
-			}
-			if ($subject instanceof \ArrayAccess) {
-				return self::ACCESSOR_ARRAY;
 			}
 		}
 
--- a/src/Core/Variables/VariableExtractor.php
+++ b/src/Core/Variables/VariableExtractor.php
@@ -191,7 +191,7 @@ class VariableExtractor {
 	 * @return string|NULL
 	 */
 	protected function detectAccessor($subject, $propertyName) {
-		if (is_array($subject) || ($subject instanceof \ArrayAccess && $subject->offsetExists($propertyName)) {
+		if (is_array($subject) || ($subject instanceof \ArrayAccess && $subject->offsetExists($propertyName))) {
 			return self::ACCESSOR_ARRAY;
 		}
 		if (is_object($subject)) {
--- a/src/Core/Variables/VariableExtractor.php
+++ b/src/Core/Variables/VariableExtractor.php
@@ -191,9 +191,10 @@ class VariableExtractor {
 	 * @return string|NULL
 	 */
 	protected function detectAccessor($subject, $propertyName) {
-		if (is_array($subject) || $subject instanceof \ArrayAccess) {
+		if (is_array($subject)) {
 			return self::ACCESSOR_ARRAY;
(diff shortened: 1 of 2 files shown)


NamelessCoder:
Just so I am up-to-date:

* which current use case has an ArrayAccess with getter methods?
* is the use case valid due to normal procedures or is the incompatibility caused by off-label use of ArrayAccess?

I lean towards not allowing getter methods to take priority over ArrayAccess since ArrayAccess is a very public contract for your object which clearly says "I can be accessed as an array". Putting that contract as fallback sounds wrong to me - off the bat. A demonstrated use case would convince me ;)

NamelessCoder:
I've adjusted the code a tiny bit so it completely preserves the current logic w/o consulting getter methods if an offset exists in the ArrayAccess. This way we have a preference for ArrayAccess but fallback to getter/asserter methods and public property. Hopefully this is agreeable :)

NamelessCoder:
I do so love the new "allow edits from maintainers" feature!

kitsunet:
Thanks, that looks fine by me, will test with our use case later today but should work 👍

NamelessCoder:
Hang on - I will check this out locally and add a unit test case for this fallback :)

kitsunet:
ok, works for our use case :+1: Good to go from my side.

bwaidelich:
Just for the record (sorry, I only saw this PR now with the update of #184): This is _not_ yet in sync with the behavior of the Flow package, see for example:
```php
class SomeClass implements \ArrayAccess {
  public function isFoo() {
    return 'bar';
  }
}

$instance = new SomeClass();
$instance['foo'] = 'baz';
```

NamelessCoder:
@bwaidelich If we change this around we will be giving priority to the getter/asserter methods and completely prevent extraction of the value in the internal array/storage and it would be a breaking change (because we view the point of decoupling, not the Flow package, as the starting point). I would insist we give the PHP-native implementations priority and our own extraction methods secondary priority - this current solution is a compromise to at least continue consulting other methods if the native method does not know about the value.

NamelessCoder:
...if there are still things in Flow this would not support I will be more than happy to take a look and provide suggestions!

bwaidelich:
Thanks for the feedback.

> I would insist we give the PHP-native implementations priority

Absolutely and I'm not arguing for a Flow compatibility here but for a consistent behavior.
I (and all those we discussed it with when we implemented it that way) would expect any manually implemented getter to have precedence over an offset in the internal array (which might even change during runtime).
Otherwise it will be quite difficult to implement some domain logic on top of value-bags.

Take this (simplified) example:

```php
public class PersonNode extends Node // that implements \ArrayAccess
{
  public function getFullName()
  {
    return $this['givenName'] . ' ' . $this['familyName'];
  }
}
```

I would expect
```
{person.fullName}
```

to always execute the code above, but currently it would depend on the state of the instance which I'd consider rather error prone.

> we view the point of decoupling, not the Flow package, as the starting point

It was always like this in Fluid. However, please consider this just as FYI – I'm not really involved in the sync process anymore.

> if there are still things in Flow this would not support I will [...] to take a look and provide suggestions

Thanks! It's not so much about our code, we can fix that. I just fear some hard-to-debug behavioral changes in existing code out there.

NamelessCoder:
I will of course take this into consideration but from the perspective of Fluid standalone, `ArrayAccess` is a language construct and getter methods are a convention so natural priority goes to the language construct. This is why I don't agree that "it was always like this in Fluid". It was always like that in *Flow's Fluid* but it was never that way in *Standalone Fluid* because it never contained `ObjectAccess`.

What I meant with giving suggestions was not so much about fixing code in Fluid but finding an approach which accommodates your use cases while not breaking this priority preference. Something like (completely freestyle) `{object[property]}` to be absolutely explicit about the chosen access method might make sense. So don't hesitate to include me, please, as it provides me with real life use cases!

NB: using `f:debug` on such an object causes the dumping to go through the framework's debugger (if you overrule `f:debug` which you most likely do) which would expose both the properties and the internal array of an `ArrayAccess` (I am aware there are exceptions so this is just generally speaking). So, knowing that the `ArrayAccess` has priority should be everything you need in order to predict and debug this case.

Issue #141, opened 2016-09-13
CycleViewHelper doesn't work when static rendered


kitsunet:
The CycleViewHelper depends on a class property to store the current cycle it is rendering. This works when rendered initially but when rendered from a compiled state a new instance of the VH is created on every cycle effectively keeping the result stuck with the first cycle value.

Need to make up my mind how to solve that in a way that allows nesting as well...

NamelessCoder:
You may want to read https://github.com/TYPO3Fluid/Fluid/pull/25 as inspiration here.

kitsunet:
Perfect, will come up with a PR then!

NamelessCoder:
Fixed a while ago :)

PR #193, opened 2016-10-12
[BUGFIX] CycleViewHelper correctly cycles even compiled (#141)

kitsunet:
The CycleViewHelper relied on it's own instance to iterate
through all possible cycle values, which didn't work
when the template was compiled because the calls were made
to separate instances of the ViewHelper.

This fix adjusts the CycleViewHelper to work with the
`ViewHelperVariableContainer` and a proper static rendering.

### DIFF
--- a/src/ViewHelpers/CycleViewHelper.php
+++ b/src/ViewHelpers/CycleViewHelper.php
@@ -6,6 +6,7 @@ namespace TYPO3Fluid\Fluid\ViewHelpers;
  * See LICENSE.txt that was shipped with this package.
  */
 
+use TYPO3Fluid\Fluid\Core\Rendering\RenderingContextInterface;
 use TYPO3Fluid\Fluid\Core\ViewHelper;
 use TYPO3Fluid\Fluid\Core\ViewHelper\AbstractViewHelper;
 
@@ -47,25 +48,10 @@ use TYPO3Fluid\Fluid\Core\ViewHelper\AbstractViewHelper;
  */
 class CycleViewHelper extends AbstractViewHelper
 {
-
     /**
      * @var boolean
      */
     protected $escapeOutput = false;
-
-    /**
-     * The values to be iterated through
-     *
-     * @var array|\SplObjectStorage
-     */
-    protected $values = null;
-
-    /**
-     * Current values index
-     *
-     * @var integer
-     */
-    protected $currentCycleIndex = null;
 
     /**
      * @return void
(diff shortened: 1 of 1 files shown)


kitsunet:
Fixes #141

NamelessCoder:
Doing a quick edit for CGL in a few seconds. Will add some notes with a suggested improvement.

NamelessCoder:
Suggestion: you can use the `CompileWithContentArgumentAndRenderChildren` trait on this one to gain a much more efficient compiled execution. That trait also allows the first optional argument to serve as a default content argument, meaning `$values = $renderChildrenClosure();` can be used and you can assume you get either the first optional argument value or output of `renderChildren()`.

This should however be done in a followup PR since this can be backported to 1.x but the Trait cannot.

NamelessCoder:
Oh, one last thing. Would you mind adding an example usage for `f:cycle` in https://github.com/TYPO3Fluid/Fluid/blob/master/examples/example_structures.php and an assertion in https://github.com/TYPO3Fluid/Fluid/blob/master/tests/Functional/ExamplesTest.php#L169 (test simply asserts that each string in that array exists in the rendered output). That would make this perfect :)

If you don't have time let me know and I will put the task on my queue.

kitsunet:
👍 should be doable in the next days. Thanks for CGL fix, should be ok in my IDE too by now...

NamelessCoder:
We may be converting to PSR-2 very, very soon so don't fall in love with the current standard ;)

NamelessCoder:
All this needs is a conversion to PSR-2 and we're good to go :)

mneuhaus:
Ping? :)

kitsunet:
Yep, sorry, was super busy the last few days :(

mneuhaus:
Don't worry, just wanted to ping, so it's not forgotten at some point :)

kitsunet:
Ok, at least the PSR-2 and rebase is done.

src/ViewHelpers/CycleViewHelper.php
```diff
@@ -85,45 +71,70 @@ public function initializeArguments()
...
+        if (is_object($values) && $values instanceof \Traversable) {
+            return iterator_to_array($values, false);
         }
-        $this->currentCycleIndex = 0;
+
+        throw new ViewHelper\Exception('CycleViewHelper only supports arrays and objects implementing \Traversable interface', 1248728393);
+    }
+
+    /**
+     * @param $as
```
NamelessCoder: Lacks `string` in annotation.

NamelessCoder:
Tests run, code looks good, one last CGL to take care of!

PR #195, opened 2016-10-13
[FEATURE] Improve multiple aspects of caching/compiing (#181)

NamelessCoder:
This is a rather extensive change. It implements:

* The ability to specify a RenderingContext class name for the CLI command.
* A new `warmup` action for the CLI command.
* FluidCacheWarmerInterface and StandardCacheWarmer.
* Method on FluidCacheInterface to return Warmer.
* Methods on TemplateCompiler to govern warmup mode.
* ViewHelper `f:cache.disable` to prevent compiling.
* ViewHelper `f:cache.warmup` to assign warmup-time default variables.
* ViewHelper `f:cache.static` to force converting all child nodes of ViewHelper to a static string in compiled result.
* FluidCacheWarmupResult assistant class to report warmup results to user.

The CLI command’s new action allows calling for example:

```sh
./bin/fluid --warmup --renderingContext My\Special\Context \
  --templateRootPaths "/path/one" "/path/two" \
  --partialRootPaths "/partials/one" "/partials/two" \
  --layoutRootPaths "/path/to/layouts" \
  --arguments "/path/to/jsonfile.json"
```

Which then performs a complete compiling of all template files located in the specified paths, with TemplateCompiler in warmup mode to allow `f:cache.warmup` to apply its
variables while compiling. If/when a piece of template code cannot be compiled the CLI reports possible mitigations, small advise which is handled by the CacheWarmer.

The CacheWarmer pattern is also possible to implement in other ways - the API of this class requires just a RenderingContext as input (which then returns configured paths, variables and other custom context) and returns a FluidCacheWarmupResult which allows analysis of the
result of compiling each individual file and reporting those results to the user in new ways, CLI or not.

Includes examples for warmup-only variables and static caching in example files.


NamelessCoder:
NB: previously passed review as https://github.com/TYPO3Fluid/Fluid/commit/9c0fd385abb0afeb6b19c9651ec9d558143cff56 which was reverted due to a breaking change. This can be merged again and we can release 2.0.0.

Issue #201, opened 2016-10-17
Error with 1.1.1


kitsunet:
I get a `Call to undefined method TYPO3Fluid\Fluid\Core\Compiler\TemplateCompiler::getCurrentlyProcessingState()` since updating to 1.1.1

Called in AbstractTemplateView line 335.

PR #202, opened 2016-10-17
Revert "[TASK] Allow rendering without stack through View (#174)"

NamelessCoder:
This reverts commit 81dbc0e5d72700a7797dfffface001a77d0d880b.

Close: #201

### DIFF
--- a/src/View/AbstractTemplateView.php
+++ b/src/View/AbstractTemplateView.php
@@ -322,17 +322,17 @@ abstract class AbstractTemplateView extends AbstractView {
 	 */
 	protected function getCurrentRenderingType() {
 		$currentRendering = end($this->renderingStack);
-		return $currentRendering['type'] ? $currentRendering['type'] : self::RENDERING_TEMPLATE;
+		return $currentRendering['type'];
 	}
 
 	/**
-	 * Get the parsed template which is currently being rendered or compiled.
+	 * Get the parsed template which is currently being rendered.
 	 *
 	 * @return ParsedTemplateInterface
 	 */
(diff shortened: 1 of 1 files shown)


NamelessCoder:
@kitsunet if you are still up, you are welcome to merge this one (and I will then tag 1.1.2 right after). Good catch btw, but be careful you don't have a bug in the code you are writing regarding https://github.com/TYPO3Fluid/Fluid/issues/201 - the fact that you hit this case could indicate that you have some incompletely initialised View being used to render a template, e.g. `startRendering` was not called and the stack is therefore empty, or the stack was incorrectly popped. Just something to watch out for :)

kitsunet:
Yep, still up, works flawlessly. I think I know why we hit that case, maybe something to discuss about individually because it's an important use case for us. We render sections within a template without actually rendering the full template.

Merging, thanks for fast reaction! 👍

NamelessCoder:
> We render sections within a template without actually rendering the full template.

That is the *exact* use case I had in mind. You may want to check out https://github.com/TYPO3Fluid/Fluid/commit/a72efbdc667b90dc2594f8a9a4e640cc32e601ae which is the reason the fallbacks were implemented - I knew in advance this would come up, but it should still be exclusive to the 2.0 branch and 1.0 / 1.1 will have to make do :)

There's a related patch I've been wanting to do which moves the whole stack handling away from the View, further separating the concerns. Maybe that will also allow us to somewhat clean up the TemplateParser which has a pretty big set of couplings and dependencies...

NamelessCoder:
1.1.2 has been tagged :)

PR #208, opened 2016-10-21
[BUGFIX] Require minimum pcre lib 8.36

NamelessCoder:
This clearly reports a requirement of the TemplateParser
which uses expressions that are only supported on the
stated minimum version of PCRE. Before this the parser
might cause segmentation faults.


ksjogo:
It was some code from the default introduction package, I could look into getting the actual code after the con.

PR #214, opened 2016-10-31
[FEATURE] Add additional VH assistant Traits

NamelessCoder:
This adds three new traits:

* DefaultRenderMethod which contains the frequently duplicated render() method passthrough to renderStatic.
* PassthroughRenderChildren which has a render() method that only returns child content.
* CompileEmpty which causes the ViewHelper to compile to an empty string.

Together they remove some code duplication and provide further ways to remove code
duplication in packages providing ViewHelpers.


NamelessCoder:
This is mostly a proposal - since we face a new 3.0 release it could make sense to allow a slight breaking change to achieve less code duplication in Fluid and adapters/VH-packages too.

PR #215, opened 2016-10-31
[BUGFIX] Variable access ending in digits in array expressions (#190)

### LINKED ISSUE #190: Variable name cut off
Given an array assigned to `templateVariables` that contains a key `user2`.
Now I do:

```
<f:alias map="{user: testVariables.user2, firstName: testVariables.user2.firstName}">
{user.lastName}, {firstName}
</f:alias>
```

this fails, `user.lastName` (and after debugging also `user`) is null inside the alias (the direct access to `testVariables.user2.lastName` works just fine). Looking at the compiled template the resulting variable access is:
`$array1['user'] = $renderingContext->getVariableProvider()->getByPath('testVariables.user', $array2);`

As you can see the path was cut off and the `2` is missing in the end.

Will try to deliver more debug details, but for now this seems to b a bug in parse/compile.

kitsunet:
This fixes the accessing of variables with names (or keys) ending in
digits used in array expressions.

Specifically an expression like this: `foo: bar2` was previously wrongly
parsed to `foo: bar` and resulted in wrong variables being accessed.

This has been fixed while at the same time retaining existing behavior
for parsing raw numeric values.

Tests are added to prove the corrected behavior.

Fixes #190

### DIFF
--- a/src/Core/Parser/Patterns.php
+++ b/src/Core/Parser/Patterns.php
@@ -229,22 +229,25 @@ abstract class Patterns
      *
      */
     static public $SPLIT_PATTERN_SHORTHANDSYNTAX_ARRAY_PARTS = '/
-		(?P<ArrayPart>                                             # Start sub-match
-			(?P<Key>                                               # The keys of the array
-				[a-zA-Z0-9\\-_]+                                   # Unquoted
-				|"(?:\\\"|[^"])+"                                  # Double quoted
-				|\'(?:\\\\\'|[^\'])+\'                             # Single quoted
+		(?P<ArrayPart>                                                      # Start sub-match
+			(?P<Key>                                                        # The keys of the array
+				 [a-zA-Z0-9_-]+                                             # Unquoted
+				|"(?:\\\\"|[^"])+"                                          # Double quoted
+				|\'(?:\\\\\'|[^\'])+\'                                      # Single quoted
 			)
-			\s*:\s*                                                # Key|Value delimiter :
-			(?:                                                    # Possible value options:
-				(?P<QuotedString>                                  # Quoted string
-					(?:"(?:\\\"|[^"])*")
-					|(?:\'(?:\\\\\'|[^\'])*\')
+			\\s*:\\s*                                                       # Key|Value delimiter :
+			(?:                                                             # Possible value options:
+				(?P<QuotedString>                                          # Quoted string
+					 "(?:\\\\"|[^"])*"
+					|\'(?:\\\\\'|[^\'])*\'
 				)
-				|(?P<VariableIdentifier>[a-zA-Z0-9\-_.]*[a-zA-Z]+)  # variable identifiers must contain letters (otherwise they are hardcoded numbers)
-				|(?P<Number>[0-9.]+)                               # Number
-				|{\s*(?P<Subarray>(?:(?P>ArrayPart)\s*,?\s*)+)\s*} # Another sub-array
-			)                                                      # END possible value options
-		)                                                          # End array part sub-match
+				|(?P<VariableIdentifier>
+					(?:(?=[^,{}\.]*[a-zA-Z])[a-zA-Z0-9_-]*)                        # variable identifiers must contain letters (otherwise they are hardcoded numbers)
+					(?:\\.[a-zA-Z0-9_-]+)*                                  # but subsequent access of keys and properties may be only numbers
+				)
+				|(?P<Number>[0-9]+(?:\\.[0-9]+)?)                           # Number
+				|\\{\\s*(?P<Subarray>(?:(?P>ArrayPart)\\s*,?\\s*)+)\\s*\\}  # Another sub-array
+			)                                                               # END possible value options
+		)\\s*(?=\\z|,|\\})                                                           # End array part sub-match
 	/x';
 }
--- a/tests/Unit/Core/Parser/TemplateParserPatternTest.php
+++ b/tests/Unit/Core/Parser/TemplateParserPatternTest.php
@@ -461,6 +461,129 @@ class PatternsTest extends UnitTestCase
     /**
      * @return array
      */
+    public function dataProviderValidArrayExpressionsBeginningAndEndingOnDigits()
+    {
+        return [
+            [
+                'expression' => '{baz: foo.bar1}',
+                'expected' => [
+                    0 => [
+                        0 => 'baz: foo.bar1',
+                        'ArrayPart' => 'baz: foo.bar1',
+                        1 => 'baz: foo.bar1',
+                        'Key' => 'baz',
+                        2 => 'baz',
+                        'QuotedString' => '',
+                        3 => '',
+                        'VariableIdentifier' => 'foo.bar1',
+                        4 => 'foo.bar1'
+                    ]
+                ]
+            ],
+            [
+                'expression' => '{"foo2": "bar2"}',
+                'expected' => [
+                    0 => [
+                        0 => '"foo2": "bar2"',
+                        'ArrayPart' => '"foo2": "bar2"',
+                        1 => '"foo2": "bar2"',
+                        'Key' => '"foo2"',
+                        2 => '"foo2"',
+                        'QuotedString' => '"bar2"',
+                        3 => '"bar2"'
+                    ]
+                ]
+            ],
+            [
+                'expression' => '{foo3: bar.baz3}',
+                'expected' => [
+                    0 => [
+                        0 => 'foo3: bar.baz3',
+                        'ArrayPart' => 'foo3: bar.baz3',
+                        1 => 'foo3: bar.baz3',
+                        'Key' => 'foo3',
+                        2 => 'foo3',
+                        'QuotedString' => '',
+                        3 => '',
+                        'VariableIdentifier' => 'bar.baz3',
+                        4 => 'bar.baz3'
+                    ]
+                ]
+            ],
+            [
+                'expression' => '{foo4: _4bar.-baz4}',
+                'expected' => [
+                    0 => [
+                        0 => 'foo4: _4bar.-baz4',
+                        'ArrayPart' => 'foo4: _4bar.-baz4',
+                        1 => 'foo4: _4bar.-baz4',
+                        'Key' => 'foo4',
+                        2 => 'foo4',
+                        'QuotedString' => '',
+                        3 => '',
+                        'VariableIdentifier' => '_4bar.-baz4',
+                        4 => '_4bar.-baz4'
+                    ]
+                ]
+            ],
+            [
+                'expression' => '{5foo5: -5bar}',
+                'expected' => [
+                    0 => [
+                        0 => '5foo5: -5bar',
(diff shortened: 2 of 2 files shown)


liayn:
i'm missing testcases with the other allowed chars.
eg. a trailing underscore is not possible after a number.

Can you tell me the goal? "any number inside an identifier, but not at the beginning or after a dot" ? right?

src/Core/Parser/Patterns.php
```diff
@@ -241,7 +241,7 @@
 					(?:"(?:\\\"|[^"])*")
 					|(?:\'(?:\\\\\'|[^\'])*\')
 				)
-				|(?P<VariableIdentifier>[a-zA-Z0-9\-_.]*[a-zA-Z]+)  # variable identifiers must contain letters (otherwise they are hardcoded numbers)
+				|(?P<VariableIdentifier>[a-zA-Z0-9\-_.]*[a-zA-Z]+[a-zA-Z0-9.]*) # variable identifiers must contain letters (otherwise they are hardcoded numbers)
```
liayn: The `\-` should be escaped, also for PHP single quoted strings, or put the dash to the very end of the character class.
Is a dot at the very end really allowed? What about an underscore at the very end?

kitsunet:
Right, I added mostly test cases for the bug this is supposed to fix. I can of course add more test cases although that is IMHO out of scope of this change.

The problem are the constraints with the given parser and expected results.

`foo: 234` as well as `foo: 23.4` should result in number values being parsed, while
`foo: foo2`, `foo: foo.bar2`, `foo: foo.2`, `foo: 2bar` should all be variable access.

I tried to come up with a nicer regex for doing this, but it usually ends up in some of the other test cases breaking. Basically everything with a char in it and no quotes is supposed to be a variable access for this part of the parser.

src/Core/Parser/Patterns.php
kitsunet: I can do the escaping of course, but I concentrated on the bugfix and not adjusting already existing code.

Dot at the end is not really valid, but I currently see no good way to prevent it in this stage of parsing, let me think this through again though. Underscore is debatable, it wasn't before, so I left it out as well.

liayn:
Do we consider `foo: 2_foo_._toBe4Yo_2_`  a valid variable access?

NamelessCoder:
> Do we consider foo: 2_foo_._toBe4Yo_2_ a valid variable access?

Technically, no. But there are now arguments for supporting it, e.g. we need to function in other frameworks that might be using such naming in arrays for example.

liayn:
I'll make up my head for a proper regex to cover that all.

kitsunet:
Fair question, as said, currently it's not, IMHO that's a separate issue to discuss. Digits AND underscores definitely were allowed in TYPO3 and Flow before.
Also just checked the Flow Fluid parser at this stage allows expressions ending in dots even if they are technically not valid. But IMHO that's fine.

NamelessCoder:
Thanks @liayn - don't forget there are two different expressions; one for array syntax and another for object access / shorthand VH / expression node.

tests/Unit/Core/Parser/TemplateParserPatternTest.php
```diff
@@ -461,6 +461,112 @@ public function SPLIT_PATTERN_SHORTHANDSYNTAX_ARRAY_PARTS_matchesQuotedKeys()
...
+     * @dataProvider dataProviderValidArrayExpressionsBeginningAndEndingOnDigits()
+     * @test
+     */
+    public function SPLIT_PATTERN_SHORTHANDSYNTAX_ARRAY_PARTS_matchesKeysEndingInDigits($expression, $expected)
+    {
+        $pattern = Patterns::$SPLIT_PATTERN_SHORTHANDSYNTAX_ARRAY_PARTS;
+        $success = preg_match_all($pattern, $expression, $matches, PREG_SET_ORDER) > 0;
+        $this->assertTrue($success);
+        $this->assertEquals(
+            $expected, $matches,
```
NamelessCoder: Very minor nitpick - these should be on individual lines.

tests/Unit/Core/Parser/TemplateParserPatternTest.php
```diff
@@ -461,6 +461,112 @@ public function SPLIT_PATTERN_SHORTHANDSYNTAX_ARRAY_PARTS_matchesQuotedKeys()
...
+                        2 => 'foo3',
+                        'QuotedString' => '',
+                        3 => '',
+                        'VariableIdentifier' => 'bar.baz3',
+                        4 => 'bar.baz3'
+                    ]
+                ]
+            ],
+            [
+                'expression' => '{\'foo4\': 5}',
```
NamelessCoder: I'm not 100% sure about this one... we do support *double*quoted strings (so we match JSON there) but single-quoted strings I think should only be possible as values. Hopefully it isn't some coincidence that lets this work.

No objection if a FEATURE does this later on, pending what @liayn is looking into. But it does not seem to be part of this bugfix?

src/Core/Parser/Patterns.php
NamelessCoder: The `.` at the end of that expression is probably too broad?

liayn:
Seems like Christian and me found a working regex solution... which fixes another 7 trillion of bugs, at least it feels like that :-D

kitsunet:
Here we go, thanks @liayn for the nice co-working :)

src/Core/Parser/Patterns.php
```diff
@@ -229,22 +229,25 @@
...
      */
     static public $SPLIT_PATTERN_SHORTHANDSYNTAX_ARRAY_PARTS = '/
-		(?P<ArrayPart>                                             # Start sub-match
-			(?P<Key>                                               # The keys of the array
-				[a-zA-Z0-9\\-_]+                                   # Unquoted
-				|"(?:\\\"|[^"])+"                                  # Double quoted
-				|\'(?:\\\\\'|[^\'])+\'                             # Single quoted
+		(?P<ArrayPart>                                                      # Start sub-match
+			(?P<Key>                                                        # The keys of the array
+				 [a-zA-Z0-9_-]+                                             # Unquoted
```
NamelessCoder: You are sure that the `-` should *not* be escaped? Normally this indicates a range.

src/Core/Parser/Patterns.php
```diff
@@ -229,22 +229,25 @@
...
+					|\'(?:\\\\\'|[^\'])*\'
 				)
-				|(?P<VariableIdentifier>[a-zA-Z0-9\-_.]*[a-zA-Z]+)  # variable identifiers must contain letters (otherwise they are hardcoded numbers)
-				|(?P<Number>[0-9.]+)                               # Number
-				|{\s*(?P<Subarray>(?:(?P>ArrayPart)\s*,?\s*)+)\s*} # Another sub-array
-			)                                                      # END possible value options
-		)                                                          # End array part sub-match
+				|(?P<VariableIdentifier>
+					(?:(?=[^,{}\.]*[a-zA-Z])[a-zA-Z0-9_-]*)                        # variable identifiers must contain letters (otherwise they are hardcoded numbers)
+					(?:\\.[a-zA-Z0-9_-]+)*                                  # but subsequent access of keys and properties may be only numbers
```
NamelessCoder: I would reword this to something like `but keys coming after the root variable may contain pure numbers and need a less strict expression`. I had a bit of difficulty reading and understanding this comment at first.

src/Core/Parser/Patterns.php
```diff
@@ -229,22 +229,25 @@
...
+				)
+				|(?P<Number>[0-9]+(?:\\.[0-9]+)?)                           # Number
```
NamelessCoder: This expression seems to have changed from scanning for a straight up number, to scanning for a number as well as strings like `1.5`. If the intention is to add support for decimals I propose to do that in a separate FEATURE pull request (and add tests for that specific behavior).

src/Core/Parser/Patterns.php
```diff
@@ -229,22 +229,25 @@
...
+				|\\{\\s*(?P<Subarray>(?:(?P>ArrayPart)\\s*,?\\s*)+)\\s*\\}  # Another sub-array
+			)                                                               # END possible value options
+		)\\s*(?=\\z|,|\\})                                                           # End array part sub-match
```
NamelessCoder: Perhaps explain this ending pattern a bit closer. Should this pattern not only be matching for a closing `}`?

src/Core/Parser/Patterns.php
liayn: if a dash is at the end of a character class, it does not need escaping.

src/Core/Parser/Patterns.php
liayn: I was told this has to be supported

src/Core/Parser/Patterns.php
liayn: The ArrayPart is used recursively. Hence at least comma and brace need to be covered. I just added the end of string anchor to ensure we match for an end in any case. This was totally missing before which leads to false positives to be accepted. `2.2bar` would be parially excepted then

src/Core/Parser/Patterns.php
kitsunet: Looking at the old Number expression, it was already supposed to work before, but was not really strict enough. IMHO this is not a change.

src/Core/Parser/Patterns.php
kitsunet: It's definitely needed as @liayn explains. Ending of string is needed for recursive sub arrays that will be given to this expression without opening/closing braces due to line 249.

kitsunet:
@NamelessCoder we both agreed that these expressions could possibly need a a bigger overhaul, but that shouldn't be part of this change.

src/Core/Parser/Patterns.php
liayn: I agree that we have to make the description more precise here.

NamelessCoder:
@kitsunet I am aware of that - one of my review questions also tries to discern whether the decimals support belongs here. Others are about potential issues so I do need a response to those, and I personally need to understand the code I'm tasked with maintaining ;)

NamelessCoder:
Oops, sorry, did not see the review comments. GitHub's AJAX refresh-comment-feed thingie is not perfect it seems. Reloading the thread page shows the comments!

NamelessCoder:
I think we're down to one decision then:

* Should the decimals support be done in a separate change / overhaul of expressions?

I would prefer to do that since it is a change, and an important one. It should not be stowed away as a sub-part of a bug fix about variable names. I'm happy to have it, but separate it to a FEATURE please :)

liayn:
@NamelessCoder IMHO you understood that wrong. The support for decimals was always there!
The regex just allowed something like `2.2.`, which was wrong. We just fixed that to ensure such a thing is not matched.

NamelessCoder:
@liayn Thanks for clarifying - yep, I had understood that wrong.

kitsunet:
Ups, now AJAX updates were bad for me, thanks for merging @NamelessCoder 
I just pushed another commit to clarify the comments ;) Will open a new PR

Issue #209, opened 2016-10-26
RawViewHelper doesn't use `value` argument


kitsunet:
The `\TYPO3Fluid\Fluid\ViewHelpers\Format\RawViewHelper` registers a value argument to be used instead of the children but in fact never uses it in renderStatic nor in the compiled version, leading to NULL results in case the VH was used with the value argument eg.

`<f:format.raw value="don't encode" />`

helhum:
This was introduced with #113 

The fix should include tests for this as weil

NamelessCoder:
When using the CompileWithContentArgumentAndRenderStatic trait the *Closure* that renderStatic receives is where the decision is made to use the content argument or the original renderChildren closure that was built. Therefore the argument will never explicitly be consumed in the renderStatic method - but in your compiled template you definitely should see a ternary condition being built which generates a different closure if the content argument is set (`value` in this case; name is not required to be anything, the FIRST optional argument is always used).

kitsunet:
Well, I see that a render children closure is build that returns null and this is just called and returned.
I have a test case here that exposes the problem.
Note that the RawViewHelper overwrites the `renderStatic` method which is probably why the trait doesn't have an effect.

kitsunet:
Yes, that was a good info, removing the `renderStatic` method from the `RawViewHelper` fixes the problem.

NamelessCoder:
I have something for this, hang on :)

PR #217, opened 2016-11-02
[BUGFIX] Fix f:format.raw used with "value" argument

NamelessCoder:
Since https://github.com/TYPO3/Fluid/pull/113, `f:format.raw` needs
a corresponding ternary condition in compiled template code because
it overrides the `compile()` method inherited by the Trait.

Close: #209


NamelessCoder:
This one should fix the issue in compiled templates - and https://github.com/TYPO3/Fluid/pull/210 should fix it in uncompiled/uncompilable templates.

Issue #219, opened 2016-11-02
[FEATURE] easier way to disable "{{" (and maybe also "{' for certain Javascript templates then CDATA per "{"


AndreasA:
I think it would be really nice to have an option to disable the "{{" support due to certain JavaScript templates without losing "{" support and not having to add CDATA per "{" that has such values.

For Javascript Code itself containing JSON disabling "{" would also be an idea.

Maybe in a way like disabling htmlspecialchars for certain code blocks that disables either "{{" or "{"

Probably not the highest priority though.

kitsunet:
That would mean you need another language construct to make that happen. Ofc that could be a Viewhelper but it makes parsing a lot more complicated.
IMHO all parsing on or off is a solid way to work with this everyting else is overly complicated.

AndreasA:
I think that would be enough too, I can just reenable the parsing for my fluid values then.

PR #220, opened 2016-11-02
[FEATURE] Template source passthrough pre-processor

NamelessCoder:
This adds a new pre-processor which handles a new type of
modifer in templates:

`{passthrough on}`

Which when enabled causes the template to be output instead
of parsed. This can then be used in partials, templates or layouts
to completely avoid parsing the template.

Useful if you have a piece of template code that collides with
Fluid's syntax - putting this code in a partial, enabling passthrough
and rendering that partial then outputs the source (and removes
the modifier itself).

Close: #219

### DIFF
--- a/examples/Resources/Private/Singles/Passthrough.html
+++ b/examples/Resources/Private/Singles/Passthrough.html
@@ -1,0 +1,5 @@
+{parsing off}
+
+<f:section name="Main">
+    <f:format.raw>This does not get parsed; the source is passed through with Fluid markup</f:format.raw>
+</f:section>
--- a/examples/example_passthrough.php
+++ b/examples/example_passthrough.php
@@ -1,0 +1,24 @@
+<?php
+
+/*
+ * EXAMPLE: Parsing modifier
+ *
+ * This example shows you how to use the "parsing off"
+ * modifier in template files to disable all Fluid
+ * parsing. Useful when for example you want to render
+ * a template or partial file without processing it as
+ * Fluid, but still being able to render it (=reading
+ * and passing through) using `f:render`.
+ */
+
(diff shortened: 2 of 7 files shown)


mneuhaus:
i'm generally ok with the idea, *but* imho "passthrough" is not clear enough, what it actually causes. maybe something like ```{fluid off}```, ```{parsing off}````, ```{static on}``` ?

NamelessCoder:
We could switch this to `{parsing off}`, that might be clearer indeed.

mneuhaus:
```{fluid:parsing off}``` would make it even more obvious, that it is something related to fluid and would introduce a general syntax, if we ever need similar "pre-proccessor" options?

NamelessCoder:
Since pre-processors aren't really parsed but rather just string-detected and replaced with empty strings they shouldn't mess with anything. There's no chance of erroneously catching a JSON expression.

Besides, `{namespace foobar}` is also possible to register an ignored NS. So it matches the other ones we have.

mneuhaus:
after sleeping over it ```{parsing off}``` sounds perfect

NamelessCoder:
So let it be written - so let it be done!

NamelessCoder:
And done it was.

AndreasA:
Just to be clear:
I put this at the begining of a file and the parsing for that file is disabled?
That also includes sub-partials / sub-templates (in case of e.g. a layout)?

sascha-egerer:
Don't know how the behavior is but how would you reache a sub template if fluid is disabled. But the current result would be internesting...

AndreasA:
Of course, you are right, it would just output <f:render ...> instead :D

PR #230, opened 2016-11-11
[WIP][FEATURE] Convert TagBasedViewHelper to a Trait

NamelessCoder:
This change turns the AbstractTagBasedViewHelper into
a Trait to implement in any ViewHelper that outputs a
tag, without having to subclass any particular class.

Gives third parties more freedom to create their own
type of sub-classes. Also theoretically allows things like
segmented link/uri ViewHelpers to be combined into
one and render differently depending on presence
of child content.

### DIFF
--- a/src/Core/ViewHelper/TagBuilder.php
+++ b/src/Core/ViewHelper/TagBuilder.php
@@ -44,6 +44,24 @@ class TagBuilder
     protected $forceClosingTag = false;
 
     /**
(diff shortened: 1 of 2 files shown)


NamelessCoder:
Don't merge - will split this into two features; one for the new option on TagBuilder and another for the trait.

NamelessCoder:
Will be replaced with new PR.

PR #246, opened 2016-12-29
[BUGFIX] MathExpression not working with camel case variables

thommyhh:
Adjust the regular expression to allow camel case variable names.

### DIFF
--- a/src/Core/Parser/SyntaxTree/Expression/MathExpressionNode.php
+++ b/src/Core/Parser/SyntaxTree/Expression/MathExpressionNode.php
@@ -26,7 +26,7 @@ class MathExpressionNode extends AbstractExpressionNode
 		(
 			{                                # Start of shorthand syntax
 				(?:                          # Math expression is composed of...
-					[a-zA-Z0-9\.]+[\s]?[*+\^\/\%\-]{1}[\s]?[a-zA-Z0-9\.]+   # Various math expressions left and right sides with any spaces
+					[a-zA-Z0-9\.]+(?:[\s]?[*+\^\/\%\-]{1}[\s]?[a-zA-Z0-9\.]+)+   # Various math expressions left and right sides with any spaces
 					|(?R)                    # Other expressions inside
 				)+
 			}                                # End of shorthand syntax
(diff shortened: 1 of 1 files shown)


thommyhh:
I came across those two bugs while trying to use the expression {imageWidth / 16 * 9}, which should be possible.

src/Core/Parser/SyntaxTree/Expression/MathExpressionNode.php
```diff
@@ -26,7 +26,7 @@ class MathExpressionNode extends AbstractExpressionNode
 		(
 			{                                # Start of shorthand syntax
 				(?:                          # Math expression is composed of...
-					[a-zA-Z0-9\.]+[\s]?[*+\^\/\%\-]{1}[\s]?[a-zA-Z0-9\.]+   # Various math expressions left and right sides with any spaces
+					[a-zA-Z0-9\.]+(?:[\s]?[*+\^\/\%\-]{1}[\s]?[a-zA-Z0-9\.]+)+   # Various math expressions left and right sides with any spaces
```
NamelessCoder: This looks like it would change the returned matches grouping - why is it necessary? The current expression already uses `+` to ensure multiple characters get matched?

thommyhh: In this expression are no returned matching groups. Multiple characters are good, but is should also match multiple groups. I tried, e.g. the following expressions:

{imageWidth / 16 * 9} worked, this matched as "imageWith / 1" and "6 * 9"
{imageWidth / 3 * 2} did not work.
{imageWidth / 3 + 7 * 5} did not work.

With the changed regular expression all of them work now.

NamelessCoder: Thanks for the explanation and examples! Please add these to the commit message and adjust the topic so it explains camelCase isn't the only bug fixed - alternatively, split this into two commits which fix the two bugs separately.

NamelessCoder: Nevermind, sorry - I overlooked that you already made two commits for this. Will merge without squashing.

NamelessCoder:
Merged and released as 2.1.2.

PR #248, opened 2017-01-10
[FEATURE] Add variable assigning ViewHelper

NamelessCoder:
New ViewHelper `f:variable` with arguments "name"
and "value", assigns a variable in the VariableProvider
using the `add()` method and does not remove the
variable again afterwards.

Usages:

    {f:variable(name: 'myvariable', value: 'some value')}
    <f:variable name="myvariable">some value</f:variable>
    {oldvariable -> f:format.htmlspecialchars() -> f:variable(name: 'newvariable')}
    <f:variable name="myvariable"><f:format.htmlspecialchars>{oldvariable}</f:format.htmlspecialchars></f:variable>

### DIFF
--- a/src/ViewHelpers/VariableViewHelper.php
+++ b/src/ViewHelpers/VariableViewHelper.php
@@ -1,0 +1,64 @@
+<?php
+namespace TYPO3Fluid\Fluid\ViewHelpers;
+
+/*
+ * This file belongs to the package "TYPO3 Fluid".
+ * See LICENSE.txt that was shipped with this package.
+ */
+
+use TYPO3Fluid\Fluid\Core\Compiler\TemplateCompiler;
+use TYPO3Fluid\Fluid\Core\Parser\SyntaxTree\ViewHelperNode;
+use TYPO3Fluid\Fluid\Core\Rendering\RenderingContextInterface;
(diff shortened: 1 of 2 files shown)


src/ViewHelpers/VariableViewHelper.php
```diff
@@ -0,0 +1,71 @@
...
+ *     <f:variable name="myvariable">some value</f:variable>
+ *     {oldvariable -> f:format.htmlspecialchars() -> f:variable(name: 'newvariable')}
+ *     <f:variable name="myvariable"><f:format.htmlspecialchars>{oldvariable}</f:format.htmlspecialchars></f:variable>
+ *
+ * Please note:
+ *
+ * If you assign a variable with a dotted name, e.g.
+ * "myarray.property" then this may or may not be
+ * supported by the VariableProvider that's being
+ * used. The default
```
mneuhaus: The default...?

mneuhaus: + i'm not really sure we should allow a dot in the name. My gut tells me, that prohibiting dot here will save use plenty of headaches later on

NamelessCoder: That part can go in another PR after we discuss it ;)

PR #251, opened 2017-01-13
[FEATURE] Add Hide ViewHelper

NamelessCoder:
Adds a new ViewHelper which executes but does
not output anything in the tag content.

Inspired by the "Nothing" formatting ViewHelper
from the TYPO3 extension "news".


NamelessCoder:
Adding to universal ViewHelper package instead.

Issue #255, opened 2017-01-17
Accessor detection during compile can be (very) wrong


kitsunet:
The accessors for ObjectAccessNodes are detected during compile, but this can have very weird side effects if variable names are "reused".

This is a made up, simplified example from a bigger template/partial construct but it shows the issue:

template has a variable "foo" assigned which is an array with a key "bar" which is an object and a key "baz" which is a string. In the template foo is used like `{foo.baz}` and everything works fine.
Now a partial is rendered from the template with `arguments="{foo: foo.bar}"`, inside the partial again foo is accessed `{foo.mee}`. Now in the partial foo is the object bar unlike in the template were foo is an array.

The result is that accessors are detected as array for both cases as the detection for template and partials is working with the original variable provider given to the template. Obviously in the above example the problem is easy to spot and could in theory be fixed by changing the naming. Although in more complex combinations of templates and partials it can easily happen to reuse a variable name (also considering that variable names are assigned at "runtime", eg. by a ForViewHelper.

I am not sure how much introspection the compile step can do to avoid this, but currently it can be very messy.

PR #256, opened 2017-01-18
[BUGFIX] Disable variable extractors

NamelessCoder:
Although this means a minor performance hit in compiled
templates in particular, we drop the variable extraction method
detection temporarily.

The extractor detection will be restored in another and much
larger pull request which will also merge the variable extraction
detection and the extraction logic into VariableProvider rather
than forcing use of Fluid's VariableExtractor.

Close: #255

### DIFF
--- a/src/Core/Parser/TemplateParser.php
+++ b/src/Core/Parser/TemplateParser.php
@@ -445,12 +445,7 @@ class TemplateParser
 
         // Object Accessor
         if (strlen($objectAccessorString) > 0) {
-            if ($state->isCompilable() && !$state->isCompiled()) {
-                $accessors = VariableExtractor::extractAccessors($state->getVariableContainer(), $objectAccessorString);
-            } else {
-                $accessors = [];
-            }
-            $node = new ObjectAccessorNode($objectAccessorString, $accessors);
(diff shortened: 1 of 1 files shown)


NamelessCoder:
This will be phase one of a larger operation. First, we'll disable the extraction method detection which will allow the described use case in #255 to work at the cost of some performance in compiled templates.

Phase two will be to merge the variable extraction detection and actual extraction into the VariableProvider, so VariableExtractor is no longer forced and becomes redundant.

Phase three will be to correct the object accessor parsing logic away from using the variable provider temporarily assigned to the ParsingState (which is how we store parsed sections and layout names). The full extent of this operation is currently unknown.

When all this is done it should be a formality to make the VariableProvider remember the accessors it uses for a particular variable. The trick here is that VariableProvider is already being replaced when we call renderPartial / renderSection - so when a new VariableProvider gets created, the accessors are not remembered and must be detected again.

It then also becomes completely optional for a third party VariableProvider, if it even wishes to be capable of remembering accessors (as a performance gaining measure).

Thanks to @kitsunet for the initial investigation and inspiration for temporary workaround.

Issue #257, opened 2017-01-19
Refactor condition ViewHelper logic


NamelessCoder:
The following refactoring is suggested:

1. Move render() method from IfViewHelper to AbstractConditionViewHelper.
2. Declare evaluateCondition abstract on AbstractConditionViewHelper, requiring implementations to implement the method.
3. Move "condition" argument from AbstractConditionViewHelper to IfViewHelper.
4. Change method signature for evaluateCondition to require an array and not allow null values.

**All of the above points would constitute breaking changes.**

The result is that condition ViewHelpers can be written by simply subclassing AbstractConditionViewHelper, implementing evaluateCondition and finally (if needed by evaluateCondition method) override initializeArguments, call parent::initializeArguments to gain "then" and "else" argument support, then add whichever argument the specific condition ViewHelper requires (in the case of IfViewHelper, that means the "condition" argument). If a ViewHelper requires no additional arguments as input, only the evaluateCondition method is needed (example: conditions which check hardcode-named environment variables).

The benefit will be that condition ViewHelpers get a streamlined interface that is strict enough to ensure operability, but loose enough to let third party condition ViewHelpers safely use parent's arguments for "then" and "else" and add any named argument it wants, to determine true/false evaluation.

This will also prevent frameworks from being able to use render() method arguments on any condition ViewHelper; but arguments registered in initializeArguments will of course still work.

Example implementation after conversion (complete class):

```php
class IfViewHelper extends AbstractConditionViewHelper
{

    /**
     * Initializes the "then" and "else" arguments
     */
    public function initializeArguments()
    {
        parent::initializeArguments();
        $this->registerArgument('condition', 'boolean', 'Condition expression conforming to Fluid boolean rules', false, false);
    }

    /**
     * Simplest of all possible evaluations - returns true if the
     * boolean argument in "condition" is true.
     *
     * @param array|NULL $arguments
     * @return boolean
     * @api
     */
    protected static function evaluateCondition($arguments = null)
    {
        return (boolean) $arguments['condition'];
    }
}
```

Example abstract condition ViewHelper class (partial class):

```php
abstract class AbstractConditionViewHelper extends AbstractViewHelper
{

    /**
     * @var boolean
     */
    protected $escapeOutput = false;

    /**
     * Initializes the "then" and "else" arguments
     */
    public function initializeArguments()
    {
        $this->registerArgument('then', 'mixed', 'Value to be returned if the condition if met.', false);
[...]

PR #258, opened 2017-01-20
[WIP][TASK] Refactor condition ViewHelper logic

NamelessCoder:
This patch makes a few but breaking changes to the
way AbstractConditionViewHelper and implementations
function. The following is changed:

* Method `render` is now delared on AbstractConditionViewHelper
   and removed from IfViewHelper. Implementers of the
   AbstractConditionViewHelper class no longer need to
   implement a `render` method - `evaluateCondition` is
   enough; and if necessary, additional/other VH arguments.
* Method `evaluateCondition` is now declared abstract
   on AbstractConditionViewHelper, meaning that any
   implementations must implement this method.
* Argument `condition` is no longer automatically added
   by subclassing AbstractConditionViewHelper. This
   argument is instead added by implementations, which
   must also call `parent::initializeArguments` to gain "then"
   and "else" argument support. The "then" and "else"
   arguments are in addition now optional (but `f:then` and
   `f:else` node support is forced).
* Implementations of AbstractConditionViewHelper can now
   implement any argument as condition (file for "file exists",
   haystack and needle for "string contains", and so on).

The change is breaking to any implementations of the
AbstractConditionViewHelper class which 1) do not declare
an `evaluateCondition` method and/or 2) overrides arguments
initialisation without calling `parent::initializeArguments` or
finally, 3) assumes that `condition` argument gets passed but
does not explicitly declare it as argument.

Subclasses of IfViewHelper are not prone to this problem.
Should you need a quick fix for your custom condition VH then
simply switching parent class to IfViewHelper should in most
cases make your subclass work again.

### DIFF
--- a/src/Core/ViewHelper/AbstractConditionViewHelper.php
+++ b/src/Core/ViewHelper/AbstractConditionViewHelper.php
@@ -24,14 +24,13 @@ use TYPO3Fluid\Fluid\Core\Rendering\RenderingContextInterface;
  * <[aConditionViewHelperName] .... then="condition true" else="condition false" />,
  * or as well use the "then" and "else" child nodes.
  *
- * @see TYPO3Fluid\Fluid\ViewHelpers\IfViewHelper for a more detailed explanation and a simple usage example.
+ * @see \TYPO3Fluid\Fluid\ViewHelpers\IfViewHelper for a more detailed explanation and a simple usage example.
  * Make sure to NOT OVERRIDE the constructor.
  *
  * @api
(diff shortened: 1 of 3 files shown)


NamelessCoder:
This is **SUGGESTED AND BREAKING** - do not merge until this has been thoroughly covered in https://github.com/TYPO3/Fluid/issues/257.

NamelessCoder:
Replaced by https://github.com/TYPO3/Fluid/pull/393

PR #260, opened 2017-01-30
fix typo in annotation

Kleisli:
This just changes the @retrurn annotation to @return in the DebugViewHelper.php.

### DIFF
--- a/src/ViewHelpers/DebugViewHelper.php
+++ b/src/ViewHelpers/DebugViewHelper.php
@@ -140,7 +140,7 @@ class DebugViewHelper extends AbstractViewHelper
 
     /**
      * @param mixed $variable
-     * @retrurn array
+     * @return array
      */
     protected static function getValuesOfNonScalarVariable($variable)
     {


NamelessCoder:
Thanks @Kleisli, nice catch :)

albe:
This should probably have been targeted lower in retrospect, since the buggy annotation was introduced as a Feature into version back down to 1.1.0 in Jul 2016 (? why ?): TYPO3/Fluid@5acd2e9

Yet this fix only landed in 2.3.x+ only in Jan 2017: https://github.com/TYPO3/Fluid/commit/c2ebf95083ea69f70d317c3b65e04dd6ce4e460a

So now, there are versions 1.1.x - 2.2.x containing a wrong annotation.

Just would like to understand a bit better how this came to be :)

NamelessCoder:
Hi @albe,

> Feature into version back down to 1.1.0 in Jul 2016 (? why ?)

Introduced in order to provide validation of argument types contained within Iterable objects.

> So now, there are versions 1.1.x - 2.2.x containing a wrong annotation.

1.x is no longer maintained, if missing in 2.0.x and 2.1.x that's probably a mistake (missing backport) and likely my fault.

albe:
Ah, okay, that's just human and happens :) I thought maybe there was a concious decision. Any chance of getting that backported still or is it too much work?

mrimann:
That would be great if this fix could be backported to v2.1.x

NamelessCoder:
Done!

PR #264, opened 2017-02-15
[FEATURE] Synchronise xmlns extraction with TYPO3 CMS

NamelessCoder:
Due to slight differences in how Fluid detects xmlns declarations
in templates, TYPO3 CMS has had to deliver not one, but two
separate methods of detecting and registering namespaces.

This patch synchronises the behavior and removes the need for
TYPO3 CMS to process the template for namespaces.

There is one main difference introduced by this patch:

All xmlns declarations are now stripped from the tag; and the tag
itself is stripped away if `data-namespace-typo3-fluid="true"`
is set on the tag.


NamelessCoder:
Pinging @kitsunet - you may be able to remove a couple of files if you bump the Fluid version requirement to the up-coming 2.3.0 release (which I will make if/when this PR gets merged).

More precisely:

* https://github.com/neos/fluidadaptor/blob/master/Classes/Core/Parser/SyntaxTree/Expression/LegacyNamespaceExpressionNode.php
* https://github.com/neos/fluidadaptor/blob/master/Classes/Core/Parser/TemplateProcessor/NamespaceDetectionTemplateProcessor.php

And of course all usages of those. The behaviors they provide should all be supported by Fluid itself after this PR. Removing them should also improve performance slightly (it's always good to not need a pre-processing step!).

PR #269, opened 2017-02-20
[BUGFIX] Reset compiler after each render operation

NamelessCoder:
Prevents issues where calling renderSection multiple
times caused incorrect results because TemplateCompiler
holds on to the first parsed template.

By resetting the compiler in stopRendering() we ensure
that we have a fresh scope when a user calls render,
renderPartial or renderSection multiple times on the
same View instance.


NamelessCoder:
@bmack if you get around to this one before Marc and I do and you decide to merge: if it isn't already selected, please select "squash and merge" from the dropdown before pressing merge! ;)

PR #272, opened 2017-02-28
[BUGFIX] PHP warning with "empty" namespaces

thommyhh:
Check if the namespace identifier exists and is not null before calling
in_array to avoid php warnings.

This is necessary to successfully work with construct like
```xml
{namespace content}
....
<content:encoded>
...
</content:encoded>
```
This is used in templates for rss feeds.


NamelessCoder:
Thanks Tommy - nice catch!

Issue #282, opened 2017-05-10
Foreign namespace declaration builds broken proxy templates


zdreicom:
e.g.

if i use svg image in templates like:

    <f:layout name="Default"/>
    <f:section name="main">
            <svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px" viewBox="0 0 400 160" style="enable-background:new 0 0 400 160;" xml:space="preserve">
                <style type="text/css">
   ...........

will create an  proxy output like this:

    public function addCompiledNamespaces(\TYPO3Fluid\Fluid\Core\Rendering\RenderingContextInterface $renderingContext) {
    $renderingContext->getViewHelperResolver()->addNamespaces(array (
      'core' => 
      array (
        0 => 'TYPO3\\CMS\\Core\\ViewHelpers',
      ),
      'f' => 
      array (
        0 => 'TYPO3Fluid\\Fluid\\ViewHelpers',
        1 => 'TYPO3\\CMS\\Fluid\\ViewHelpers',
      ),
      'formvh' => 
      array (
        0 => 'TYPO3\\CMS\\Form\\ViewHelpers',
      ),
      'xlink' => null // -> Foreign namespace
    ));
    }

this will result in an error:

     1476107295: PHP Warning: in_array() expects parameter 2 to be array, null given in .../typo3_src-8.7.1/vendor/typo3fluid/fluid/src/Core/ViewHelper/ViewHelperResolver.php line 92

NamelessCoder:
@zdreicom thanks for reporting! If you could confirm #283 on TYPO3 CMS that would be great :)

EnzephaloN:
I think I ran into this using TYPO3 8.7.2.
There is a template with that code `<rss version="2.0" xmlns:content="http://purl.org/rss/1.0/modules/content/" xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:atom="http://www.w3.org/2005/Atom">`
And I get this error:
```
PHP Warning: in_array() expects parameter 2 to be array, null given in /home/www/typo3_src/typo3_src-8.7.2/vendor/typo3fluid/fluid/src/Core/ViewHelper/ViewHelperResolver.php
...
24 TYPO3Fluid\Fluid\Core\ViewHelper\ViewHelperResolver::addNamespace("content", NULL)
...
```
[...]

PR #283, opened 2017-05-10
[BUGFIX] Handle adding namespaces to ignored namespaces

NamelessCoder:
Corrects the following two misbehaviors:

1. A second call to add a namespace with a `null` value causes an error; expected: silently keep ignoring namespace.
2. A second call to add a namespace that was previously ignored causes an error; expected: converts ignored namespace to active.

Fixes: #282


opi99:
Patch fixes the issue from #282

tiger-tw:
This fix is not included in 8.7.6. When will it be published in an offical release?
Many thanks for working on this!

NamelessCoder:
This is really more of a question for the framework using Fluid as a composer dependency. The fix *is* published in an official release as far as Fluid itself is concerned.

Generally speaking: use `composer update typo3fluid/fluid` to update the Fluid dependency or set the minimum required Fluid version to `^2.4` if your project requires it.

tiger-tw:
Ah ok, sorry for misunderstanding. 
I was thinking that Typo3 itself would be automatically updated. Thanks for the hint

opi99:
This issue should also go into a branch, so composer package information for TYPO3 v8 LTS can be updated.

NamelessCoder:
Guys, please don't put TYPO3 CMS questions and requests here. Try to keep this to the Fluid library alone. As stated above, this patch *is already released in tagged version 2.4.0 of Fluid* so it is up to third party frameworks whether they want to bump this dependency or just let users do `composer update typo3fluid/fluid`.

PR #285, opened 2017-05-13
[BUGFIX] Do not attempt to escape non-string or -compatible values

NamelessCoder:
Avoids calling htmlspecialchars() on incompatible values. Changes compiling of the escaping node to generate a small closure which checks for string or string-compatible value before escaping.


src/Core/Compiler/NodeConverter.php
```diff
@@ -107,7 +107,12 @@ public function convert(NodeInterface $node)
     protected function convertEscapingNode(EscapingNode $node)
     {
         $configuration = $this->convert($node->getNode());
-        $configuration['execution'] = sprintf('htmlspecialchars(%s, ENT_QUOTES)', $configuration['execution']);
+        $configuration['execution'] = sprintf(
+            'call_user_func_array( function ($var) { ' .
+            'return (is_string($var) || (is_object($var) && method_exists($var, \'__toString\')) ' .
```
sascha-egerer: The `is_string()` should be `is_scalar()`

src/Core/Parser/SyntaxTree/EscapingNode.php
```diff
@@ -41,7 +41,11 @@ public function __construct(NodeInterface $node)
      */
     public function evaluate(RenderingContextInterface $renderingContext)
     {
-        return htmlspecialchars($this->node->evaluate($renderingContext), ENT_QUOTES);
+        $evaluated = $this->node->evaluate($renderingContext);
+        if (is_string($evaluated) || (is_object($evaluated) && method_exists($evaluated, '__toString'))) {
```
sascha-egerer: The `is_string()` should be `is_scalar()`

tests/Unit/Core/Compiler/NodeConverterTest.php
```diff
@@ -145,7 +145,7 @@ public function getConvertTestValues()
             ],
             [
                 new EscapingNode(new TextNode('foo')),
-                'htmlspecialchars(\'foo\', ENT_QUOTES)'
+                'call_user_func_array( function ($var) { return (is_string($var) || (is_object($var) && method_exists($var, \'__toString\')) ? htmlspecialchars((string) $var, ENT_QUOTES) : $var); }, [\'foo\'])'
```
sascha-egerer: The `is_string()` should be `is_scalar()`

src/Core/Compiler/NodeConverter.php
NamelessCoder: I'm not sure I agree - could you illustrate a case where a non-string value would potentially result in unsafe output if (implicitly) cast to a string w/o doing `htmlspecialchars()`?

sascha-egerer: Oh ah yeah... well... thats true. No potential security issue. And we'll get always a string as result which may be antoher problem. So yes - ignore my commts

Issue #286, opened 2017-05-15
Variable extractor should be able to use `has$propertyName()`


cedricziel:
Given a variable with the following class:

```
class Cake
{
  protected $glitter;

  public function hasGlitter () {
    return $this->glitter;
  }
}
```

As of today, the extractor wouldn't be able to extract the value out of the expression `{foo.glitter}`.

The Variable extractor should be able to use methods like `has$Property` like it does handle `is$property`, so they can be used interchangeably.

NamelessCoder:
Change would be welcome. If you do create a PR, please wrap the additional method call/check into the condition which uses the "asserter" constant and make the "hasFoobar" call/check the second priority after the "isFoobar" one.

cedricziel:
Before I open another issue - I think magic methods should be supported as well - today that's not possible, but `symfony/property-access` (think a battle tested `VariableExtractor` on steroids) supports it as well.

What do you think?

NamelessCoder:
I will take a property access replacement into consideration. If other spare parts you've seen lying around will apply to Fluid do mention those as well - if we *are* going to adopt our first dependency, we may as well look into which others might make sense ;)

Thanks for the patch btw!

PR #287, opened 2017-05-15
[FEATURE] Add ability to extract has* accessors

cedricziel:
Closes: #286

### DIFF
--- a/src/Core/Variables/VariableExtractor.php
+++ b/src/Core/Variables/VariableExtractor.php
@@ -1,4 +1,5 @@
 <?php
+
 namespace TYPO3Fluid\Fluid\Core\Variables;
 
 /*
@@ -160,7 +161,7 @@ class VariableExtractor
         } elseif ($accessor === self::ACCESSOR_GETTER) {
             return ($class !== false && method_exists($subject, 'get' . ucfirst($propertyName)));
         } elseif ($accessor === self::ACCESSOR_ASSERTER) {
-            return ($class !== false && method_exists($subject, 'is' . ucfirst($propertyName)));
+            return ($class !== false && $this->isExtractableThroughAsserter($subject, $propertyName));
         } elseif ($accessor === self::ACCESSOR_PUBLICPROPERTY) {
             return ($class !== false && property_exists($subject, $propertyName));
         }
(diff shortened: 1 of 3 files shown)


src/Core/Variables/VariableExtractor.php
```diff
@@ -160,7 +161,7 @@ protected function canExtractWithAccessor($subject, $propertyName, $accessor)
         } elseif ($accessor === self::ACCESSOR_GETTER) {
             return ($class !== false && method_exists($subject, 'get' . ucfirst($propertyName)));
         } elseif ($accessor === self::ACCESSOR_ASSERTER) {
-            return ($class !== false && method_exists($subject, 'is' . ucfirst($propertyName)));
+            return ($class !== false && $this->isExctractableThroughAsserter($subject, $propertyName));
```
NamelessCoder: Typo in `Exctractable`

tests/Unit/ViewHelpers/Fixtures/UserWithoutToString.php
```diff
@@ -40,4 +40,12 @@ public function isNamed()
     {
         return !empty($this->name);
     }
+
+    /**
+     * @return boolean
+     */
+    public function hasNamed()
```
NamelessCoder: I think we should use another name here to make sure we are testing both the `has` and `is` prefixes, for example `hasVirtualProperty` then test property `user.virtualProperty` to determine if the `has` prefix is being detected and used correctly.

src/Core/Variables/VariableExtractor.php
```diff
@@ -18,6 +18,7 @@ class VariableExtractor
     const ACCESSOR_ARRAY = 'array';
     const ACCESSOR_GETTER = 'getter';
     const ACCESSOR_ASSERTER = 'asserter';
+    const ACCESSOR_ASSERTER_HAS = 'asserterHas';
```
NamelessCoder: Rather than add a new constant I would simply use the existing `ACCESSOR_ASSERTER` and when this is specified, check which of the methods to use. But when defining the accessor you should specify "asserter" if the property has either `is` or `has` prefix.

src/Core/Variables/VariableExtractor.php
```diff
@@ -184,6 +185,8 @@ protected function extractWithAccessor($subject, $propertyName, $accessor)
                 return call_user_func_array([$subject, 'get' . ucfirst($propertyName)], []);
             } elseif ($accessor === self::ACCESSOR_ASSERTER) {
                 return call_user_func_array([$subject, 'is' . ucfirst($propertyName)], []);
+            } elseif ($accessor === self::ACCESSOR_ASSERTER_HAS) {
```
NamelessCoder: The `isExtractableThroughAsserted` method is fine - maybe we should add an `extractThroughAsserter` method and call that if `self::ACCESSOR_ASSERTER` is used. See first comment in review.

cedricziel:
Restructured to your suggestions

PR #289, opened 2017-05-16
[BUGFIX] Don't render else-child if its "if" condition evaluates to "false"

### LINKED ISSUE #284: The ìf' argument in an "elseif"-like behavior is ignored when no `else` without an ìf` argument exists
I just created a construct like this in my template
```html
<f:if condition="{condition_1}">
	<f:then>
		...
	</f:then>
	<f:else if="{condition_2}">
		...
	</f:else>
</f:if>
```
and expected that nothing would be output if `condition_1` and `condition_2` both evaluate to false but the contents of `<f:else if="{condition_2}">` were printed. I had to add an extra `<f:else>` (without an `if` argument) to prevent the contents of the `elseif` from being printed.

This seems to be a bug as I don't think that this is the desired behaviour.

jdreesen:
This simple example should render an empty string because the `condition` argument of the `f:if` tag evaluates to `false` and the `if` argument of the `f:else` tag evaluates to `false` too.
```html
<f:if condition="0">
	<f:then>Not true</f:then>
	<f:else if="0">Not true either!</f:else>
</f:if>
```
But the last else-child in a conditional view helper was rendered regardless of the outcome of its "if" condition in a non-compiled (i.e. the first) run. The compiled code seems to work fine, though.

Fixes #284

### DIFF
--- a/src/Core/ViewHelper/AbstractConditionViewHelper.php
+++ b/src/Core/ViewHelper/AbstractConditionViewHelper.php
@@ -168,8 +168,10 @@ abstract class AbstractConditionViewHelper extends AbstractViewHelper
             if ($childNode instanceof ViewHelperNode
                 && substr($childNode->getViewHelperClassName(), -14) === 'ElseViewHelper') {
                 $arguments = $childNode->getArguments();
-                if (isset($arguments['if']) && $arguments['if']->evaluate($this->renderingContext)) {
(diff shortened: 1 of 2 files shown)


NamelessCoder:
You solved the case of the ghost elseif - nice work, and thanks for adding a test!

PR #291, opened 2017-05-19
[TASK] Clean up unused code in BooleanParser

NamelessCoder:
Code is unused, redundant or never reached. Added a couple of test cases and examples illustrating negation of variable in group (this case was not covered in tests).


bmack:
Hey Claus,

in general the patch looks good, I just don't exactly know which parts are not evaluated anymore (currently) and why. Like why removing the Exception that is thrown - does this affect anything or is this the code that was never reached in the first place?

NamelessCoder:
From all that I could test, the code is never reached (either while evaluating or compiling expressions). But better leave this one open and give @mneuhaus a chance to check it - he wrote the boolean parser.

mneuhaus:
It really seems, like there was so much "junk" still left in there, that was simply not reached 😂 
The unit and functional tests are quite extensive and should fail if the code was actually executed.

PR #295, opened 2017-05-19
[FEATURE] Introduce ParserRuntimeOnly trait

NamelessCoder:
Implement in ViewHelpers which only have a purpose
during parse time - such as f:comment, f:section
and f:layout.


src/ViewHelpers/LayoutViewHelper.php
```diff
@@ -11,6 +11,7 @@
 use TYPO3Fluid\Fluid\Core\ViewHelper\AbstractViewHelper;
 use TYPO3Fluid\Fluid\Core\ViewHelper\PostParseInterface;
```
lolli42: note: this interface does not exist.

NamelessCoder: I think we've got this and a few other ones still lingering around. Will follow up with a final cleaning after my local development stuff is all pushed (about 50% is opened as PRs the rest is still waiting here).

PR #296, opened 2017-05-19
[TASK] Throw ViewHelper exception in f:count on uncountable subject

NamelessCoder:


sgrossberndt:
Would it be possible to backport this to to a 2.3.4 to have this in TYPO3 8.7? https://forge.typo3.org/issues/82227

PR #299, opened 2017-06-09
[BUGFIX] Keep one space in spaceless viewhelper

### LINKED ISSUE #298: Spaceless ViewHelper removes too many spaces
The Spaceless ViewHelper makes the assumption, that all whitespaces between a closing (`>`) and opening (`<`) HTML tag marker (if there are only whitespaces) can be removed.

This is especially troublesome, if you have markup with two consecutive inline elements, for example:

```
Lorem ipsum dolor sit amet, <em>consectetuer</em> <strong>adipiscing</strong> elit.
```
This, unexpectedly, removes the space(s) between closing `</em>` and opening `<strong>` and causes those words to stick together when they shouldn't.

This problem can be observed in TYPO3 8 with the new RTE, when the text is styled as shown above.

I would suggest that the spaces should be removed down to one single space, if there is at least one whitespace between `>` and `<`.

nitori:
Fixes #298

### DIFF
--- a/src/ViewHelpers/SpacelessViewHelper.php
+++ b/src/ViewHelpers/SpacelessViewHelper.php
@@ -51,6 +51,6 @@ class SpacelessViewHelper extends AbstractViewHelper
      */
     public static function renderStatic(array $arguments, \Closure $childClosure, RenderingContextInterface $renderingContext)
     {
-        return trim(preg_replace('/\\>\\s+\\</', '><', $childClosure()));
+        return trim(preg_replace('/\\>\\s+\\</', '> <', $childClosure()));
     }
 }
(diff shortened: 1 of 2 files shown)


nitori:
Somehow the hhvm test fails because it's not supported anymore(?)

bmack:
Can you rebase your PR please?

NamelessCoder:
@bmack / @lolli42 I lean towards rejecting this one (see discussion on https://github.com/TYPO3/Fluid/issues/298). I do not (yet) have an alternative suggestion for improved whitespace trimming.

lolli42:
Ok, we abandon this PR now.

PR #303, opened 2017-06-19
[TASK] Discontinue HHVM support

NamelessCoder:
HHVM is no longer supported by testing framework
(and not very popular since the arrival of PHP7+).

This patch disables testing; manifest of tested PHP
versions designate officially supported versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,6 @@ php:
   - 5.6
   - 7
   - 7.1
-  - hhvm
 
 before_install:
   - composer self-update


DanielRuf:
> HHVM is no longer supported by testing framework

The HHVM team fixed several issues, it runs now on the Trusty platform with the latest HHVM release ;-)

See https://github.com/TYPO3/Fluid/pull/314

NamelessCoder:
No longer required.

PR #306, opened 2017-06-19
[FEATURE] Add TagGenerating trait and TagViewHelper

NamelessCoder:
Provides a Trait as replacement for tag based ViewHelpers,
which can be implemented by any ViewHelper, along with
an implementation that allows generating arbitrary tags while
utilising the features of the TagBuilder.

Using the trait is slightly different from using the tag based
abstract ViewHelper class in that all the actual rendering
takes place through a static API method createTag() which
receives arguments and attributes separately and returns
an instance of TagBuilder which is then rendered automatically.

Also refactors TagBuilder to allow chaining and support for
closures as tag content.

### DIFF
--- a/src/Core/ViewHelper/AbstractTagBasedViewHelper.php
+++ b/src/Core/ViewHelper/AbstractTagBasedViewHelper.php
@@ -76,18 +76,12 @@ abstract class AbstractTagBasedViewHelper extends AbstractViewHelper
     }
 
     /**
-     * Sets the tag name to $this->tagName.
-     * Additionally, sets all tag attributes which were registered in
-     * $this->tagAttributes and additionalArguments.
-     *
-     * Will be invoked just before the render method.
-     *
-     * @return void
-     * @api
+     * @param array $arguments
      */
-    public function initialize()
+    public function setArguments(array $arguments)
     {
-        parent::initialize();
+        $this->tag->reset();
+        parent::setArguments($arguments);
         if ($this->hasArgument('additionalAttributes') && is_array($this->arguments['additionalAttributes'])) {
             $this->tag->addAttributes($this->arguments['additionalAttributes']);
         }
--- a/src/Core/ViewHelper/TagBuilder.php
+++ b/src/Core/ViewHelper/TagBuilder.php
@@ -47,7 +47,7 @@ class TagBuilder
      * Constructor
      *
      * @param string $tagName name of the tag to be rendered
-     * @param string $tagContent content of the tag to be rendered
+     * @param string|\Closure $tagContent content of the tag to be rendered
      * @api
      */
     public function __construct($tagName = '', $tagContent = '')
@@ -60,12 +60,13 @@ class TagBuilder
      * Sets the tag name
      *
      * @param string $tagName name of the tag to be rendered
-     * @return void
+     * @return TagBuilder
      * @api
      */
     public function setTagName($tagName)
     {
         $this->tagName = $tagName;
+        return $this;
     }
 
     /**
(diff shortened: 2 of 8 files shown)


src/Core/ViewHelper/Traits/TagGenerating.php
```diff
@@ -0,0 +1,70 @@
...
+ * Tag Generating ViewHelper Trait
+ *
+ * Implement in ViewHelpers which should generate tags.
+ */
+trait TagGenerating
+{
+    /**
+     * @return ArgumentDefinition[]
+     */
+    abstract public function prepareArguments();
```
bmack: Well, this is really hard-to-grasp for people not writing the code itself.
- trait has an abstract public method, but the concrete implementation does not implement it, but the abstractviewhelper implements this.
So this trait should be a an "abstract trait" de-facto. However, "abstract" is part of the AbstractViewHelper but not the TagViewHelper itself.

In addition:
- I don't see a benefit of introducing a TagGenerating trait which is used only in one place (TagViewHelper) which in turn overrides the only method - createTag() viable for TagViewHelper. Instead we could use the new TagViewHelper to extend for new tags directly.

src/Core/ViewHelper/AbstractTagBasedViewHelper.php
```diff
@@ -76,18 +76,12 @@ public function initializeArguments()
...
-     * $this->tagAttributes and additionalArguments.
-     *
-     * Will be invoked just before the render method.
-     *
-     * @return void
-     * @api
+     * @param array $arguments
      */
-    public function initialize()
+    public function setArguments(array $arguments)
```
bmack: Can you move this to a separate PR please?

src/Core/ViewHelper/TagBuilder.php
```diff
@@ -47,7 +47,7 @@ class TagBuilder
      * Constructor
      *
      * @param string $tagName name of the tag to be rendered
```
bmack: Can you move the Fluid handling of the TagBuilder to a separate PR please?

src/Core/ViewHelper/Traits/TagGenerating.php
NamelessCoder: > trait has an abstract public method, but the concrete implementation does not implement it

That is the way we declare that the implementer must also implement that method if it wishes to implement the trait. Doesn't make the trait abstract-like ;)

> I don't see a benefit of introducing a TagGenerating trait which is used only in one place (TagViewHelper)

It is meant to replace the current method of tag capable ViewHelpers but cannot do so by changing the AbstractTagBasedViewHelper class because of 1) the differences in behavior which would be breaking and 2) the desire to make any VH capable of rendering tags rather than demanding the abstract class be implemented.

PS: if no usages of a feature is a reason for dropping, the AbstractTagBasedViewHelper would not exist because Fluid itself does not use it ;)

src/Core/ViewHelper/AbstractTagBasedViewHelper.php
NamelessCoder: Will do - by the way, this one relates to a coming change which cleans up the lifecycle methods we no longer use / we can get rid of easily.

src/Core/ViewHelper/TagBuilder.php
NamelessCoder: Not sure what you precisely mean with this - could you elaborate? Do you want the capability to render closures and the chaining additions in a separate PR or something else?

NamelessCoder:
Abandoned; Fluid 3.0 will provide an improved solution to the use case through refactoring to deprecate AbstractTagBasedViewHelper and encourage direct use of TagBuilder instead, combined with easier handling of arbitrary arguments that become tag attributes.

Issue #309, opened 2017-06-29
LayoutPathAndFilename not used after set


dj-rabel:
Hello,

i just experienced that a expecially set layoutPathAndFilename will never be returned from the getLayoutPathAndFilename method in \TYPO3Fluid\Fluid\View\TemplatePaths

See https://github.com/TYPO3/Fluid/blob/c9f9df06579a351650736c34a87588a1fef292e8/src/View/TemplatePaths.php#L633

It is mentioned in the Docblock Comment but is never returned ...

I've added the following lines to fix this:
```
if ($this->layoutPathAndFilename !== null) {
    return $this->layoutPathAndFilename;
}
```

I would also suggest to change the line https://github.com/TYPO3/Fluid/blob/c9f9df06579a351650736c34a87588a1fef292e8/src/View/TemplatePaths.php#L150 so sanitize the layout path as well
`$this->layoutPathAndFilename = (string) $this->sanitizePath($layoutPathAndFilename);`

And as last, I mentioned that theres no method to set the layoutPathAndFilename in the TYPO3\CMS\Fluid\View\StandaloneView.

greets
Ruben

NamelessCoder:
Sounds like the right way to patch this one (and I agree it should be supported, must have been unintentionally left out during the refactoring - it's a very rarely used feature). You are very welcome to open a pull request with the change exactly as suggested ;)

dj-rabel:
> You are very welcome to open a pull request with the change exactly as suggested ;)
With pleasure

PR #310, opened 2017-06-30
[BUGFIX] Fixes layoutPathAndFilename behavior in TemplatePaths (#309)

dj-rabel:
Class TemplatePaths will now correctly return the class variable layoutPathAndFilename if it was set before

Resolve: #309

### DIFF
--- a/src/View/TemplatePaths.php
+++ b/src/View/TemplatePaths.php
@@ -147,7 +147,7 @@ class TemplatePaths
      */
     public function setLayoutPathAndFilename($layoutPathAndFilename)
     {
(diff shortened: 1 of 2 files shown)


NamelessCoder:
@dj-rabel Thanks for the patch! If you are able, would you mind adding a simple unit test that confirms that if you set this layout path and filename it gets returned from the getter? That will protect the behaviour from being accidentally removed in the future :)

dj-rabel:
Like this?

lg Ruben

NamelessCoder:
That will do it - thanks! :)

PR #311, opened 2017-07-01
[TASK] Return compiled code string from template compiler

NamelessCoder:
Allows the compiler to be used in other ways than just
putting the generated code in the compiler cache. For
example, to analyse or debug-dump the generated code.


DanielRuf:
Great, this would be very helpful for tools which pipe the templates through the compiler and directly work with the resulting markup.

NamelessCoder:
Just to be clear, it returns compiled PHP code (as in, a string containing a PHP class definition where the class name is determined based on template identifier) rather than return markup. The only way you should be working with the resulting output is to save it to a class file or do things like static analysis of the generated code.

DanielRuf:
Understood and perfect for what I would need it =)

NamelessCoder:
Excellent :) just out of curiosity, what are you thinking about building with this?

DanielRuf:
> Excellent :) just out of curiosity, what are you thinking about building with this?

Some realtime template compiler for different template "languages" as a small side project / experiment (not sure if it would be useful for many).

NamelessCoder:
Sounds interesting - you may want to check out TemplateParser->parse() and ParsedTemplateInterface (implementation ParsingState) and how the NodeConverter works. With the right overrides there you can achieve quite a lot of low level access. It's not exactly supported API but it's there for you to (ab)use ;)

Issue #301, opened 2017-06-15
Non-zero value evaluates to FALSE in condition of <f:if> viewhelper


stoppeye:
Testing for a non-zero value in Fluid 8.7.0 doesn't evaluate as TRUE as it used to in 6.2.0 (or PHP in general):

    <f:if condition="-1">
    TEST
    </f:if>

This doesn't print TEST. It seems that only positive non-zero values will test TRUE.

DanielRuf:
Event the docs are not right:
https://docs.typo3.org/typo3cms/ExtbaseFluidBook/8-Fluid/7-using-boolean-conditions.html

> A number for example is evaluated as true if it is greater than 0.

See http://php.net/manual/en/language.types.boolean.php
> Warning
-1 is considered TRUE, like any other non-zero (whether negative or positive) number!

https://softwareengineering.stackexchange.com/a/300792/41772

stoppeye:
Ich bin zur Zeit abwesend und ab dem 3.7. wieder erreichbar. Ihre Mails werden nicht weitergeleitet. Bei dringenden Nachfragen wenden Sie sich bitte an info[at]eye.ch oder telefonisch an 061 416 91 81.

Mit freundlichen Grüssen
Michael Stopp

PR #312, opened 2017-07-01
[BUGFIX] Evaluate negative numbers as boolean like PHP does

DanielRuf:
This fixes https://github.com/TYPO3/Fluid/issues/301

Even the docs are not right:
https://docs.typo3.org/typo3cms/ExtbaseFluidBook/8-Fluid/7-using-boolean-conditions.html

> A number for example is evaluated as true if it is greater than 0.

See http://php.net/manual/en/language.types.boolean.php
> Warning
-1 is considered TRUE, like any other non-zero (whether negative or positive) number!

https://softwareengineering.stackexchange.com/a/300792/41772

Negative numbers in PHP are true. Currently it is not consistent with the negative number handling in PHP. This PR fixes it.

### DIFF
--- a/src/Core/Parser/SyntaxTree/BooleanNode.php
+++ b/src/Core/Parser/SyntaxTree/BooleanNode.php
@@ -155,7 +155,7 @@ class BooleanNode extends AbstractNode
             return $value;
         }
         if (is_numeric($value)) {
-            return (boolean) ((float) $value > 0);
+            return (boolean) ((float) $value);
         }
         if (is_string($value)) {
             if (strlen($value) === 0) {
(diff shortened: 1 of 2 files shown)


NamelessCoder:
Just noting for posterity that the "Fluid Extbase Book" is not the official source of documentation for the Fluid template rendering engine. It applies to Fluid as it exists in TYPO3 CMS up to version 7.6. But we do want to evaluate negative numbers like a boolean cast in PHP would, so patch is very much appreciated!

DanielRuf:
Thanks for the information.

So the logic of Fluid should be as native (how PHP internally works) as possible to have the same code behavior and we can simply transform existing code to Fluid templates if I understand it correctly.

NamelessCoder:
In almost all cases yes. The one exception I can think of right now is how we handle modulo in conditions, i.e. `<f:if condition="{number} % 2">` which works opposite of in PHP (we imply the `=== 0` you'd normally add in PHP). The primary reason is more about predictability/transparency than about being able to transform Fluid template code to PHP code - although that is of course also part of the concern.

DanielRuf:
Good to know, thanks.

> The primary reason is more about predictability/transparency

Right. That's one of my (future) main concerns (making Fluid more transparent).

PR #313, opened 2017-07-01
[TASK] use (float) and (int)

DanielRuf:
See the performance difference at https://3v4l.org/eQaNr. Typecasting is generally faster than the functions.


NamelessCoder:
Casting is preferred so if there are other occurrences it would be great to fix all of them in this patch.

DanielRuf:
I did a full search for boolval, intval and floatval. I can check if there are more occurences, change them and push some more commits.

Will check this later today.

NamelessCoder:
Thanks - and although I doubt there are any, don't forget to scan for `strval` vs. `(string)`.

DanielRuf:
Right.

I might check some other things for future PR's, especially when it comes to unicode handling just to make sure that we are on the safe side and other "problematic" PHP functions and internals.

DanielRuf:
> Thanks - and although I doubt there are any, don't forget to scan for strval vs. (string).

Found no occurence of `strval`.

DanielRuf:
Should I also change the travis.yml config? 
> HHVM is no longer supported on Ubuntu Precise. Please consider using Trusty with `dist: trusty`.

lolli42:
could you rebase the patch to current master? We just removed hhvm from .travis.yml.

NamelessCoder:
Whomever merges this: don't forget to squash-and-merge and fix the commit casing; `[TASK] Use (float) and (int)` (uppercase "U")

NamelessCoder:
FYI to all maintainers: I've enabled "Branch must be up to date" which means there's now an "Update branch" button - so we don't have to ask for rebases. Don't forget to always squash-and-merge after that so we avoid the merge commits ;)

DanielRuf:
> FYI to all maintainers: I've enabled "Branch must be up to date" which means there's now an "Update branch" button - so we don't have to ask for rebases. Don't forget to always squash-and-merge after that so we avoid the merge commits ;)

Cool, thanks =)

PR #314, opened 2017-07-01
[BUGFIX] use the Trusty build environment

DanielRuf:
This fixes the Travis build for HHVM. HHVM is available on Trusty but not Precise anymore.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: trusty
 language: php
 
 branches:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 dist: trusty
+
 language: php
 
 branches:


NamelessCoder:
Thanks Daniel - I'll re-label this as a `[TASK]` since it only affects the test build.

DanielRuf:
> Thanks Daniel - I'll re-label this as a [TASK] since it only affects the test build.

You are welcome.

Is there some overview of the right usage of the labels for contributors (eg some CONTRIBUTING.md file)?

NamelessCoder:
We're following a reduced set of the TYPO3 commit CGL (code is plain PSR, not many surprises there except for how we like certain whitespace cosmetics). Short version:

* `[TASK]` is used when the change for example doesn't change active code, or when the code being changed is considered a formality (e.g. cleanups).
* `[BUGFIX]` is used when it fixes a bug that a user would experience.
* `[FEATURE]` is for newly added features (basically, new public API).
* `[DOC]` is of course used when the change is purely about documentation. When the change is (mostly) about PHPDOC changes we usually use this one but sometimes `[TASK]` depending on how public API is.

It isn't applied yet but the plan is to automate change log generation based on these prefixes to give a filtered list of relevant changes that don't include our internal CI stuff etc.

DanielRuf:
Ok, I understand. That makes much sense.

PR #316, opened 2017-07-02
[TASK] require and include are statements

DanielRuf:
`require` and `include` are statements, not functions. Therefore the parentheses are not needed.


NamelessCoder:
Commit message style was missed on this one. It happens, but please be careful and use squash-merge to edit ;)

DanielRuf:
Ah, so it should be the same as in the PR title? Sorry, missed that. Next time I know it =)

PR #318, opened 2017-07-11
Update FLUID_VIEWHELPERS.md

Weissheiten:
[TASK-172] I propose to edit the example in the documentation. At the moment following this example will produce the error:
Notice: Undefined index: src/ViewHelper\ in D:\TYPO3Fluid\vendor\typo3\fluid-schema-generator\bin\generateschema on line 40
Directory "" does not exist but is declared in composer autoload entries
due to the change in the command line tool (a path to the ViewHelpers directory is considered a second namespace which cannot be found and results in this rather cryptic output for end users)

### DIFF
--- a/doc/FLUID_VIEWHELPERS.md
+++ b/doc/FLUID_VIEWHELPERS.md
@@ -105,11 +105,10 @@ When installed with development dependencies, `TYPO3.Fluid` includes a CLI comma
 native ViewHelpers and any inside your own packages. To use this command:
 
 ```bash
-./vendor/bin/generateschema TYPO3Fluid\\Fluid\\ViewHelpers src/ViewHelpers > schema.xsd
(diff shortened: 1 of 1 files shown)


NamelessCoder:
Thanks for catching this one. Actually, that entire section needs to be rewritten or replaced with a single link to the https://github.com/TYPO3/Fluid.SchemaGenerator package. It is also no longer true that the command automatically is installed when you install with dev dependencies.

lolli42:
The change itself looks good, though and could be accepted. Rewriting the section later could be another patch. I'd tend to approve this one. Claus?

Weissheiten:
just re-read it when the comment came in and spotted a typo though ("in the schame"). As I used the information under https://github.com/TYPO3/Fluid.SchemaGenerator , the best way might be not to accept the change, replace the section with the link for the moment as Claus suggested and fix the documentation at the links location.

lolli42:
ok, so we close this PR.

PR #321, opened 2017-07-26
[BUGFIX] Use sanitized identifier in TemplateCompiler::has()

sascha-egerer:
Instances are stored with the sanitzied identifier into a runtime cache.
The `TemplateCompiler::has()` must also use the sanitzed version of the
identifier to check if the instance exists in the runtime cache.

### DIFF
--- a/src/Core/Compiler/TemplateCompiler.php
+++ b/src/Core/Compiler/TemplateCompiler.php
@@ -142,13 +142,14 @@ class TemplateCompiler
      */
     public function has($identifier)
     {
+        $identifier = $this->sanitizeIdentifier($identifier);
+
         if (isset($this->syntaxTreeInstanceCache[$identifier])) {
             return true;
         }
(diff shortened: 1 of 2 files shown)


NamelessCoder:
Expectations in unit tests need to be adjusted, two of them don't expect sanitizeIdentifier to be called. Obviously, now they should expect it :)

Apart from that, looks good and would prevent possible issues with custom identifiers.

NamelessCoder:
We missed the `sanizted` typo in the commit message - I've noticed with GitHub that if you edit the PR title here, it will be used as suggested squash commit title, but only if:

* The PR page was reloaded inbetween; the XHR updates don't seem to catch this.
* There's more than one commit in the PR

If either of those are not true, the commit message must be manually edited right before merging even if the PR title was changed. Not a biggie - just a small peculiarity to watch out for.

sascha-egerer:
Oh damn... But good to know

PR #324, opened 2017-07-28
[FEATURE] add and/or alternative for ||/&& syntax

### LINKED ISSUE #319: Support "and" in addition to "&&" in boolen expressions
The boolean expression parser only allows `&&` for AND conditions - https://github.com/TYPO3/Fluid/blob/master/src/Core/Parser/BooleanParser.php#L202

When using `&&` in attribute values, the resulting files are not well-formed anymore because in XHTML each `&` must either start an entity, or be escaped as `&amp;`.

A solution would be to add `and` as a synonym for `&&`, which would allow us to write well-formed HTML templates that can be validated with `xmllint`.

mneuhaus:
With this commit the user can choose to use „and“ instead of „&&“ and „or“ instead of „||“ in boolean conditions

closes #319

### DIFF
--- a/src/Core/Parser/BooleanParser.php
+++ b/src/Core/Parser/BooleanParser.php
@@ -74,7 +74,11 @@ class BooleanParser
 			|
 				\|\|
 			|
+			    [aA][nN][dD]
+			|
 				&&
+			|
+			    [oO][rR]
 			|
 				.?
 			)\s*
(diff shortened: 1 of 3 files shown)


tmotyl:
I don't really see a benefit of this feature. It just makes things more complicated and slower. Especially as developers are used to use only || and && in PHP because of our CGL.

NamelessCoder:
> Especially as developers are used to use only || and && in PHP because of our CGL.

In TYPO3 CMS's CGL, yes. But we do not apply our framework coding style preference to everyone else and in this case, it solves a real problem: `&&` reports as invalid XML syntax but `AND` does not, so enabling this allows well-formed XML. And logically if we allow one, we should also allow the other.

Re: slower this would only be a marginal difference if there is one at all (the difference between several conditions and one in_array + strtolower plus a few more bytes of regexp). The source code becomes a bit more complicated, sure, but the user experience becomes less complicated (both work as opposed to having to remember which one is valid).

tmotyl:
ok, making it correct xml is a valid argument :)

PR #329, opened 2017-07-30
[FEATURE] Support Vendor.Package as namespace name

NamelessCoder:
Using this format (or simply Package if no vendor name is
used for the package) makes ViewHelperResolver look in
that namespace + ViewHelpers, for example:

```xml
<TYPO3Fluid.Fluid:render partial="Partial" />
```

Will internally resolve to:

```php
\TYPO3Fluid\Fluid\ViewHelpers\RenderViewHelper
```

Without the need to register the namespace. Note that the
example uses the native Fluid namespace to demonstrate,
although this namespace is always present (unless explicitly
removed by a TemplatePaths override).


NamelessCoder:
Needs https://github.com/TYPO3/Fluid/pull/328 before tests will succeed. Will need to be conflict resolved and rebased.

lolli42:
Like the patch, tried to rebase it, but didn't succeed, got quite some unit test failures after a straight conflict resolving. Could you have a look at this, Claus?

NamelessCoder:
Thanks for reviewing / testing, Christian :)

This last force-push should fix the issue. The error messages were slightly different and there was a PHP 5.x problem as well.

NamelessCoder:
Fixed a minor commit message problem (TemplatePaths was mentioned as the place to change ViewHelper namespaces, I guess I was thinking about something else while writing that).

mbrodala:
I'd say yes since there is no rule that a namespace must use `UpperCamelCase`, see for example mikey179/vfsStream which uses `org\bovigo\vfs`.

PR #330, opened 2017-07-30
[TASK] Expect explicit exception type in unit test

NamelessCoder:


lolli42:
This patch is somehow bogus: the "subject" argument is registered as "array" in ->registerArgument(), so the \InvalidArgumentException is currently thrown by ->validateArguments() from ->initializeArgumentsAndRender() already. In normal code flow, the Exception testing for array or countable in renderStatic() can't be reached, and it is afaik not possible to register an argument with "array or object" or similar?

NamelessCoder:
Being addressed as a more thorough upgrade of phpunit itself, which has a different way of handling expected exceptions. TBH I still don't know why this seemingly innocuous change caused the type of errors you reported, but we haven't seen a trace of them since or anywhere else, which is good.

PR #332, opened 2017-07-30
[REMOVAL] Deprecate legacy ViewHelper attributes

NamelessCoder:
This patch deprecates:

* $this->childNodes (access through $this->viewHelperNode)
* $this->templateVariableContainer (access through $this->renderingContext)
* $this->viewHelperVariableContainer (access through $this->renderingContext)
* method validateArguments() - to be called from dedicated object
* method resetState() - has not been called for a good while
* method initialize() - should be handled by framework lifecycle methods or constructor/render call
* method setChildNodes() - no longer needed; access through $this->viewHelperNode

The public methods of the interface are moved to a separate
interface that is implemented by the base ViewHelper class.

The deprecation itself is done by implementing a new Trait and
companion interface which moves all deprecated routines
away from the abstract ViewHelper class. Further refactoring
is pending to move additional methods here.

The interface and trait can later be implemented by third party
ViewHelpers which require cross-version compatibility and
removal from the base class can be done by simply removing
the interface and trait.

NB: validation logic is moved here although the refactoring
that will finally remove the call to validateArguments() is not
yet submitted as PR.

### DIFF
--- a/src/Core/ViewHelper/AbstractViewHelper.php
+++ b/src/Core/ViewHelper/AbstractViewHelper.php
@@ -8,19 +8,23 @@ namespace TYPO3Fluid\Fluid\Core\ViewHelper;
 
 use TYPO3Fluid\Fluid\Core\Compiler\TemplateCompiler;
 use TYPO3Fluid\Fluid\Core\Parser;
+use TYPO3Fluid\Fluid\Core\Event\EventInterface;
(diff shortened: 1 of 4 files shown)


NamelessCoder:
Will be almost completely redundant after https://github.com/TYPO3/Fluid/pull/457 - and work following that PR will make it completely redundant. Closing.

PR #339, opened 2017-09-13
Update FLUID_VIEWHELPERS.md

kftw:
`$view->getTemplatePaths()->addNamespace()` seems to belong to an older version?


NamelessCoder:
Nice catch - just a typo, the method was never on TemplatePaths :)

Issue #96, opened 2016-04-01
RenderingContext does not use view


kitsunet:
Hi,

slowly starting to get into the code and I wonder why this is done (why is $view a dependency of the RenderingContext):
https://github.com/TYPO3Fluid/Fluid/blob/master/src/Core/Rendering/RenderingContext.php#L121

I don't see a getter nor any usage of $this->view in there. So IMHO it just complicates the creation process as the View depends on the Context and this cross dependency is not really nice.

NamelessCoder:
Hi Christian,

Thanks for looking into the code.

I agree that it's less than ideal that we have View on the RenderingContext. There is a reason though - as always - but an alternative solution was never created/discussed and could potentially require a breaking change.

The reason for having View on RenderingContext right now is that we need access to the View instance from `f:render` in order to call public methods to render partials and/or sections. No other class is currently aware of the View which is why RenderingContext has the instance. One of the reasons for making the constructor require the View but create it if not provided, is that we encourage custom RenderingContext classes which may be (if programmer decides) coupled to a specific View class.

On a side note, this was also the case in Flow's Fluid as well as CMS Fluid - although on those, the constructor would not create the instance (instead, template rendering would break if `f:render` was used and `setView()` had not been called on RenderingContext).

It might be possible to avoid the constructor creation but that would put this instance preparation requirement on another place and make the RenderingContext harder to create (like it was before, but now the RenderingContext has a much more prominent role).

The alternative might be to move it to ParsedTemplate/ParsingState, but I'm not sure without checking, whether or not that instance is consistently available to `f:render` in compiled and uncompiled modes alike.

Suggestions are welcome! If any part of the reasoning above needs further explaining please fire away ;)

kitsunet:
Hey, wanted to come back to this. The current codebase uses `$this->viewHelperVariableContainer->getView()` inside the `RenderViewHelper` and this view again is set `\TYPO3Fluid\Fluid\View\AbstractTemplateView::initializeRenderingContext`, via `$this->baseRenderingContext->getViewHelperVariableContainer()->setView($this);` so I think the view property inside the `RenderingContext` itself is definitely unused and could be removed making this contruct much nicer to handle IMHO.

NamelessCoder:
[...]

PR #340, opened 2017-10-10
[TASK] Remove dependency on View in RenderingContext

NamelessCoder:
For a while there has been a near-circular dependency between
RenderingContext and View. The only component that truly
depends on the View is the ViewHelperVariableContainer, so
removing this interdependency is fairly painless.

A few key points:

* This remains compatible with method signatures in overridden
   RenderingContext classes that still require the view. Such classes
   can simply remove this constructor argument once the minimum
   fluid requirement is raised to a version with this patch included.
* The ViewHelperVariableContainer->getView method may now
   return NULL if a Fluid implementation overrides the normal View
   initialization logic (which sets the View in ViewHelperVariableContainer).
   This is not an issue since the method is already clearly annotated
   as "not public, may change" - and now it has changed. For a very
   limited set of ViewHelpers which actually use this internal API, this
   could be a breaking change.
* One such ViewHelper exists in Fluid - f:render. This patch changes
   that ViewHelper so it throws a ViewHelper Exception if detecting a
   ViewHelperVariableContainer without a View reference. This then
   reports a very precise cause and instructions for fixing.
* Tested and confirmed with TYPO3 dev-master, and third party app.

Close: #96

### DIFF
--- a/src/Core/Rendering/RenderingContext.php
+++ b/src/Core/Rendering/RenderingContext.php
@@ -26,7 +26,6 @@ use TYPO3Fluid\Fluid\Core\ViewHelper\ViewHelperInvoker;
 use TYPO3Fluid\Fluid\Core\ViewHelper\ViewHelperResolver;
 use TYPO3Fluid\Fluid\Core\ViewHelper\ViewHelperVariableContainer;
 use TYPO3Fluid\Fluid\View\TemplatePaths;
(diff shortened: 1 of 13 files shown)


NamelessCoder:
Replaced by https://github.com/TYPO3/Fluid/pull/548 and https://github.com/TYPO3/Fluid/pull/549.

PR #344, opened 2017-10-31
[TASK] Call getter instead of accessing property

NamelessCoder:
Allows class overrides to only override getRootNode
on ParsingState in order to change which node gets
returned, instead of also having to override the
render() method. Does not affect behavior.


NamelessCoder:
Just for anyone reading this: nothing was merged, it appears GitHub will use this message if a commit gets removed and branches become in sync. Will reopen this as new pull request.

opi99:
Looks good to me

PR #347, opened 2017-11-10
[BUGFIX] Respect $ignoreUnknown when rendering section/partial

NamelessCoder:
Somewhere, respect for $ignoreUnknown was lost. This patch
adds a check and only throws the exception if it is false.

### DIFF
--- a/examples/Resources/Private/Singles/ErrorHandling.html
+++ b/examples/Resources/Private/Singles/ErrorHandling.html
@@ -5,6 +5,7 @@
 <!-- View errors; rendering an asset that doesn't exist -->
 <f:render partial="DoesNotExist" />
 <f:render section="DoesNotExist" />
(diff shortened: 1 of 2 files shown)


sascha-egerer:
Tests need to be adjusted, right?

NamelessCoder:
Yes; there are a couple of tests to adjust before merging.

NamelessCoder:
Patch is now ready to be considered. Had to specifically catch both template file and section missing exceptions and check $ignoreUnknown to prevent falling through to catching of generic View exceptions.

opi99:
Code looks good for me ... where can I find the changes in the tests?

NamelessCoder:
FYI: there are no changes in the tests as these specific cases are unfortunately not covered (the unit tests that do test for optional, don't test the ViewHelper). The code does get executed and is confirmed in production - but there are no assertions on that behavior to try/catch in the VH itself.

PR #348, opened 2017-11-13
[BUGFIX] Make template file resolving recursive

NamelessCoder:
Fixes resolving of files in subfolders, among other
things making all resolveAvailableXyz methods
return all files in subfolders.

### DIFF
--- a/src/View/TemplatePaths.php
+++ b/src/View/TemplatePaths.php
@@ -327,10 +327,14 @@ class TemplatePaths
         if (!is_dir($folder)) {
             return [];
         }
+
         $directoryIterator = new \RecursiveDirectoryIterator($folder, \FilesystemIterator::FOLLOW_SYMLINKS | \FilesystemIterator::SKIP_DOTS);
-        $filterIterator = new \RecursiveRegexIterator($directoryIterator, '/^.+\.' . $format . '$/i', \RecursiveRegexIterator::GET_MATCH);
-        $recursiveIterator = new \RecursiveIteratorIterator($filterIterator, \RecursiveIteratorIterator::SELF_FIRST);
-        return array_map('strval', array_column(iterator_to_array($recursiveIterator), 0));
+        $recursiveIterator = new \RecursiveIteratorIterator($directoryIterator, \RecursiveIteratorIterator::SELF_FIRST);
+        $filterIterator = new \CallbackFilterIterator($recursiveIterator, function($current, $key, $iterator) use ($format) {
+            return $current->getExtension() === $format;
+        });
+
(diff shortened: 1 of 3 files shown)


NamelessCoder:
Don't merge - will amend with recursion limit to prevent infinite loops in case someone made some unfortunate symlinks.

NamelessCoder:
Updated to use a filtered recursive directory iterator. This iterator has built-in infinite symlink protection - tested and confirmed.

opi99:
Code change looks good from my site, any way to test this change inside the cache warm up?

NamelessCoder:
You can test this with cache warmup, if you have a partial template in a subfolder then without this change, you'd see no compiled template for it - with it, you'd see a compiled class named like the partial (with an underscore to replace the slash).

NamelessCoder:
@mbrodala If this still builds green and you find the patch okay would you mind merging this one?

mbrodala:
I'll try to add a unit test first.

mbrodala:
@NamelessCoder Can you first rebase onto the latest master?

src/View/TemplatePaths.php
```diff
@@ -325,8 +325,13 @@ protected function resolveFilesInFolders(array $folders, $format)
      */
     protected function resolveFilesInFolder($folder, $format)
     {
-        $files = glob($folder . '*.' . $format);
-        return !$files ? [] : $files;
+        if (!is_dir($folder)) {
+            return [];
+        }
+        $directoryIterator = new \RecursiveDirectoryIterator($folder, \FilesystemIterator::FOLLOW_SYMLINKS | \FilesystemIterator::SKIP_DOTS);
```
wmdbMattes: Would it be viable to run Symfony Finder here?

NamelessCoder: Except for Fluid not having any dependencies at all right now, no. I would be happy to reconsider this if we begin adding dependencies, but I'd not do that until we have to. IMHO it doesn't save enough code in this case to justify.

wmdbMattes: jup, that's reasonable

NamelessCoder:
@mbrodala While a unit test surely would be appreciated, I don't consider it a problem if a patch that adds the test comes later. By now we have a few uncovered places so a general update to tests is required at some point anyway. We've got tests aplenty to ensure that non-recursive resolving works ;)

mbrodala:
The test is green now, please have another look. If you're fine, I'll merge this.

NamelessCoder:
Test looks good to me - would work the same with any of the three resource types. Thanks for taking care Mathias!

PR #360, opened 2017-12-22
[TASK] Support HTTPS typo3.org namespace

### LINKED ISSUE #359: Support HTTPS for xmlns ViewHelper NS registration
It took me some time to figure out why my ViewHelper didn't render until i debugged and found out `\TYPO3Fluid\Fluid\Core\Parser\TemplateProcessor\NamespaceDetectionTemplateProcessor::registerNamespacesFromTemplateSource` (Line 108) does only check for `http://typo3.org/ns/` but not `https://typo3.org/ns/`.
It would be really nice to have the HTTPS scheme supported :)

mbrodala:
Fixes #359

### DIFF
--- a/src/Core/Parser/TemplateProcessor/NamespaceDetectionTemplateProcessor.php
+++ b/src/Core/Parser/TemplateProcessor/NamespaceDetectionTemplateProcessor.php
@@ -105,8 +105,9 @@ class NamespaceDetectionTemplateProcessor implements TemplateProcessorInterface
             preg_match_all('/' . $namespacePattern . '/', $matches[0], $namespaces, PREG_SET_ORDER);
             foreach ($namespaces as $set) {
                 $namespaceUrl = trim($set[2], '"\'');
-                if (strpos($namespaceUrl, 'http://typo3.org/ns/') === 0) {
-                    $namespaceUri = substr($namespaceUrl, 20);
+                $namespaceUrlMatches = [];
+                if (preg_match('~^(https?://typo3\.org/ns/)~', $namespaceUrl, $namespaceUrlMatches)) {
+                    $namespaceUri = substr($namespaceUrl, strlen($namespaceUrlMatches[0]));
                     $namespacePhp = str_replace('/', '\\', $namespaceUri);
                 } elseif (!preg_match('/([^a-z0-9_\\\\]+)/i', $namespaceUrl)) {
                     $namespacePhp = $namespaceUrl;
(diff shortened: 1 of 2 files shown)


NamelessCoder:
Just to make it clear: using http or https has no effect whatsoever for XML namespace URIs. In fact, these are not URIs which are supposed to be fetched - and serving any content on them points to an incorrect understanding of what the NS URI actually is. Furthermore: when you generate an XSD schema, it is fitted with the exact NS URI in order to make the match between what's in the XML NS and the XSD file itself. If these don't match, the standard behavior is to not use the XSD file (though some editors do make the connection regardless, or have a manually configured URI-to-XSD mapping).

So with all that in mind, I don't think it's a good idea to make this change unless we completely and utterly switch every last namespace in every last XSD over to the https protocol.

This is, in effect, the same as changing a persisted ID and then needing to change every reference to it.

mbrodala:
Personally I understand the concept of namespaces here but I assume that sooner or later everything will be switched to HTTPS for consistency. And even if one is not supposed to call these URLs today, who knows if this will change in the future.

NamelessCoder:
Don't get me wrong, I don't mind the added support personally - but that's because I know exactly how XSD mapping works. If I take on the critical view and look at this from a user's perspective:

* At best, it adds confusion about whether to use http or https in the namespace URI in templates.
* At worst, it tempts users to use the https flavor now and not be aware that this will ruin the mapping between template and XSD file.
* There are no positive effects from this except for perceived consistency, since these aren't URLs.

> And even if one is not supposed to call these URLs today, who knows if this will change in the future.

I've been considering this for years and years, ever since starting working with Fluid. Not to dive into all the details here, but the main challenge is: what to serve on that URL? You can't serve the XSD that belongs to the namespace, unless you version the URL so the fetch gets the right version of the XSD. And if you do, this introduces a crazy amount of inconsistencies and required XSD mappings: every template file might reference a different versioned NS URI, causing complete chaos. A single TYPO3 site could depend on as many as 30+ different Fluid schemas and you'd need to map all of them.

That is, in short, why the XSD namespace URI is not fetchable and in my opinion should never become fetchable. It's simply a technical version mess concern combined with the fact that these aren't supposed to be actual, real URLs according to the very nature of XML so there's no good motive to make them serve content. It doesn't help our users to begin serving something on these, it merely adds to the confusion.

So even though this change looks completely innocent and I personally have no problem with it, I would still recommend abandoning it until such time where we make this switch completely - and prepare ourselves for the possibility that these URIs might never change qua being a distributed, persisted ID that doesn't have any security concerns applied to it. Changing a NS url scheme to achieve consistency is, imho, far too breaking to have merit. If we want to actually do this, we instead should invent a new place for namespaces and have it *require* https.

Hopefully you understand my concerns? I guess TL;DR is: this change is the same as allowing users to enter an invalid NS ID without causing an error.

mbrodala:
Thanks for the elaborate explanation why this change doesn't make sense. :-)

I'll abandon the change and update the related ticket.

NamelessCoder:
Using https will make perfect sense some time in the future, I'm sure - but in another form: by adding a new official, https-carrying NS URI / prefix, and deprecating the old. Maybe even making it more agnostic about domain (letting users use their own) and using the path or other, to identify an NS as Fluid NS. No doubt there are smarter ways to do this and if you come up with ideas I'm happy to hear them :)

PR #364, opened 2018-01-08
2.3

opi99:


liayn:
ehm... ;-) ?

NamelessCoder:
Contains individually travis-tested cherry-picks to create a 2.3 branch and a 2.3.5 release.

liayn:
Ah.. a few words in the PR would have helped.
😉

liayn:
Please include https://github.com/TYPO3/Fluid/pull/367 as well here

opi99:
We can add #367 also after merging this in the 2.3 branch

helhum:
Can somebody explain to me, why we need a 2.3 branch in fluid?

NamelessCoder:
@helhum Sure - we "need" this branch so we can cherry-pick and tag additional 2.3.x releases of Fluid. The reason for this is as a service to packages depending on Fluid, such as TYPO3: by creating new 2.3.x releases we allow TYPO3 and others to keep their `^2.3` minimum dependency and not bump to `^2.4`. Users can in turn restrict their projects to avoid the features added by Fluid 2.4 until they choose to add them.

This was also done in the past for a limited time, which is why we have other similar branches. I'm hoping that with a bit more help in maintenance we can resume creating such bug-fix releases for a couple versions back.

I think to most people and projects it doesn't matter much because the versions are semantic and thus non-breaking. For me personally it's fine to only tag in most recent compatible version but I can see why some might prefer to keep a strict `2.3.x`-only policy.

lolli42:
the branch is there already. i think having these patches merged into the branch is ok.
for typo3 core v8, we switch to 2.5.0. still, that PR won't harm.

Issue #366, opened 2018-01-15
Log is spammed with is_dir(): open_basedir restriction warnings


liayn:
```
PHP Warning: is_dir(): open_basedir restriction in effect. File(/) is not within the allowed path(s): (....) in /.../typo3_src-8.7.9/vendor/typo3fluid/fluid/src/View/TemplatePaths.php line 411
```

The reason is that the configuration of the extbase plugin's view look like this:
`plugin.tx_ext.view.templateRootPaths.10 = {$someEmptyConst}`

This results in `\TYPO3Fluid\Fluid\View\TemplatePaths::sanitizePath()` receiving `$path === ''` as parameter. This is furthermore converted to a `/` and a check with `is_dir('/')` is performed, which results in the above warning.

The proper solution would be to "reject" an empty path as it needs no sanitizing.

liayn:
Sorry, incorrect report. Will improve it.

liayn:
The actual issue is that an empty path is excepted and causes issues in TYPO3 core afterwards.
`\TYPO3\CMS\Fluid\View\TemplatePaths::getContextSpecificViewConfiguration`

My proposal is to fix this issue in Fluid, which should skip empty paths in general, as those do not make sense.

NamelessCoder:
Fixed by https://github.com/TYPO3/Fluid/commit/becacd0d58fcd75ae804816c66d1f2cb4667ebf8

PR #367, opened 2018-01-15
[BUGFIX] Do not append slash to empty paths (#366)

liayn:


NamelessCoder:
Saw the issue now. I think the right solution is to not perform sanitisation, but also to not remove the path, if it is empty. E.g. early return empty string from `sanitizePath` if path is empty.

liayn:
"entry script's directory"

How come? Empty means current dir? Hmmm. That's what we have a dot for.

Empty dirs must be ignored, since invalid.

NamelessCoder:
> Empty dirs must be ignored, since invalid.

I'm trying to tell you they are _not_ invalid...

```php
$path = realpath('');
// $path contains for example "/Users/claus"
```

So you are incorrect in your assertion. Sorry.

opi99:
`realpath('')` on unix systems returns the actual path (or namely working directory) (mostly the one where you start the script). So it is a bit unpredictable where you really are.

So I also wouldn't take an empty path as correct setting. (And I also don't understand why realpath was changed at this point in PHP 5.2).

NamelessCoder:
> Configuring an empty path is nothing but a mistake by the dev.

Opinion. Hyperbole.

> Ignore the path, since it is invalid

Opinion. Disregards my previous feedback.

> throw an exception telling that one just fed bullsh** into Fluid.

Would constitute a breaking change. Opinion. Not bullshit, as I've tried explaining.

Why did I mention `realpath()` behavior? - because it's used in TemplatePaths when ensuring an absolute (directory) path: https://github.com/TYPO3/Fluid/blob/master/src/View/TemplatePaths.php#L439

Why do I insist empty paths are valid? - because there is explicit support for empty paths in TemplatePaths: https://github.com/TYPO3/Fluid/blob/master/src/View/TemplatePaths.php#L409

I am not just making these things up when I say that you considering an empty path invalid, *is a misunderstanding of what Fluid is supposed to support!*. You need to tone this opinion and hyperbole down a bit, would be my advise. I'm listening to everything you say, but when I say you're wrong, you simply ignore me. This is not creating a good environment for cooperation. TYPO3 CMS is not the only use case for Fluid.

> the actual issue here is, that this empty string is converted to `/`

Aha! See, now we're getting somewhere! So the question is: why does that happen? In TemplatePaths in Fluid itself, there are guards against slash-suffixing paths which are not directories, which are empty (yet another explicit empty support), or which already start with a slash. This behavior could be buggy.

Or could it be that [TemplatePaths of TYPO3 CMS does something the vanilla template paths does not, and neglects to test for an empty value](https://github.com/TYPO3/TYPO3.CMS/blob/master/typo3/sysext/fluid/Classes/View/TemplatePaths.php#L205)?

Seriously, Markus. I apologize for the tone in this reply but you don't leave me much wiggle room when everything you write is formulated as opinion and hyperbole, disregards my previous comments and keeps insisting that behavior-by-design in Fluid is wrong, just wrong, because it doesn't work with TYPO3 CMS. I request that we treat each other with a bit more respect and try to actually listen to what is being said rather than just impose that what we believe is right, must be right for everyone else.

liayn:
Claus. You know that I don't mean anything personal, and as always, I appreciate your efforts!
Sorry, for keeping communication short, quick and fast, might sound a bit too rough.

I still disagree on the "feature" of empty paths. I still stick to my opinion that this is a bad design - explicit vs implicit.

So, no matter what. This needs to be fixed as this is embarrassing.

NamelessCoder:
I sincerely do not take this personally and nor should you ;) I am merely trying to say that these formulations with hyperbole is a bad idea and that in this case, there's actually intention behind it, and good reason to believe that the *real* problem is in fact hidden in how TYPO3 CMS overrides TemplatePaths. Simply put, it doesn't filter out and prevent absolute-path'ing empty paths, even though by your own problem description (TS with unresolvable constant) is a TYPO3-explicit use case.

Sometimes we have to look for the fixes in our own implementations rather than the parent library. This is, I believe, one of those cases. My advise is to fix the TemplatePaths implement in TYO3 CMS by making the method I linked to last above, filter out or not absolute-convert when the path is empty (whatever the reason may be).

NamelessCoder:
Oh, by the way I forgot to mention this: the reason empty paths are allowed in TemplatePaths in Fluid is actually in part due to how TYPO3 CMS works. If you configure a path in TS that cannot be resolved, the resulting path also becomes empty (e.g. EXT: reference points to non-existing dir, false gets returned). Which just further indicates that such filtering is the responsibility of TYPO3 CMS - as it too was historically passing invalid paths to Fluid due to uncritically passing output from getFileAbsFilename et al.

liayn:
> (TS with unresolvable constant)

to be precise. This the constant is there, but empty. The typical usecase where you allow integrators to define a custom path for templates for the extension via TS constant.

I'll fiddle around now and try to find out where the absolute path stuff really comes from. I believe my last finding was that it was Fluid itself, that's why I proposed a patch here. But I'll take a second look now, to be really sure.

liayn:
Okay, the whole stacktrace now says:

- `\TYPO3\CMS\Fluid\View\TemplatePaths::fillDefaultsByPackageName` retrieves the configuration with the empty patch
- `\TYPO3Fluid\Fluid\View\TemplatePaths::fillFromConfigurationArray` receives the empty path
- `\TYPO3Fluid\Fluid\View\TemplatePaths::extractPathArrays` is called within `fillFromConfigurationArray`, which itself calls
- `\TYPO3Fluid\Fluid\View\TemplatePaths::ensureSuffixedPath`, which finally appends the slash to the empty path.

So to my judgment this is broken in Fluid and any software using Fluid with empty paths will run into the same issue.

liayn:
Found another possible solution, which does not require test adjustments. (even though this points out that we actually lack a test for this usecase)

NamelessCoder:
Looks like the right fix! If you change the commit title / PR title I will accept-review.

liayn:
Would this be okay?

`[BUGFIX] Do not append slash to empty paths (#366)`

NamelessCoder:
Perfect :)

PR #368, opened 2018-02-05
[TASK] Update deprecated license for Packagist

NamelessCoder:
Updates the composer manifest license value to the
current way of saying LGPL 3 or later.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,7 +1,7 @@
 {
 	"name": "typo3fluid/fluid",
 	"description": "The TYPO3 Fluid template rendering engine",
-	"license": ["LGPL-3.0"],
+	"license": ["LGPL-3.0-or-later"],
 	"require": {
 		"php": ">=5.5.0"
 	},


composer.json
```diff
@@ -1,7 +1,7 @@
 {
 	"name": "typo3fluid/fluid",
 	"description": "The TYPO3 Fluid template rendering engine",
-	"license": ["LGPL-3.0"],
+	"license": ["LGPL-3.0-or-later"],
```
mbrodala: Isn't this technically a license change? Before it was _GNU Lesser General Public License v3.0 only_, with this it's _GNU Lesser General Public License v3.0 or later_.

NamelessCoder: Technically, the license is inherited from TYPO3 CMS which if I'm not mistaken uses v2-or-later? Unless that's a misunderstanding I would allow the same "or later" for Fluid.

lolli42: Note the LGPL came from flow. And yes "LGPL-3.0" is deprecated, it should be either "LGPL-3.0-only" or "LGPL-3.0-or-later" (according to https://spdx.org/licenses/) - i'd be fine with -or-later, and i think we could just pick the patch as as.

mbrodala:
I think we're fine here, see https://github.com/composer/composer/issues/7026#issuecomment-363034233.

Issue #300, opened 2017-06-10
MathExpressionNode inside f:alias or f:for doesn't find local variables


christoph-bessei:
I tried to use a math expression inside f:alias and f:for, but always get an "A non-numeric value encountered" error. 
The problem seems to be the rendering context in [MathExpressionNode](https://github.com/TYPO3/Fluid/blob/055303e18659978ac991885d9c8f6407fa4c6e63/src/Core/Parser/SyntaxTree/Expression/MathExpressionNode.php#L51), which doesn't know the variables set by f:for or f:alias but only the global variables. Since it can't find the variable it returns the variable name, which is not a numeric value.

Example:
```html
<f:alias map="{itemCount: '10'}">
    {itemCount / 2}
</f:alias>

<f:for each="{0:'10',1:'5'}" as="itemCount">
    {itemCount / 2}
</f:for>
```

I'm using fluid inside TYPO3 8.7.1 (composer install), so it could be a TYPO3 problem.
Is there an easy way to check if this also occurs in fluid standalone?

liayn:
We encountered a similar issue with the following construction

```
	<f:for each="{0: 1, 1: 2, 2: 3}" as="value" iteration="it">
		<f:alias map="{nextIndex: '{it.index + 1}'}">
			{nextIndex} foo<br />
		</f:alias>
	</f:for>
```

This caused an error where "it.index" is
> A non-numeric value encountered

The preconditions for this to fail is PHP 7.1 + typo3 development-preset.

Reason this fails is because:
`\TYPO3Fluid\Fluid\Core\Parser\TemplateParser::textAndShorthandSyntaxHandler` calls `$expressionNode->evaluate($this->renderingContext);`, where `$expressionNode` is `\TYPO3Fluid\Fluid\Core\Parser\SyntaxTree\Expression\MathExpressionNode`.
At this parsing stage there are no variables in the context to resolve hence it tries to execute this PHP statement: `"it.index" + "1"`, which is obviously invalid in PHP 7.1. (in lower versions this results in `1`)

Moreover also the `"1"` being a string is critical during actual execution as it would trigger the same PHP warning.

Steps to do therefore are:
1.) Reconsider whether it is clever to run the evaluation during parsing
2.) Make `\TYPO3Fluid\Fluid\Core\Parser\SyntaxTree\Expression\MathExpressionNode::evaluateOperation` type aware and validate the parameters and cast them accordingly.

NamelessCoder:
[...]

PR #369, opened 2018-02-05
[BUGFIX] Prevent evaluating ExpressionNode unless demanded

NamelessCoder:
Adds a new interface which must be used by expression node
types in order to signal that the expression node must be
evaluated during parsing of the template.

History:

Parse-time evaluation of ExpressionNodes is a remnant from
the time when there was a separate legacy namespace
extraction based on expression nodes. While there currently
are expression-like template processors, there are no known
implementations out there which must have the node evaluated
during parse time.

Choices:

Rather than simply remove the evaluation altogether, the new
interface is added as a quicker way to migrate *if* there are
private repositories out there implementing these nodes.
If the evaluation was removed entirely, the only alternative left
would be migrating to a template pre-processor which is by
nature quite different from the expression nodes. So, as a
convenience, this interface is left in place to be implemented
as substitute for the old interface, requiring no further change.

Close: #300


NamelessCoder:
Alternatives:

1. Demand refactoring old expression nodes which depend on parse-time evaluation.
2. Add this interface as-is and consider parse-time evaluation a (very infrequently used) existing feature.
3. Add this interface, but deprecate it immediately, and inform that parse-time evaluation will be removed in Fluid 3.0

I lean towards 1 or 3. Parse-time evaluation is the domain of template pre-processors. Any thoughts?

liayn:
Option 3 sounds good

NamelessCoder:
Code updated to reflect option 3.

src/Core/Parser/SyntaxTree/Expression/ParseTimeEvaluatedExpressionNodeInterface.php
```diff
@@ -0,0 +1,30 @@
...
+namespace TYPO3Fluid\Fluid\Core\Parser\SyntaxTree\Expression;
+
+/*
+ * This file belongs to the package "TYPO3 Fluid".
+ * See LICENSE.txt that was shipped with this package.
+ */
+
+use TYPO3Fluid\Fluid\Core\Compiler\TemplateCompiler;
+use TYPO3Fluid\Fluid\Core\Parser\SyntaxTree\NodeInterface;
+use TYPO3Fluid\Fluid\Core\Rendering\RenderingContextInterface;
```
liayn: unused?

PR #370, opened 2018-02-08
Add tests for objects with countable-interface

DrillSergeant:
While updating TYPO3 to 8.7.10 I came accross some errors in fluid. Seems it does not handle countable objects correctly when used in a conjunction (&& or ||). Also it seems to make a difference if the conditions is used within an inline stlye viewhelper or in a tag-style viewhelper.
TYPO3-Bug-Ticket: https://forge.typo3.org/issues/83812

I added 3 functional test, 2 of them are failing, which confirms the bug-report.


mneuhaus:
there is a new PullRequest that addresses this Issue :) 
https://github.com/TYPO3/Fluid/pull/380

Issue #374, opened 2018-03-01
Debug ViewHelper uses incompatible format


markusguenther:
The ISO8601 format is a left over and is not compatible with DateTime format.
We should use the recommended ATOM or DATE_ATOM.

More informations on:
https://secure.php.net/manual/en/class.datetime.php#datetime.constants.cookie

PR #375, opened 2018-03-01
[BUGFIX] Switch incompatible use of DateTime format

markusguenther:
The debug viewhelper use now DateTime::ATOM instead of DateTime::ISO8601.

This format is not compatible with ISO-8601, but is left this way for backward compatibility reasons. Use DateTime::ATOM or DATE_ATOM for compatibility with ISO-8601 instead.

Resolves: #374

### DIFF
--- a/src/ViewHelpers/DebugViewHelper.php
+++ b/src/ViewHelpers/DebugViewHelper.php
@@ -49,7 +49,6 @@ class DebugViewHelper extends AbstractViewHelper
 
     /**
      * @return void
-     * @throws \TYPO3Fluid\Fluid\Core\ViewHelper\Exception
      */
     public function initializeArguments()
     {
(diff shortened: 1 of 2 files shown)


mbrodala:
What's the purpose of this change? Any format is fine as long as it displays enough information about the debugged value. Thus `DateTime::ISO8601` is fine actually.

markusguenther:
This is a template engine and the code should be as clean as possible. So if the language says that this is not compatible and recommend another option we should use this.

Maybe just my opinion but guess a engine should also have the purpose to be clean.

liayn:
I agree with Markus here. Change this to ATOM.

jdreesen:
Right, `DateTime::ISO8601` does not what it says which means it's broken and should not be used.

markusguenther:
Hmm i can move this in a different pull request, but to be honest. The project has enough open PR.

mbrodala:
@markusguenther Here's the deal: I'll speed-merge both PRs if you split them. :-)

markusguenther:
Hehe we will see ;) don’t make it before evening.

markusguenther:
@mbrodala Cleanups has been removed. So this one is ready for review.

mbrodala:
@NamelessCoder beat me to it. ;-)

Issue #376, opened 2018-03-01
Cleanup Debug ViewHelper


markusguenther:
Little follow up from issue #375 

While adjusting this in the bugfix the wish popped up to have this clean up in a separate PR.
No PR without issue  ;)

NamelessCoder:
On GitHub, a PR is technically also an issue - but it never hurts to have one of each :)

lolli42:
Hmm. The PR has been closed. I guess it's fine to close here, too.

PR #377, opened 2018-03-01
TASK: Some smaller clean ups in the DebugViewHelper

markusguenther:
- using qualified references
- respect line limits
- simplify some constraints
- add missing throw annotation

Resolves: #376

### DIFF
--- a/src/ViewHelpers/DebugViewHelper.php
+++ b/src/ViewHelpers/DebugViewHelper.php
@@ -1,4 +1,5 @@
 <?php
+
 namespace TYPO3Fluid\Fluid\ViewHelpers;
 
 /*
@@ -49,13 +50,32 @@ class DebugViewHelper extends AbstractViewHelper
 
     /**
      * @return void
+     * @throws \TYPO3Fluid\Fluid\Core\ViewHelper\Exception
      */
     public function initializeArguments()
     {
         parent::initializeArguments();
-        $this->registerArgument('typeOnly', 'boolean', 'If TRUE, debugs only the type of variables', false, false);
-        $this->registerArgument('levels', 'integer', 'Levels to render when rendering nested objects/arrays', false, 5);
-        $this->registerArgument('html', 'boolean', 'Render HTML. If FALSE, output is indented plaintext', false, false);
+        $this->registerArgument(
+            'typeOnly',
+            'boolean',
+            'If TRUE, debugs only the type of variables',
+            false,
+            false
+        );
+        $this->registerArgument(
+            'levels',
+            'integer',
+            'Levels to render when rendering nested objects/arrays',
+            false,
+            5
+        );
+        $this->registerArgument(
+            'html',
+            'boolean',
+            'Render HTML. If FALSE, output is indented plaintext',
+            false,
+            false
+        );
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/ViewHelpers/DebugViewHelper.php
```diff
@@ -69,12 +89,12 @@ public static function renderStatic(array $arguments, \Closure $renderChildrenCl
         $typeOnly = $arguments['typeOnly'];
         $expressionToExamine = $renderChildrenClosure();
         if ($typeOnly === true) {
-            return (is_object($expressionToExamine) ? get_class($expressionToExamine) : gettype($expressionToExamine));
+            return (
+            \is_object($expressionToExamine) ? \get_class($expressionToExamine) : \gettype($expressionToExamine)
```
NamelessCoder: Please avoid slash-prefixing function calls. Though you and I and a few others understand why this micro-optimisation would be done, explaining it to people who ask "what's this" becomes a pretty tedious task for a benefit that is nearly non-existing when the code actually runs.

This is also the reason why such a conversion was not applied globally to this repository.

src/ViewHelpers/DebugViewHelper.php
```diff
@@ -111,11 +131,16 @@ protected static function dumpVariable($variable, $html, $level, $levels)
                 }
             }
         } else {
-            if (is_scalar($variable) || is_null($variable)) {
+            if (is_scalar($variable) || $variable === null) {
                 $string = sprintf(
                     '<code>%s = %s</code>',
                     $typeLabel,
-                    htmlspecialchars(var_export($variable, true), ENT_COMPAT, 'UTF-8', false)
+                    htmlspecialchars(
```
NamelessCoder: I don't mean to nit-pick here, but the whitespace formatting is now applied inconsistently - there are several other places where a similar breaking isn't done even though there are nested function calls.

Rule of thumb is: if you have to use three or more functions, always break them out. If the line remains shorter than the ~120 char limit, usually there's no need to break it.

src/ViewHelpers/DebugViewHelper.php
```diff
@@ -144,27 +169,33 @@ protected static function dumpVariable($variable, $html, $level, $levels)
      */
     protected static function getValuesOfNonScalarVariable($variable)
     {
-        if ($variable instanceof \ArrayObject || is_array($variable)) {
-            return (array) $variable;
-        } elseif ($variable instanceof \Iterator) {
+        if ($variable instanceof \ArrayObject || \is_array($variable)) {
+            return (array)$variable;
```
NamelessCoder: TYPO3 CMS uses a CGL where there is no space between cast and value. TYPO3 Fluid CGL by convention is to have a space.

The elseif to sequence of if statements here imho don't add to the readability and they make no difference to the execution. The rule set for number of elseif checks is not declared in this repository but again, if it's a reasonably low number there's no problem with it - but in larger quantities, a switch is preferred because it compiles to a hash map.

In this case, a better refactoring might have been a `switch (true) { case $variable instanceof Foobar: break; }` etc.

src/ViewHelpers/DebugViewHelper.php
```diff
@@ -1,4 +1,5 @@
 <?php
+
```
NamelessCoder: No line break here. None of the other classes contain this.

src/ViewHelpers/DebugViewHelper.php
jdreesen: > a switch is preferred because it compiles to a hash map.
> In this case, a better refactoring might have been a `switch (true) { case $variable instanceof Foobar: break; }` etc.

Actually, this only happens since PHP 7.2 and **only** if **all** cases are integers or strings. Otherwise it behaves like an `elseif` internally. Thus your example wouldn't trigger the optimization because it uses `instanceof` in the cases.

See: https://derickrethans.nl/php7.2-switch.html

PR #378, opened 2018-03-02
[BUGFIX] Fix numeric key array reversing in f:for

### LINKED ISSUE #372: ForViewHelper looses array keys with "reverse"
When setting reverse="true" array keys get lost.
Example from DateMenu in tx_news:
<f:for each="{data.single}" key="year" as="months" reverse="true">

The problem is, that in Line 112 of the viewhelper class "array_reverse" is used without second argument:
112: $arguments['each'] = array_reverse($arguments['each']);

Instedad it could be solved with:
$arguments['each'] = array_reverse($arguments['each'],true);

Greetings,
Manfred

mbrodala:
Fixes #372

### DIFF
--- a/src/ViewHelpers/ForViewHelper.php
+++ b/src/ViewHelpers/ForViewHelper.php
@@ -105,7 +105,7 @@ class ForViewHelper extends AbstractViewHelper
                 $each = $arguments['each'];
                 $arguments['each'] = iterator_to_array($each);
             }
-            $arguments['each'] = array_reverse($arguments['each']);
+            $arguments['each'] = array_reverse($arguments['each'], true);
         }
         if (isset($arguments['iteration'])) {
             $iterationData = [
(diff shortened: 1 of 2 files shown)


NamelessCoder:
Change looks fine to me.

But reconsidering the Generator usage in test cases... as far as I am aware we don't have any cases of that being used at this time, so for consistency and to avoid confusion, I would rather we have the data provider method return a simple array like it does in other tests.

Side note: Generator makes a lot more sense when your data providers are more complex than returning mere static arrays. I think we should pick Generators only when arrays would be worse to use - we do actually have some cases where they could reduce the size of data provider methods.

mbrodala:
Fine with me, personally I settled with using generators for all test data providers simply for better readability.

mbrodala:
Not sure why Travis keeps hanging. :-/

NamelessCoder:
download

mbrodala:
Cancelled and restarted the hanging job, now the whole build is green but the status still doesn't update here. 🤔

NamelessCoder:
Let's retry this later tonight

NamelessCoder:
Or not.

mbrodala:
Thanks. :-)

PR #380, opened 2018-03-08
[BUGFIX] Fix Countable- and multibyte support in conditions

mneuhaus:
replaces #370


tests/Unit/Core/Parser/BooleanParserTest.php
```diff
@@ -107,15 +108,15 @@ public function getSomeEvaluationTestValues()
             ['0 || 0', false],
             ['0 || 1', true],
             ['(0 && 1) || 1', true],
-            ['(0 && 0) || 0', true],
```
ohader: That was the "critical mass" of zeros... I see... ;)

mneuhaus: yea, that was a multibyte issue that could come from typing ```(0[SPACE]&&[SPACE])[SPACE]||[ALT-SPACE]0``` that "alt+space" after the "or" creates a custom multibyte space character that the parser choked on :D

Issue #385, opened 2018-05-16
Fluid xmlns declaration silently fails if using https


georgringer:
Currently fluid just checks `http://typo3.org/ns' in `NamespaceDetectionTemplateProcessor` with 
```
if (strpos($namespaceUrl, 'http://typo3.org/ns/') === 0) {
    $namespaceUri = substr($namespaceUrl, 20);
    $namespacePhp = str_replace('/', '\\', $namespaceUri);
} elseif (!preg_match('/([^a-z0-9_\\\\]+)/i', $namespaceUrl)) {
    $namespacePhp = $namespaceUrl;
    $namespacePhp = preg_replace('/\\\\{2,}/', '\\', $namespacePhp);
} else {
    $namespacePhp = null;
}
```

It should also allow `https://typo3.org/ns``
This issue is a copy of https://forge.typo3.org/issues/85018

NamelessCoder:
I'm afraid this is a misunderstanding (see PR thread). The schema has only one identifier. Allowing it to have a secondary identifier with https will merely confuse an IDE because the schema no longer contains the same ID as your import. Keep in mind that the xmlns-style imports have a meaning both in Fluid and generally as XML!

sascha-egerer:
As already said the namespace is a case sensitive identifier. This change does not really make sense.
See also https://www.w3.org/TR/2009/REC-xml-names-20091208/#NSNameComparison

webloft:
It may be fine that namespace is case sensitve - but it should not silently ignore anything.

webloft:
OK. But using a different scheme does not change the actual URI in my sight even it is not a goal to support multiple schemes. Its a more a logic thing - like forward non-www to www-address. If the URI contains the token //typo3.org/ns/ it should be clear that a PHP namespace should be used and the problem is gone - any points against?

NamelessCoder:
> any points against?

Absolutely there are points against. At the core of the issue is the fact that this isn't a URI, it is an ID. We do not allow "loose" IDs anywhere - that would add to the confusion, not prevent it. They *must* be unique and there *must* exist only one ID for each namespace.

This doesn't only relate to the PHP namespaces. It's also about identifying XSD files - the schema isn't something we invented and we have to follow the rules for implementing it. It's easy to get tunnel vision and look at only the xmlns-to-PHP relationship of Fluid, but it actually builds on well defined standards that impose this ID pragma.

[...]

PR #386, opened 2018-05-16
Allow https://typo3.org/ns/ as namespace declaration

georgringer:
Resolves: #385


NamelessCoder:
Hi Georg,

The very short version of why this is not going to happen, is that the NS "URLs" are in fact not URLs. They are identifiers. The URLs serve no content and you are not supposed to download them. All they do is _act as an identity of this specific schema so that you can connect the schema to the import statement by way of this URL instead of a file path, then import and use the schema(s) independently of their file locations_.

So there is no sense in adjusting which URLs are supported: anything you add that is *not* the correct URL (e.g. if you use https instead) is the same as using an invalid ID and will only sever the connection between schema and import.

Please also note that the namespace URL (the ID) is recorded in the schema file itself.

sascha-egerer:
See https://github.com/TYPO3/Fluid/pull/360

NamelessCoder:
Closed with same reason as https://github.com/TYPO3/Fluid/pull/360. It may (likely will) be possible in the future to use different prefixes (that you create) if you don't mind the situation described in #260.

PR #388, opened 2018-05-29
Add class name to exception message

revoltek-daniel:


NamelessCoder:
For posterity:

* The commit message subject does not confirm to our requirements (`[TASK]` prefix for this one)
* The code CGL is incorrect. When we have this structure, we split arguments into individual lines.

I would suggest always leaving pull requests open for at least 24 hours unless they've been cleared with a team member.

mbrodala:
@NamelessCoder Sorry, I didn't check the commit message itself, only the change.

If this change violates the CGL, shouldn't Travis and/or Scrutinizer say so?

NamelessCoder:
> If this change violates the CGL, shouldn't Travis and/or Scrutinizer say so?

I don't think either will report these things currently, though it would be nice if they did. But we really need to be more careful as well because when squash merging (which you should always do) the commit message most often must be rewritten and there are no checks at this point, what you write is what gets committed w/o needing a Travis test pass.

mbrodala:
OK, I'll try to keep this in mind. :-)

PR #393, opened 2018-07-09
[FEATURE] Deprecate evaluateCondition() in favor of verdict()

NamelessCoder:
The AbstractConditionViewHelper method evaluateCondition
was less than ideal, in that it:

* Only received arguments and thus could not support context
  related conditions.
* Was protected which prevents a future refactoring to call the
  method to evaluate directly from the compiled code.
* Was not strictly requiring an array of arguments which meant
  technically unsafe code (static analysis would demand guard)

Instead of breaking the API, the old method is deprecated in
favor of a new method which strictly requires arguments as an
array and the current instance of the rendering context.

The new method is made public though it is still not being
called as such - this is in preparation for a future improvement
of the compiled code to clean up the current closure-based
solution for evaluating the if-elseif-else structure.

In addition, this patch:

* Refactors the condition ViewHelper so using the base class
  does not automatically register a “condition” argument.
* Refactors `f:if` to use the new API method.

### DIFF
--- a/src/Core/ViewHelper/AbstractConditionViewHelper.php
+++ b/src/Core/ViewHelper/AbstractConditionViewHelper.php
@@ -44,7 +44,60 @@ abstract class AbstractConditionViewHelper extends AbstractViewHelper
     {
         $this->registerArgument('then', 'mixed', 'Value to be returned if the condition if met.', false);
         $this->registerArgument('else', 'mixed', 'Value to be returned if the condition if not met.', false);
-        $this->registerArgument('condition', 'boolean', 'Condition expression conforming to Fluid boolean rules', false, false);
+    }
+
+    /**
+     * Renders <f:then> child if $condition is true, otherwise renders <f:else> child.
+     * Method which only gets called if the template is not compiled. For static calling,
+     * the then/else nodes are converted to closures and condition evaluation closures.
+     *
+     * @return string the rendered string
+     * @api
+     */
+    public function render()
+    {
+        if (static::verdict($this->arguments, $this->renderingContext)) {
+            return $this->renderThenChild();
+        }
+        return $this->renderElseChild();
+    }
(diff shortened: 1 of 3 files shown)


src/Core/ViewHelper/AbstractConditionViewHelper.php
```diff
@@ -60,6 +60,7 @@ public function initializeArguments()
      * subclasses that will be using this base class in the future. Let this
      * be a warning if someone considers changing this method signature!
      *
+     * @deprecated Deprecated in favor of ClassName::verdict($arguments, renderingContext), will no longer be called in 3.0
```
mbrodala: This method should be moved downwards (after `public`, before `private`) then.

NamelessCoder:
Made a few additions which should be completely safe to do and means custom condition ViewHelpers can be made to very clearly support different arguments than `condition`. Not having this as a required argument on the base class is IMHO cleaner - but technically it means that if someone, somewhere, has a custom condition VH which does not subclass `f:if` but instead subclasses the abstract condition VH class, and that class uses the `condition` argument but does not register it.

I think that use case is extremely unlikely.

kitsunet:
Nice change! I like the direction a lot :)

NamelessCoder:
Thanks @kitsunet - do you know of any condition VH use cases in Neos that might be negatively affected by removing the `condition` argument from the abstract class? We've got a few "role.isFoobar" conditions in TYPO3, which come from Flow so you probably have the same or more.

kitsunet:
Will check, but even if, that is easily fixed :)

NamelessCoder:
Rebased for your convenience :)

NamelessCoder:
I'd throw an "Approved" for the unit test change - but I'm not allowed. So here it is instead: 👍

Issue #148, opened 2016-09-16
Cant check for "variable is not set" in Ternary Expressions 


sascha-egerer:
It is not possible to check if a variable is not set in a ternary expression.

`{ myArray.invalidVariable ? 'yes' : 'no' }` Will always return `yes` because `myArray.invalidVariable` is evaluated as a string as not variable is found with this name

`{('{myArray.invalidVariable}' !== '') ? 'yes' : 'no'}` does also not work as the parser does not recognize this syntax as a ternary expression but simply outputs the text.

NamelessCoder:
Looks like we should change the default to NULL when variable does not exist. Regarding the more complex expression the TernaryExpressionNode and other ExpressionNodes only support strict variable references - and no nested expressions either.

sascha-egerer:
I tried that already but had no time to dig further into this. Simply changing this behavior does not work as the function `getTemplateVariableOrValueItself` is used and this behavior is also required in some cases.

NamelessCoder:
It would seem we should not use `getTemplateVariableOrValueItself` - instead we can use `$renderingContext->getVariableProvider()->get('name')`. The "variable or value itself" logic is intended for use in cases where you *do need the hardcoded value*, for example in MathExpressionNode where expression parts can be operators/operands/hardcoded numbers rather than merely variable references. The same is *not true* for TernaryExpressionNode which must explicitly only support template variable references.

NamelessCoder:
On a side note: this expression node class could use a bit of cleanup to remove the `'true'` and `'false'` expression matches since they do not make sense in this context.

PR #394, opened 2018-07-09
[BUGFIX] Fix ternary condition behaviors

NamelessCoder:
This patch corrects the following problems for
ternary conditions:

* Dotted paths were not detected correctly.
* The “then” part was mandatory
* Negating the condition did not work correctly

All three described cases are working now.

Fixes: #148

### DIFF
--- a/examples/Resources/Private/Singles/Conditions.html
+++ b/examples/Resources/Private/Singles/Conditions.html
@@ -10,7 +10,12 @@ Conditions template from Singles.
 	<f:else>Not touched</f:else>
 </f:if>
 
-Ternary expression: {checkTernary ? ternaryTrue : ternaryFalse}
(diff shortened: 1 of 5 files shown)


examples/Resources/Private/Singles/Conditions.html
```diff
@@ -10,7 +10,12 @@
 	<f:else>Not touched</f:else>
 </f:if>
 
-Ternary expression: {checkTernary ? ternaryTrue : ternaryFalse}
+Ternary expression without then case: {ternaryTrue ? : ternaryFalse}
```
mbrodala: I assume this also works with `?:` instead of `? :`? If yes, then we should prefer that since that's the common notation in PHP.

NamelessCoder: Works both without and with any number and type of spaces between the parts.

PR #397, opened 2018-07-16
[BUGFIX] Process paths individually in warmup

NamelessCoder:
A bug in the warmup code meant that all templates,
partials and layouts which were in sub-folders in
the root paths, would cause file not resolved errors
due to using basename() on the absolute file path.

To remedy this, each path has to be traversed by
itself and the relative file path has to be determined
by trimming off the root path. The result is that any
level of sub-folders can be supported in warmup.


mbrodala:
I can test as soon as https://github.com/NamelessCoder/typo3-cms-fluid-precompiler/pull/5 is merged. ;-)

NamelessCoder:
Small Fluid patches are just perfect for gearing down from being "in the zone" for almost a year now. But I *am* gearing down to actually go on break ;)

PR #398, opened 2018-07-26
[TASK] Extend MathExpressionNode tests with chained math

NamelessCoder:
Adds test cases for expressions like {a + 1 * 2}.

Also labels test cases.


NamelessCoder:
This would actually have been rebased away since it got fixed in another PR ;)

mbrodala:
Hm, so I just wasted time coming up with nice labels? ;-)

mbrodala:
OK, that's true indeed. Notice that I would rewrite `$a(value)` to `$a (value)` to make it clear that this is not a function call but instead a variable with its internal value.

NamelessCoder:
@mbrodala I think this is ready - trimmed down to only adding test cases for chained math.

tests/Unit/Core/Parser/SyntaxTree/Expression/MathExpressionNodeTest.php
```diff
@@ -39,16 +39,21 @@ public function testEvaluateExpression($expression, array $variables, $expected)
...
-            ['2 * 4', [], 8],
-            ['4 / 2', [], 2],
-            ['4 ^ 2', [], 16],
-            ['a + 1', ['a' => 1], 2],
-            ['1 + b', ['b' => 1], 2],
-            ['a + b', ['a' => 1, 'b' => 1], 2],
+            'Invalid operator returns zero' => ['1 gabbagabbahey 1', [], 0],
+            'Simple addition' => ['1 + 1', [], 2],
+            'Chained addition' => ['1 + 1 + 1', [], 3],
+            'Multiplication before addition' => ['1 + 1 * 2', [], 4],
```
mbrodala: Uhm, shouldn't this be `3`?

NamelessCoder: The expressions don't know about operator priority, so `1 + 1 = 2 * 2 = 4` is what it sees. It intentionally also doesn't know about parenthesis grouping in order to KISS.

If you know a reasonable way to implement operator priority (see also source of the expression class) I'd definitely consider adding that as a feature.

PR #401, opened 2018-08-07
[BUGFIX] Prevent binary characters in compiled templates

mbrodala:
When compiled templates are retrieved from the cache and contain binary characters, arbitrary behavior may occur, usually a huge pile of Chinese-like characters is outputted and the template class cannot be loaded. Avoid this by delegating the generation of the compiled node state to the node converter.

### DIFF
--- a/src/Core/Compiler/TemplateCompiler.php
+++ b/src/Core/Compiler/TemplateCompiler.php
@@ -230,7 +230,7 @@ class TemplateCompiler
 %s {
 
 public function getLayoutName(\TYPO3Fluid\Fluid\Core\Rendering\RenderingContextInterface \$renderingContext) {
-\$self = \$this; 
+\$self = \$this;
 %s;
 }
 public function hasLayout() {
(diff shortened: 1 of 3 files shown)


mbrodala:
I'll fix the failing tests later.

mbrodala:
A bit more background: the binary characters come from the serialization of objects with protected or private properties since PHP mangles their names using 0 (null) bytes.

NamelessCoder:
Please use `PHP_EOL` instead of `chr(10)`. You could also easily sell this one as a TASK to clean up the conversion of VH argument nodes / removing duplicated conversion.

mbrodala:
I tried to stick to the line endings used in the `TemplateCompiler`. Personally I'd prefer `"\n"` or even heredoc where possible.

NamelessCoder:
I'd welcome a PR that string-replaces `chr(10)` with `PHP_EOL` - some of these are nearly a decade old by now ;)

mbrodala:
Sure thing, but first I'd like to see this one fixed. ;-)

Issue #400, opened 2018-08-02
Refactoring content argument support into AbstractViewHelper


NamelessCoder:
This issue is more of a TODO:

It would make a lot of sense to refactor the CompileWithContentArgumentAndRenderStatic trait to merge it into AbstractViewHelper, thus giving any ViewHelper the option of setting `protected $contentArgumentName = 'myargument';` in a VH class to opt-in to using content arguments.

The trait would then be deprecated.

There is one pitfall:

The content argument name resolving method in the trait cannot be copied to the AbstractViewHelper. Instead, this abstract class must only ever do content argument related processing if `$this->contentArgumentName` is explicitly not-null. Otherwise, all ViewHelpers would start to automatically use the first optional argument as content argument which we definitely want to avoid.

The deprecation of the trait does not necessarily need to be done any time soon as it would still provide the auto-resolving capability instead of requiring an explicit property in the class. But it may be necessary to refactor the trait to rename the property to avoid a collision.

NamelessCoder:
Related issue: #387

NamelessCoder:
Made redundant by up-coming Fluid 3.0.

PR #404, opened 2018-08-23
[FEATURE] Allow content argument on ViewHelpers without trait

NamelessCoder:
This change deprecates CompileWithContentArgumentAndRenderStatic
in favor of a similar, but explicit, solution on AbstractViewHelper.

Content argument is possible on any ViewHelper after this
change but must explicitly be opted-in by one of two methods:

* Set `protected $contentArgumentName` to name of argument.
* Override `resolveContentArgumentName` and return name.

The latter of the two methods is implemented so ViewHelpers
can be plug-and-play compatible even with the trait hollowed
out and deprecated.

A couple of ViewHelpers and related tests have additionally
been refactored to not assume that render() is implemented
(the method is now optional), instead using static calling. This
allowed removing the deprecated trait from those ViewHelpers.

Close: #400

### DIFF
--- a/src/Core/ViewHelper/AbstractViewHelper.php
+++ b/src/Core/ViewHelper/AbstractViewHelper.php
@@ -7,6 +7,7 @@ namespace TYPO3Fluid\Fluid\Core\ViewHelper;
  */
 
 use TYPO3Fluid\Fluid\Core\Compiler\TemplateCompiler;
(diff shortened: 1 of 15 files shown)


NamelessCoder:
Closing; the desired behavior can be achieved with null coalesce without using the trait.

`$argOrContent = $arguments['foo'] ?? $this->renderChildren();`

PR #406, opened 2018-09-09
[TASK] Refactor child node evaluation to iterable

NamelessCoder:
This patch slightly refactors the way evaluateChildNodes
behaves, allowing the return value from getChildNodes
to be any type of iterable type/object.

There are currently no use cases in existence that would
change a node to return non-arrays from getChildNodes
but in order to support this in the future we can refactor
now without ill effects.

(I am in fact working on something that would require it)

### DIFF
--- a/src/Core/Parser/SyntaxTree/AbstractNode.php
+++ b/src/Core/Parser/SyntaxTree/AbstractNode.php
@@ -31,20 +31,18 @@ abstract class AbstractNode implements NodeInterface
      */
     public function evaluateChildNodes(RenderingContextInterface $renderingContext)
     {
-        $childNodes = $this->getChildNodes();
-        $numberOfChildNodes = count($childNodes);
-        if ($numberOfChildNodes === 0) {
+        $evaluatedNodes = [];
+        foreach ($this->getChildNodes() as $childNode) {
+            $evaluatedNodes[] = $this->evaluateChildNode($childNode, $renderingContext, false);
+        }
+        // Make decisions about what to actually return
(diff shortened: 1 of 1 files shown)


src/Core/Parser/SyntaxTree/AbstractNode.php
```diff
@@ -56,12 +54,22 @@ public function evaluateChildNodes(RenderingContextInterface $renderingContext)
...
+            $output = $this->cast($output);
+        }
+        return $output;
+    }
+
+    /**
+     * @param mixed $value
+     * @return string
+     */
+    protected function cast($value)
```
mbrodala: This should be named `castToString` to clarify its purpose.

src/Core/Parser/SyntaxTree/AbstractNode.php
```diff
@@ -31,20 +31,18 @@
...
         }
-        $output = '';
-        /** @var $subNode NodeInterface */
-        foreach ($childNodes as $subNode) {
-            $output .= $this->evaluateChildNode($subNode, $renderingContext, true);
+        if (count($evaluatedNodes) === 1) {
+            return $evaluatedNodes[0];
         }
-        return $output;
+        return (string) implode('', array_map([$this, 'cast'], $evaluatedNodes));
```
mbrodala: You can drop the string cast for the return value of `implode()` since that function always returns a string.

NamelessCoder:
Good points, taking care of it :)

PR #415, opened 2018-10-29
[FEATURE] Inline fluid ViewHelper

NamelessCoder:
Allows rendering a variable containing Fluid code as
if the code contained in the variable were part of the
template invoking the ViewHelper.

    $view->assign('variable', 'value of my variable');
    $view->assign('code', 'My variable: {variable}');

And in the template:

    {code -> f:inline()}

Which outputs:

    My variable: value of my variable

You can use this to pass smaller and dynamic pieces of Fluid code
to templates, as an alternative to creating new partial templates.

### DIFF
--- a/src/ViewHelpers/InlineViewHelper.php
+++ b/src/ViewHelpers/InlineViewHelper.php
@@ -1,0 +1,62 @@
+<?php
+namespace TYPO3Fluid\Fluid\ViewHelpers;
+
+use TYPO3Fluid\Fluid\Core\Rendering\RenderingContextInterface;
+use TYPO3Fluid\Fluid\Core\ViewHelper\AbstractViewHelper;
+use TYPO3Fluid\Fluid\Core\ViewHelper\Traits\CompileWithContentArgumentAndRenderStatic;
+
+/**
+ * Inline Fluid rendering ViewHelper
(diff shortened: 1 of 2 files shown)


src/ViewHelpers/InlineViewHelper.php
```diff
@@ -0,0 +1,64 @@
...
+    /**
+     * @return void
+     */
+    public function initializeArguments()
+    {
+        $this->registerArgument(
+            'code',
+            'string',
+            'Fluid code to be rendered as if it were part of the template rendering it. Can be passed as inline '.
+            'argument or tag content'
```
mbrodala: Please merge these two lines. Right now it looks like 4 arguments are passed to `registerArgument()`

NamelessCoder: Just for the record I'm personally all for allowing exceptions of longer lines but technically we do have a 130 char max. Again personally, I think it's time we all started reconsidering this limit because it isn't always a good idea to blindly enforce it ;) at any rate: changed, pushed.

NamelessCoder:
Don't merge, double linebreak detected.

NamelessCoder:
Fixed.

PR #419, opened 2018-11-16
[BUGFIX] Prevent property spill-over to next call

jonaseberle:
We have found an issue where consecutive viewhelpers take over properties from the previous one. 

This is the TYPO3 bug report: https://forge.typo3.org/issues/86890

### DIFF
--- a/src/Core/ViewHelper/AbstractTagBasedViewHelper.php
+++ b/src/Core/ViewHelper/AbstractTagBasedViewHelper.php
@@ -88,6 +88,9 @@ abstract class AbstractTagBasedViewHelper extends AbstractViewHelper
     public function initialize()
     {
         parent::initialize();
+        $this->tag->reset();
+        $this->tag->setTagName($this->tagName);
+
         if ($this->hasArgument('additionalAttributes') && is_array($this->arguments['additionalAttributes'])) {
             $this->tag->addAttributes($this->arguments['additionalAttributes']);
         }


src/Core/ViewHelper/AbstractTagBasedViewHelper.php
```diff
@@ -88,6 +88,9 @@ public function initializeArguments()
     public function initialize()
     {
         parent::initialize();
+        $this->tag->reset();
+        $this->tag->setTagName($this->tagName);
```
mbrodala: I wonder if this could be moved to `setTagBuilder()` and drop the `$this->tagName` injection from the constructor ...

NamelessCoder: Rather, I'd remove the line that sets the tag name in `setTagBuilder` since setter methods shouldn't mutate the object that is passed.

TagBuilders are already initialized with tag name as constructor argument ;)

In the future, tag based VH should become compilable in a way that TagBuilder is constructed outside and has one instance per execution, then passed to a static method that will render the tag. Then the TagBuilder reset and mutability can be reworked. Point being, if we aim for that then anything that treats the TagBuilder instance as a more stupid thing that needs init on every call, is imho a good thing. Even if it didn't also solve a bug ;)

mbrodala:
@NamelessCoder so are you fine with the change as is?

NamelessCoder:
This is okay with me. The one risk I see is someone overriding `setTagBuilder` to do manipulation which in the first place is not the right approach - init done there would be reset after this patch. I've never seen an implementation of this in the wild so I'd say: safe!

mbrodala:
OK, fine with me.

M4rt1n17:
This particular Problem is fixed at the attribute "required" at inputfields but  still is here at the attribute "checked" at radio buttons.

All of my radiobuttons have this field "checked='checked'".
I insert 8 radio buttons and did not even select one to be selected but all of them got the attribute.
This leads to: the last of them will be standard selected.
If I select one to be selected this is having no effect as all already will have the attribute checked and also the last of them will be standard selected.

frontend
radio buttons
typo3 backend radio button

As I think this is related to the same issue I will just post this here.
I'm already using the latest fluidpages-master (and also vhs-master and flux-master)

NamelessCoder:
A regression has been detected as a consequence of this patch.

TagBasedViewHelpers with `data-` prefixed arguments will now discard these prefixed attributes, since they are assigned to the TagBuilder through `handleAdditionalArguments` which is called before `initializeArgumentsAndRender`, but with this patch, `initializeArgumentsAndRender` will reset the TagBuilder which discards the previously registered attributes.

Test case:

```php

    public function testTagBasedViewHelperWithDataPrefixedArgument()
    {
        $invoker = new ViewHelperInvoker();
        $viewHelper = new TagBasedTestViewHelper();
        $arguments = [
            'data-foo' => 'bar',
        ];
        $result = $invoker->invoke($viewHelper, $arguments, new RenderingContextFixture());
        $this->assertSame('<div data-foo="bar" />', $result);
    }
```

Test output:

```plain
Failed asserting that two strings are identical.
Expected :'<div data-foo="bar" />'
Actual   :'<div />'
```

Issue #420, opened 2018-12-03
Math expression allow only variables with Camelcase and not working with Snakecase


Atsyn:
I can't understand what wrong with Snake Case and why you not include it in pattern.

Database fields, in default, set in Snakecase it's mean "data" variable can't be used without additional manipulations.

One small fix in the regexp pattern should resolve it.

liayn:
> I can't understand what wrong with Snake Case and why you not include it in pattern.

I guess this is simply a mistake. @NamelessCoder can for sure tell us more about that.

mbrodala:
Yes, the `_` character is simply not accepted ATM:

https://github.com/TYPO3/Fluid/blob/f583bad/src/Core/Parser/SyntaxTree/Expression/MathExpressionNode.php#L28

@Atsyn can you open a PR to fix this?

PR #421, opened 2018-12-05
[BUGFIX] Allow variables in Snakecase in expressions and ternaries

Atsyn:
Resolves: #420


liayn:
I guess we need the same for boolean expressions and ternaries

Atsyn:
Updated for all expressions and ternaries.

mbrodala:
@NamelessCoder I've always seen the same tests failing locally that now fail on Travis. Any idea what's going on here?

sascha-egerer:
> @mbrodala Stack trace is strange.
> Error not related with Fluid. =(

I think @mbrodala means the failing tests and not the deprecation warnings. So there is no stacktrace for the tests. There are just 2 items in the wrong order.

NamelessCoder:
@mbrodala We switched from `glob()` to recursive DirectoryIterator:

https://github.com/TYPO3/Fluid/commit/6beef025fe3c1071c5994a7700f8efe3469dcdce#diff-7892b10d8d052eaed47516c29d922827

It's possible something changed in how DirectoryIterator sorts files, or how PHPUnit determines equalness when calling `assertEquals` with two arrays as input (when before it didn't care about order, now it does). It would be nice to prepare another PR that changes these two tests to check if all expected items are contained without caring about the order.

NamelessCoder:
Passing `true` as 4th parameter for `assertEquals` causes arrays to be sorted before comparison:

https://github.com/sebastianbergmann/comparator/blob/master/src/ArrayComparator.php#L46

NamelessCoder:
Seems like our PHPUnit version is also terribly outdated and lacks this parameter - and it's going to take quite some work to get it updated...

Atsyn:
@mbrodala do you mean my changes with variables?

mbrodala:
Yeah, basically just add a `snail_cased` variable to the existing tests of all classes that where changed.

mbrodala:
I'd still love to have a test for this ...

NamelessCoder:
@mbrodala I would merge this w/o requiring unit tests to cover it. This is actually a case of the TemplateParser Patterns not being reproduced correctly in the expression node types. The longer-perspective solution would be to somehow refactor, making the variable detection parts central in TemplateParser and then providing thorough tests (that also behave as declarative "which variable names do we actually support?" type info). 

Finally this then logically leads to somehow refactoring TemplateParser's many different usages of "a regexp to detect a variable reference". In my opinion, by loosening up the detection expressions entirely to allow anything, then using sub-detection to determine what type of node actually gets created (match variables with `.` and detect if garbage -> create TextNode instead of any Fluid-active type of node).

Assuming the switch to a lexer solution doesn't happen first...

mbrodala:
Fine with me but it still feels unsafe to not have any code-written declaration that we support such variable names.

NamelessCoder:
Amen to that. I'll take a look at this right away because those patterns are just... excessively hard to read.

PR #422, opened 2018-12-05
[TASK] Canonicalize arrays before comparison

NamelessCoder:
Fixes a sometimes failing, sometimes succeeding unit
test which compares arrays of resolved files.

Sorts both arrays before comparison because at this
point in tests we don’t care about the order, which is
tested by other methods.

Unblocks #421


NamelessCoder:
Note: this PR is a stop-gap measure that doesn't address the core problem - which is that our PHPUnit version is very outdated. That update however requires significantly more time than I have available, so the real solution that would fix the problem described in #421 is *replicated* for this test method only.

Issue #427, opened 2019-02-14
Feature: slots


mbrodala:
The `contentAs` argument provided by Fluid is very useful to pass arbitrary content as variable to a partial/section.

However, this is limited to a single variable, if multiple pieces of content should be passed, one has to e.g. resort to doing inline `f:render` calls within `arguments` instead.

A nice solution to this would be a system similar to [Vue.js component slots](https://vuejs.org/v2/guide/components-slots.html).

### Basic example

`Example.html` template:

```html
<f:render partial="Foo" arguments="{as: 'usual'}">
Some content with<f:if condition="{lie}">out</f:if> regular Fluid.
</f:render>
```

`Foo.html` partial:

```html
Just the {as}:
<f:slot/>
```

### Named slots example

`Example.html` template:

```html
<f:render partial="BaseLayout">
  <f:template slot="header">
    <h1>Here might be a page title</h1>
  </f:template>

  <p>A paragraph for the main content.</p>
  <p>And another one.</p>

  <f:template slot="footer">
    <p>Here's some contact info</p>
  </f:template>
</f:render>
```

`BaseLayout.html` partial:

```html
<div class="container">
  <header>
    <f:slot name="header"/>
  </header>
  <main>
    <!-- Slot without name renders the content not encapsulated in `f:template` -->
    <f:slot/>
  </main>
  <footer>
    <f:slot name="footer"/>
  </footer>
</div>
```

If we stick to the same naming or use something like `argument` is up to discussion of course as is this feature in general.

NamelessCoder:
Perhaps easier implemented and understood if using `f:variable` - hypothetical and unsupported currently:

```xml
<f:render partial="Something">
   <f:variable name="arg1">Special value for arg1 variable</f:variable>
   <f:variable name="arg2">Special value for arg2 variable</f:variable>
</f:render>
```

The result being that in addition to, or instead of, assigning these variables only when rendering the content and then clearing them before actually rendering the partial or whatnot - _the variables would be redirected to be collected as arguments for calling the partial_.

This could be implemented with very little effort (toggle set in ViewHelperVariableContainer by `f:render` and respected by `f:variable` then extracting variables before sub-rendering) and has the benefit that it only uses already supported features but makes them work slightly different when used in `f:render`.

Bonus: it becomes *excessively* easy to declare long-winded, non-inline, HTML-carrying, not-escaped variables this way since this would work regardless of `contentAs` and for any type of render statement as a means to expand arguments.

[...]

Issue #424, opened 2019-01-06
Vision: Unification of component-level rendering related API


NamelessCoder:
### Introduction

This issue aims to describe a suggested vision/strategy for achieving a couple of goals:

1. The vision to adopt atomic design principles as a native Fluid feature, https://gist.github.com/NamelessCoder/f599bf955a2cb6ab46996c1207af3a2e
2. A feature request for "importable sections", https://github.com/TYPO3/Fluid/issues/358
3. A new conception for template- or section-level definitions of required arguments, meaning variables that must exist when the template/section is rendered - making it work like a pseudo-ViewHelper without the need to create a PHP class.
4. The ability to document template files's required variables and create default values for optional arguments much the same way ViewHelpers are currently documented.
5. The fact that currently, Fluid enforces UpperCamelCase formats not only for PHP classes but also template files, https://github.com/TYPO3/Fluid/issues/371.

These goals all mesh together in a way that makes the following components very similar:

* A ViewHelper instance (until compiled)
* A node in a parsed-but-not-compiled template; this includes individual arguments
* A parsed section defined in a template
* A parsed template file, partial or layout (or atom, molecule or organism given point 1 above)
* The lesser known `RenderableInterface` which is accepted by `f:render` as `delegate` argument and can be implemented on arbitrary PHP classes you create instances of an assign to template variables.

Their similarities being that:

1. They all render a rendering context
2. They all have simple API methods (though some require initialisation, e.g. ViewHelpers)
3. They are all capable of carrying argument definitions (though some will have zero such arguments)

However, currently the situation is that templates use a `ParsedTemplateInterface`, ViewHelpers use a `ViewHelperInterface` - and sections defined in a template have no interface as such (is rendered through methods on parsed template's interface). How sections work is currently quite fragmented and rudimentary in Fluid and has on earlier occasions caused seemingly unrelated problems, e.g. reserved variable name `sections` because parsed sections were stored in template variables.

I'll use the word **component** to describe "an instance of something that can be rendered in Fluid" which currently includes templates (parsed instance of files/source) and ViewHelpers, and would include the atoms, molecules, organisms and sections once those are added.

[...]

PR #429, opened 2019-02-14
[FEATURE] Allow defining arguments for f:render with f:argument

NamelessCoder:
This patch allows you to define arguments that get passed to
f:render by using f:argument in the tag contents.

```xml
<f:render partial="Something">
   <f:argument name="arg1">Special value for arg1 variable</f:argument>
   <f:argument name="arg2">Special value for arg2 variable</f:argument>
</f:render>
```

Any argument specified with f:argument this way will override
the argument of the same name if it was passed in the “arguments”
array as well. The combined result will be used as variables for the
sub-rendering call.

References: #427

### DIFF
--- a/examples/Resources/Private/Singles/Variables.html
+++ b/examples/Resources/Private/Singles/Variables.html
@@ -40,6 +40,12 @@ Direct access of numeric prefixed variable: {123numericprefix}
 	}
 }"/>
 
+<!-- Passing arguments to a sections/partials using f:render in tag content -->
+<f:render section="Tertiary">
+    <f:argument name="arg1">First argument</f:argument>
+    <f:argument name="arg2">Second argument</f:argument>
+</f:render>
+
 </f:section>
 
 <f:section name="Secondary">
@@ -48,4 +54,9 @@ Received $array.printf with formatted string {array.printf -> f:format.printf(ar
 Received $array.baz with value {array.baz}
 Received $array.xyz.foobar with value {array.xyz.foobar}
 Received $myVariable with value {myVariable}
+</f:section>
+
+<f:section name="Tertiary">
+Input argument "arg1" was: {arg1}
+Input argument "arg2" was: {arg2}
 </f:section>
--- a/src/Core/ViewHelper/ViewHelperVariableContainer.php
+++ b/src/Core/ViewHelper/ViewHelperVariableContainer.php
@@ -6,6 +6,7 @@ namespace TYPO3Fluid\Fluid\Core\ViewHelper;
  * See LICENSE.txt that was shipped with this package.
  */
 
+use TYPO3Fluid\Fluid\Core\Variables\VariableProviderInterface;
 use TYPO3Fluid\Fluid\View\ViewInterface;
 
 /**
@@ -28,6 +29,69 @@ class ViewHelperVariableContainer
      * @var ViewInterface
      */
     protected $view;
+
+    /**
+     * @var VariableProviderInterface[]
+     */
+    protected $delegates = [];
+
+    /**
+     * Push a new delegate variable container to a stack.
+     *
+     * If a ViewHelper requires a storage to collect variables which, for
+     * example, are filled by evaluating the child (tag content) closure,
+     * this method can be used to add a special delegate variable container
+     * stored in a stack. Once the variables you need to collect have been
+     * collected, calling `popDelegateVariableProvider` removes the delegate
+     * from the stack.
+     *
+     * The point of a stack is to avoid resetting a storage every time a
+     * ViewHelper is rendered. In the case of `f:render` it means one storage
+     * is created and filled for every one call to the ViewHelper.
+     *
+     * It is VITAL that you also "pop" any delegate you push to this stack!
+     *
+     * @param VariableProviderInterface $variableProvider
+     */
+    public function pushDelegateVariableProvider(VariableProviderInterface $variableProvider)
+    {
+        $this->delegates[] = $variableProvider;
+    }
+
+    /**
+     * Get the topmost delegate variable container that was previously pushed
+     * onto the stack by pushDelegateVariableContainer(). This method returns
+     * a reference to the storage that was last added to the stack without
+     * removing the variable provider from the stack.
+     *
+     * Is used in ViewHelpers that assign variables in variable providers in
+     * the stack - as a means to get the variable storage used by the "closest
+     * parent", e.g. when called in `f:argument` used inside `f:render`, will
+     * read the delegate variable provider inserted by that parent `f:render`.
+     *
+     * @return VariableProviderInterface|null
+     */
+    public function getTopmostDelegateVariableProvider()
+    {
+        return end($this->delegates) ?: null;
+    }
+
+    /**
+     * Return and REMOVE the topmost delegate variable provider. This method
+     * must be called after you finish sub-rendering with a delegated variable
+     * provider that was added with `pushDelegateVariableProvider`. Calling
+     * the method removes the delegate and returns the stack to the previous
+     * state it was in.
+     *
+     * To avoid removing from the stack, use `getTopmostDelegateVariableProvider`.
+     *
+     * @param string $viewHelperClassName
+     * @return VariableProviderInterface|null
+     */
+    public function popDelegateVariableProvider()
+    {
+        return array_pop($this->delegates);
+    }
 
     /**
      * Add a variable to the Variable Container. Make sure that $viewHelperName is ALWAYS set
--- a/src/ViewHelpers/ArgumentViewHelper.php
+++ b/src/ViewHelpers/ArgumentViewHelper.php
@@ -1,0 +1,68 @@
+<?php
+namespace TYPO3Fluid\Fluid\ViewHelpers;
+
+/*
+ * This file belongs to the package "TYPO3 Fluid".
+ * See LICENSE.txt that was shipped with this package.
+ */
+
+use TYPO3Fluid\Fluid\Core\Rendering\RenderingContextInterface;
+use TYPO3Fluid\Fluid\Core\ViewHelper\AbstractViewHelper;
+use TYPO3Fluid\Fluid\Core\ViewHelper\Traits\CompileWithContentArgumentAndRenderStatic;
+
+/**
+ * Argument assigning ViewHelper
+ *
+ * Assigns an argument for a parent ViewHelper call when
+ * the parent ViewHelper supports it.
+ *
+ * Alternative to declaring an array to pass as "arguments".
+ *
+ * Usages:
+ *
+ *     <f:render partial="Foo">
+ *         <f:argument name="arg1">Value1</f:argument>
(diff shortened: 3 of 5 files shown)


examples/Resources/Private/Singles/Variables.html
```diff
@@ -40,6 +40,12 @@
 	}
 }"/>
 
+<!-- Passing arguments to a sections/partials using f:render in tag content -->
+<f:render section="Tertiary">
+    <f:variable name="arg1">First argument</f:variable>
```
mbrodala: I am not sure if we really want to overload this viewhelper with this additional magic. What if one wants to have some template code here, simplified using `f:variable` and then pass everything via `contentAs` as usual? Of course, one does not have to use all arguments passed to a section/partial but we should avoid passing around too much data.

I'd prefer a more explicit approach, e.g. with a new `f:argument` or `f:render.argument` viewhelper.

src/Core/ViewHelper/ViewHelperVariableContainer.php
```diff
@@ -29,6 +30,53 @@ class ViewHelperVariableContainer
...
+    }
+
+    /**
+     * Get the topmost delegate variable container that was previously pushed
+     * onto the stack by pushDelegateVariableContainer().
+     *
+     * @param string $viewHelperClassName
+     * @return VariableProviderInterface|null
+     */
+    public function getTopmostDelegateVariableContainer($viewHelperClassName)
```
mbrodala: I think normally the term `peek` would be used for stacks.

src/Core/ViewHelper/ViewHelperVariableContainer.php
```diff
@@ -29,6 +30,53 @@ class ViewHelperVariableContainer
      */
     protected $view;
 
+    /**
+     * Push a new delegate variable container to a stack. Done before you start
+     * sub-rendering with a ViewHelper that requires a special container for
+     * variables which is capable of being used in multidepth stacks without
+     * overriding the last instance or last set of variables.
```
mbrodala: No matter how often I read this, I always fail to keep track of all the parts. ;-) Please split this in few, easier to understand, sentences.

src/Core/ViewHelper/ViewHelperVariableContainer.php
```diff
@@ -29,6 +30,53 @@ class ViewHelperVariableContainer
...
+     *
+     * @param string $viewHelperClassName
+     * @param VariableProviderInterface $variableProvider
+     */
+    public function pushDelegateVariableContainer($viewHelperClassName, VariableProviderInterface $variableProvider)
```
mbrodala: The term `overlay` would be better than `delegate` IMO.

src/ViewHelpers/VariableViewHelper.php
```diff
@@ -57,6 +57,9 @@ public static function renderStatic(
     ) {
         $value = $renderChildrenClosure();
         $renderingContext->getVariableProvider()->add($arguments['name'], $value);
+        if ($delegateVariableProvider = $renderingContext->getViewHelperVariableContainer()->getTopmostDelegateVariableContainer(RenderViewHelper::class)) {
```
mbrodala: Can you split this to two lines? At least PHPStan doesn't like variable assignment in conditions. ;-)

src/Core/ViewHelper/ViewHelperVariableContainer.php
NamelessCoder: I will take that to heart and spend a bit more than 15 seconds documenting this method and the others ;)

src/Core/ViewHelper/ViewHelperVariableContainer.php
NamelessCoder: Little bit hard to name this because neither "end" nor "peek" really seemed to make sense. I opted instead for a consistent naming with other "get" methods but I see a problem now: proper naming would be `getTopmostDelegateVariableProvider`, not `-Container`!

examples/Resources/Private/Singles/Variables.html
NamelessCoder: This relates to the https://github.com/TYPO3/Fluid/issues/424 vision which calls for `f:argument` to be used for another purpose. I would personally like to keep the distinction that a "variable" is something you insert/assign, whereas an "argument" is something you declare/pass.

> What if one wants to have some template code here, simplified using `f:variable` and then pass everything via `contentAs` as usual?

Then the combined arguments received by the partial would also contain those "variables" you defined while simplifying your template. The pitfall would imho be that if you named such a variable the same as an argument then your nested variable overrules the argument.

Perhaps it is more suitable the other way around: arguments passed in the array take priority over nested variables?

In any case I would not make such a case into an exception or other error.

src/Core/ViewHelper/ViewHelperVariableContainer.php
NamelessCoder: I both agree and disagree: in this specific use case you are right that we are dealing with a sort of `overlay` logic so semantics would imply that naming. But, the methods are actually written to allow a similar approach to be used for other use cases as well - including but not limited to `f:switch` which currently has a problem with nested usages because it doesn't use a stack.

Migrating `f:switch` to this pattern would make the `overlay` naming semantically inappropriate - the next best thing is a "stack of delegates" that can receive variables but aren't the main source of variables for template rendering (is used as transient stack-based storage).

Naming is hard :)

src/ViewHelpers/VariableViewHelper.php
NamelessCoder: I can sure do that :)

NamelessCoder:
As a result of discussion:

* Feature changed so that `f:argument` is used to declare such arguments.
* Made the stack ViewHelper-class-agnostic so it will allow any VH to receive arguments this way as long as it fits the use case (works by evaluating tag content and *then* doing some output, not vice versa).
* Fixed various CGL and documenting comments.

src/ViewHelpers/ArgumentViewHelper.php
```diff
@@ -0,0 +1,68 @@
...
+class ArgumentViewHelper extends AbstractViewHelper
+{
+    use CompileWithContentArgumentAndRenderStatic;
+
+    /**
+     * @return void
+     */
+    public function initializeArguments()
+    {
+        $this->registerArgument('value', 'mixed', 'Value to assign. If not in arguments then taken from tag content');
```
mbrodala: Should we define `$contentArgumentName` instead as suggested by the `CompileWithContentArgumentAndRenderStatic` trait?

src/Core/ViewHelper/ViewHelperVariableContainer.php
```diff
@@ -29,6 +30,69 @@ class ViewHelperVariableContainer
...
+     * Is used in ViewHelpers that assign variables in variable providers in
+     * the stack - as a means to get the variable storage used by the "closest
+     * parent", e.g. when called in `f:argument` used inside `f:render`, will
+     * read the delegate variable provider inserted by that parent `f:render`.
+     *
+     * @return VariableProviderInterface|null
+     */
+    public function getTopmostDelegateVariableProvider()
+    {
+        return end($this->delegates) ?: null;
```
mbrodala: Would it make sense to fall back to `$this`? This way any calling code does not have to care about what and if something was pushed to the stack.

src/ViewHelpers/ArgumentViewHelper.php
```diff
@@ -0,0 +1,68 @@
...
+     * @param RenderingContextInterface $renderingContext
+     * @return null
+     */
+    public static function renderStatic(
+        array $arguments,
+        \Closure $renderChildrenClosure,
+        RenderingContextInterface $renderingContext
+    ) {
+        $delegateVariableProvider = $renderingContext->getViewHelperVariableContainer()->getTopmostDelegateVariableProvider();
+        if ($delegateVariableProvider) {
```
mbrodala: This check could be dropped with the change suggested above.

src/ViewHelpers/ArgumentViewHelper.php
NamelessCoder: That hint is currently not valid as you cannot override properties declared in traits. The larger plan is to refactor content argument support into the base VH class.

src/Core/ViewHelper/ViewHelperVariableContainer.php
NamelessCoder: `$this` is a ViewHelperVariableContainer and this has a different signature for `add` and will throw exceptions. Automatically adding one to the stack isn't sensible either since and empty stack means you used the VH outside of proper context.

src/ViewHelpers/ArgumentViewHelper.php
mbrodala: You can set a value in the constructor though.

src/ViewHelpers/ArgumentViewHelper.php
NamelessCoder: True, but let's avoid that ;)

src/ViewHelpers/ArgumentViewHelper.php
NamelessCoder: This is true but that is not good practice either. We really shouldn't use a new solution here if we expect to refactor it into the base class - since the solution could very easily be incompatible. Don't forget that internally Fluid will just use the first argument we declare.

NamelessCoder:
Ready for another look @mbrodala (sorry about the ping, doing open-stuff-cleanups).

mbrodala:
Looks good from reading (and amazingly simple) but I'd like to conduct at least a simple CLI test to get a feeling for this.

src/ViewHelpers/RenderViewHelper.php
```diff
@@ -121,7 +121,20 @@ public static function renderStatic(array $arguments, \Closure $renderChildrenCl
         $delegate = $arguments['delegate'];
         /** @var RenderableInterface $renderable */
         $renderable = $arguments['renderable'];
+
+        // Prepare a delegate variable provider that will be possible to extract after rendering the child closure.
+        // Any variable defined therein gets used as argument and overrides any argument of the same name.
+        // Note: not using late static binding here is a conscious decision: if late static binding had been used
```
NamelessCoder: Just noticed that this comment is no longer correct, since removing the VH scope argument from the methods it relates to.

NamelessCoder:
Abandoned; would not fit with Fluid 3.0 and is not vital.

NamelessCoder:
Re-opened, there has been some interest for the feature and we should probably re-consider it. My plan is to either cross-port the API parts and functionality to the 3.0 branch, or create an identical feature there, to make 2.x and 3.x straight compatible concerning this particular feature.

s2b:
With the upcoming component features, something like this will be possible:

```xml
<my:component>
  <f:fragement name="description">my <b>description</b></f:fragment>
</my:component>
```

and inside of the component's template:

```xml
<f:slot name="description" />
```

So I'm closing this here in favor of the new upcoming syntax.

PR #436, opened 2019-03-11
[!!!][TASK] Migrate documentation within phpdoc to ReStructuredText

DanielSiepmann:
This allows the TYPO3 documentation team to convert markdown to rst for
proper rendering of ViewHelper Reference.

This also should make it easier for anyone else to generate proper
documentation.


DanielSiepmann:
The rendered result, as it would be on docs.typo3.org with this merge request, can be checked here: https://tmp.daniel-siepmann.de/projects/typo3/typo3fluid-fluid-rendered-reference/typo3fluid/fluid/latest/Index.html

I've uploaded the result so one could check the proper rendering.

DanielSiepmann:
NOTE: This could be considered as breaking, as other parties like Neos might have a setup to render the output in their documentation.

NamelessCoder:
Doesn't affect the output but: I would prefer four-space-indented blocks instead of fenced blocks - mainly because they make it visually block-formatted in raw.

DanielSiepmann:
This would break the output right now. As `trim` is applied to the description: https://github.com/NamelessCoder/fluid-documentation-generator/blob/master/src/ViewHelperDocumentation.php#L38

Either this has to be adjusted, or we need to keep this format, in order to be able to know that this is a code block.

NamelessCoder:
Well spotted! That line should be changed so it only is capable of trimming a single whitespace char from beginning and any number of chars from end of string. That's an oversight and completely my bad!

NamelessCoder:
Even if 5 (4 minus 1 gutter) spaces work the line you found there should still be fixed. This is completely a matter of taste though - either solution produces the same output. One is easier to write, the other is easier to read. Potato, potato, tomato, tomato ;)

I suppose for convenience of contribution let's go with three backticks so we don't have to explain why it's 5 spaces total.

DanielSiepmann:
Then there is nothing left, right? So this could be merged. Maybe one should contact neos / flow whther this change is breaking for them.

NamelessCoder:
> Then there is nothing left, right? So this could be merged. Maybe one should contact neos / flow whther this change is breaking for them.

Don't merge it just yet. I am basing 80% of my decision about whether or not to allow the custom XML tags on what the Neos team has to say about docs rendering. The other 20% are the concerns and good arguments raised by @helhum because I don't entirely disagree, but:

* **Unless it causes breakage or more than light effort to go with Markdown, that's the decision.** As far as I am aware only Neos docs would be in question as to whether it breaks (and whether fences would be super-easy to support seamlessly).
* All points raised by Helmut are valid, but either don't have current use case requirements (our docs rendering doesn't use them anyway and don't necessarily need to) or are nice-to-have. I'm sorry to say I'm not convinced that using the custom XML tags is the right way: to me, it still adds unnecessary complexity because these nodes will show up in raw in XSD files or require some sort of translation at the generating level. I just can't fit this added layer of markup into all the necessary aspects of these docs without it causing even more complexity than markdown would.

Helmut, I don't want you to think I'm blankly disregarding your input about custom needs for documentation. However, I think we will be better served in the long run by establishing a universal format and aiming towards a future where we don't have any markup in the "basic" documented properties and generate XSDs with varying levels of detail, but add a capability to deliver any level of extended documentation, in any desired markup or not, along with each ViewHelper and even each argument. In a way that each package or individual ViewHelper can "tell" Fluid how it wants to be documented. So I hesitate to (re-)introduce custom formats at this time.

At that point we would then extract the Markdown parts of every phpdoc, truncating it to a basic text-only version that's great for XSDs, then moving the examples + output chunks to dedicated doc files.

(Perhaps we should not merge this until there's a solution for self-documenting ViewHelpers the right way that would let Fluid use Markdown and any other package use format XYZ without distinction...)

sypets:
It seems, you have already given the ViewHelper documentation quite some thought. 

I see the advantage of maintaining the documentation in the source. 

The main concern I currently have is:

* it's not easy to contribute: You can write the documentation. But you can't test and see the rendered version until it's live

So, is this now the final result or have you given alternatives some thought, such as:

* having a .rst or .md file for each ViewHelper. That way the docs could be maintained in the source code. It would be easier to write the markup. It would be rendered correctly in the GitHub preview as well. 

I think it is important that we have up-to-date and helpful documentation. For this reason, we should make it as easy as possible to contribute. That is currently not the case. 

Can we talk about these things some more before they are decided for good?

NamelessCoder:
@sypets This is probably not the final result - but it may be a step on the road to the final result. The decision to switch to Markdown isn't final.

> having a .rst or .md file for each ViewHelper. That way the docs could be maintained in the source code. It would be easier to write the markup. It would be rendered correctly in the GitHub preview as well.

That is precisely what I would do - or actually a combination. PHPDOC has very short info and no usage examples but a reference to the RST or MD or whichever format file. Docs generator can choose to either include or ignore these dedicated files when creating the XSD (to make an IDE-friendly or online-friendly version). Docs can be either a single file or a directory and if we decide so, we can also have dedicated example files for each argument.

This is somewhat outside of scope for this particular change though. Currently the PR will be on hold until we know what the consequences, if any, would be to Neos doc generation.

PS: ping me on TYPO3 slack if you want a copy-paste of a LONG discussion history with myself, Daniel and Helmut going over the various points of docs embedding and generation.

DanielSiepmann:
As we are receiving information about "broken rendering" from time to time, I've contacted the neos team within their slack: https://neos-project.slack.com/archives/C051VQPLR/p1558162841051700 in order to finish the decision and decide for a way to go for now.

We can change the decision later on, and follow your approach with dedicated documentation files, which I find useful to. For now this change would solve our current issue.

kitsunet:
Hey there, sorry this totally passed by me until @DanielSiepmann notified me today. I am generally for it, but would like to have a moment to look at our docs rendering, which I am not that familiar with, I guess it should be possible to adapt easily though. :)

kitsunet:
:+1: From my side.
We have a separate parser for grabbing Fluid VH doc blocks anyway, so we can just adjust that to do markdown as soon as we switch over, so that should work fine.

kdambekalns:
What @kitsunet says, fine from the Neos side! That fact the current documentation on our side is empty/broken has a different cause anyways…

DanielSiepmann:
The general discussion during the Fluid 3.0 Workshop resulted in the following:

We will write the initial Fluid documentation using markdown. This initial version will be converted to rst in the end. So Fluid will use rst in the future for his own documentation.

We at least need an `Index.rst` in order to generate a menu of all files. Also we would need to setup a whole new rendering engine for Fluid, to provide a rendering. This would then not fit into the theme of docs.typo3.org.

As rst is set, we also can use rst within the ViewHelper. I'll go over them during the Workshop and update the PR.

Once everything is done, we can add the feature to provide separate documentation files for each ViewHelper, and move information from PHPDoc to these files. I guess these should be the base for rendering in the future. They can be enhanced with information from XSD. These technical, generated information, can be added as valid RST at the end of these files before rendering.

TYPO3 will setup the rendering pipeline for Fluid, once documentation is in place. I'll prepare that already beforehand.

The rendering for ViewHelper is already in place and will work properly, once the PHPDoc is migrated to rst. Right now this is only a TYPO3 ViewHelper Reference. I would like to work on that topic (Separate rst files for ViewHelpers, and separate auto generated ViewHelper References) in the future. Maybe during the Documentation Sprint.

NamelessCoder:
Just to update this with a decision:

I've decided to not merge this until Fluid 3.0 since technically this is a breaking change to documentation generators. In many ways Fluid 3.0 is already a clean slate so I would rather not "annoy" 2.x users with this change.

NamelessCoder:
Handled by https://github.com/TYPO3/Fluid/pull/527

Issue #437, opened 2019-03-13
Feature: Support variable extraction for classes with magic getter


cumuru:
Getting an object property via getter in a template fails if the getter isn‘t a native method but delegates the call via `__call()` instead.
This used to work in the TYPO3 CMS prior to v9 flavor of fluid as the variable provider was overridden by an own implementation that used other means of extracting values internally. 
This class was dropped here: https://review.typo3.org/c/Packages/TYPO3.CMS/+/54028

NamelessCoder:
Solved via #438.

PR #438, opened 2019-03-13
[BUGFIX] Support overloaded methods when extracting variables

cumuru:
Value extraction will now support delegation of getter calls
via PHP‘s __call() magic method. Additionally, extracting a property
that results in a protected getter call will stop raising an error.

Solves #437


NamelessCoder:
Thanks Felix - perfect first PR!

NamelessCoder:
I've re-labeled this as a bug fix because to me, by far the primary thing solved here is that protection against calling protected methods. Adding the support for `__call` is a side effect (but a nice one).

PR #440, opened 2019-03-17
[TASK] Improve performance of hardcoded booleans and empty executions

NamelessCoder:
Improves performance of output escaping when the
output that would have been escaped is empty, as
determined by the compiled state of the output (empty
execution instruction means empty output).

Improves performance throughout when boolean arguments
are used with hardcoded values like ‘1’ for TRUE, which is
very frequently the case. Compiles these cases as direct
value evaluations and casting to boolean, and catches
cases where only a single node needs to be evaluated so
the node’s return value is immediately cast as boolean.

Avoids the relatively expensive 1) conversion of a one-value
array to compiled array and 2) invoking the boolean parser
to consume this array and cast the value. And makes the
compiled template source code file significantly smaller.


tests/Unit/Core/Parser/BooleanParserTest.php
```diff
@@ -41,7 +41,7 @@ public function testSomeEvaluations($comparison, $expected, $variables = [])
         $this->assertEquals($expected, BooleanNode::convertToBoolean($parser->evaluate($comparison, $variables), $this->renderingContext), 'Expression: ' . $comparison);
 
         $compiledEvaluation = $parser->compile($comparison);
-        $functionName = 'expression_' . md5($comparison . rand(0, 100000));
+        $functionName = 'expression_' . md5($comparison . microtime(true));
```
mneuhaus: shouldn't we use random_bytes(10); instead?

mneuhaus: More context: microtime is not that safe to use on windows: 
> microtime() returns the current Unix timestamp with microseconds. This function is only available on operating systems that support the gettimeofday() system call.

tests/Unit/Core/Compiler/NodeConverterTest.php
```diff
@@ -100,37 +100,52 @@ public function getConvertTestValues()
...
+                new BooleanNode(new TextNode('0')),
+                'false'
+            ],
+            'Boolean node with number 1 converts to boolean true' => [
+                new BooleanNode(new NumericNode(1)),
+                'true'
+            ],
+            'Boolean node with number 0 converts to boolean false' => [
+                new BooleanNode(new NumericNode(0)),
+                'false'
```
mbrodala: Can we add one more case for the empty string to make sure this implicit behavior cannot be broken?

NamelessCoder: Sure!

NamelessCoder:
There was indeed a bug hidden under that missing test case - good call @mbrodala. I also wasn't completely pleased with how that converter code looked and how it would look to catch an empty string, so I changed it a bit to trim+unquote any node text content before comparisons. Logic should be much easier to read now.

src/Core/Compiler/NodeConverter.php
```diff
@@ -367,9 +375,34 @@ protected function convertExpressionNode(ExpressionNodeInterface $node)
...
+        // Quick decisions: if there is only one node and it can be determined to be a string/numeric/boolean already
+        // then use it directly.
+        if (count($booleanStack) === 1) {
+            $compiledOnlyNode = $this->convert($booleanStack[0]);
+            // Execution without possible quotation marks, then converted to lowercase for easier comparison.
+            $execution = trim($compiledOnlyNode['execution'], '"\'\\');
+            $lowercaseExecution = strtolower($execution);
+            if (is_numeric($execution) || is_bool($execution)) {
+                $execution = $execution ? 'true' : 'false';
+            } elseif ($lowercaseExecution === 'false' || empty($execution)) {
```
mbrodala: Doing the empty check first should be faster IMO.

src/Core/Compiler/NodeConverter.php
```diff
@@ -367,9 +375,34 @@ protected function convertExpressionNode(ExpressionNodeInterface $node)
...
+                $execution = 'false';
+            } elseif ($lowercaseExecution === 'true') {
+                $execution = 'true';
+            } else {
+                $execution = false;
+            }
+            if ($execution) {
```
mbrodala: Can we do an explicit `!empty($execution)` here to make it clear that we don't expect some string magic here to evaluate the `$execution` in place but only check that it's code is not empty?

NamelessCoder: We can do one better and compare `!== false` and add a comment what happens if it *is* `false` :)

src/Core/Compiler/NodeConverter.php
NamelessCoder: Hmm I don't think that's the case. An explicit comparison checks for one specific case of equality whereas the empty has to check for many. But you may have a point here that these should go *before* the is-numeric and is-bool stuff!

src/Core/Compiler/NodeConverter.php
```diff
@@ -367,9 +375,36 @@ protected function convertExpressionNode(ExpressionNodeInterface $node)
...
+            if (is_numeric($execution) || is_bool($execution)) {
+                $execution = $execution ? 'true' : 'false';
+            } elseif ($lowercaseExecution === 'false' || empty($execution)) {
+                $execution = 'false';
+            } elseif ($lowercaseExecution === 'true') {
+                $execution = 'true';
+            } else {
+                $execution = false;
+            }
+            if ($execution !== false) {
```
mbrodala: In this case we should consider using `null` instead. I still try to avoid the confusion which arises when looking at this section.

src/Core/Compiler/NodeConverter.php
NamelessCoder: It also doesn't matter a whole lot because this code is only for compiling the template so it's very infrequent ;)

src/Core/Compiler/NodeConverter.php
```diff
@@ -283,7 +285,9 @@ protected function convertArrayNode(ArrayNode $node)
         foreach ($node->getInternalArray() as $key => $value) {
             if ($value instanceof NodeInterface) {
                 $converted = $this->convert($value);
-                $initializationPhpCode .= $converted['initialization'];
+                if (!empty($converted['initialization'])) {
```
lolli42, line 288: does not help and has the risk that 0 is considered empty.

src/Core/Compiler/NodeConverter.php
```diff
@@ -283,7 +285,9 @@ protected function convertArrayNode(ArrayNode $node)
...
+                    $initializationPhpCode .= $converted['initialization'];
+                }
```
lolli42, line 290: this has been done in ArrayNode already.

src/Core/Compiler/NodeConverter.php
```diff
@@ -367,9 +375,36 @@ protected function convertExpressionNode(ExpressionNodeInterface $node)
...
     protected function convertBooleanNode(BooleanNode $node)
     {
-        $stack = $this->convertArrayNode(new ArrayNode($node->getStack()));
-        $initializationPhpCode = '// Rendering Boolean node' . chr(10);
-        $initializationPhpCode .= $stack['initialization'] . chr(10);
+        $booleanStack = $node->getStack();
+
+        // Quick decisions: if there is only one node and it can be determined to be a string/numeric/boolean already
+        // then use it directly.
+        if (count($booleanStack) === 1) {
```
lolli42, line 382: this area has been marked as an @todo with 9785955cf59014dcba234f921d5d150bfdc1eec9

src/Core/Compiler/NodeConverter.php
```diff
@@ -107,12 +107,14 @@ public function convert(NodeInterface $node)
     protected function convertEscapingNode(EscapingNode $node)
     {
         $configuration = $this->convert($node->getNode());
-        $configuration['execution'] = sprintf(
-            'call_user_func_array( function ($var) { ' .
-            'return (is_string($var) || (is_object($var) && method_exists($var, \'__toString\')) ' .
-            '? htmlspecialchars((string) $var, ENT_QUOTES) : $var); }, [%s])',
-            $configuration['execution']
-        );
+        if (!empty($configuration['execution']) && strtolower($configuration['execution']) !== 'null' && !is_numeric(trim($configuration['execution'], '\'"'))) {
```
lolli42, line 110: this has been changed for empty string returns with 9785955cf59014dcba234f921d5d150bfdc1eec9
already. further changes depend on booleanNode convert() to be changed.

PR #446, opened 2019-03-26
[TASK] Compile arrays as single definition

NamelessCoder:
Changes the compiling of arrays from being compiled
as a long list of key = value assignments, to a single
array definition. Improves performance slightly but
improves readability of compiled templates even more.

### DIFF
--- a/src/Core/Compiler/NodeConverter.php
+++ b/src/Core/Compiler/NodeConverter.php
@@ -168,22 +168,11 @@ class NodeConverter
                 $this->templateCompiler
             );
 
-            $arguments = $node->getArgumentDefinitions();
-            $argumentInitializationCode = sprintf('%s = array();', $argumentsVariableName) . chr(10);
-            foreach ($arguments as $argumentName => $argumentDefinition) {
-                if (!isset($alreadyBuiltArguments[$argumentName])) {
-                    $argumentInitializationCode .= sprintf(
-                        '%s[\'%s\'] = %s;%s',
-                        $argumentsVariableName,
-                        $argumentName,
-                        var_export($argumentDefinition->getDefaultValue(), true),
-                        chr(10)
-                    );
-                }
-            }
+            $accumulatedArgumentInitializationCode = '';
+            $argumentInitializationCode = sprintf('%s = [', $argumentsVariableName);
 
-            $alreadyBuiltArguments = [];
-            foreach ($node->getArguments() as $argumentName => $argumentValue) {
+            $arguments = $node->getArguments();
+            foreach ($arguments as $argumentName => $argumentValue) {
                 if ($argumentValue instanceof NodeInterface) {
                     $converted = $this->convert($argumentValue);
                 } else {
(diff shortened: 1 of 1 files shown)


NamelessCoder:
The following ZIP contains two versions of the "ButtonBar" partial from the TYPO3 backend extension. One is produced without this patch, the other with this patch.

[Examples_PullRequest_446.zip](https://github.com/TYPO3/Fluid/files/3007781/Examples_PullRequest_446.zip)

The version with the patch applied is ~50 lines of code shorter and has approximately the same number of variable assignments fewer. Both arguments and if-then-else structures should be a lot easier to read in the new version.

NamelessCoder:
PS: our phpunit version is currently not able to handle mocking of interfaces with strict return types. We are working on upgrading that.

NamelessCoder:
Half of this patch is made redundant by https://github.com/TYPO3/Fluid/pull/457 - will adjust.

src/Core/Compiler/NodeConverter.php
```diff
@@ -168,22 +168,12 @@ protected function convertViewHelperNode(ViewHelperNode $node)
...
-                        var_export($argumentDefinition->getDefaultValue(), true),
-                        chr(10)
-                    );
-                }
-            }
+            $accumulatedArgumentInitializationCode = '';
+            $argumentInitializationCode = sprintf('%s = [', $argumentsVariableName);
 
+            $arguments = $node->getArguments();
             $alreadyBuiltArguments = [];
```
mbrodala, line 175: This is unused now AFAIS.

src/Core/Compiler/NodeConverter.php
```diff
@@ -192,24 +182,41 @@ protected function convertViewHelperNode(ViewHelperNode $node)
...
                     $argumentName,
                     $converted['execution']
-                ) . chr(10);
-                $alreadyBuiltArguments[$argumentName] = true;
+                );
+
+                $argumentInitializationCode .= ', ';
+            }
+
+            foreach ($node->getArgumentDefinitions() as $argumentName => $argumentDefinition) {
```
mbrodala, line 195: The change of order won't cause any harm?

NamelessCoder: No, all safe - actually there was a bad behavior before this patch: all arguments' default values would be built every time and then overwritten by subsequent assigning of the arguments. Problem was that `!isset($alreadyBuiltArguments[$argumentName])` would always be true because `$alreadyBuiltArguments` was always undefined (gets defined *after* the loop that checks it).

Now, they will only be built if not already built - just like the original behavior intended.

mbrodala: Makes sense, thanks for the explanation. :-)

mbrodala (approved):
Looks good and works (applied as patch to a TYPO3v10 instance.)

lolli42:
This is a pretty convenient and good patch. It currently suffers from various merge conflicts, but I think we should split it into smaller chunks and merge them one-by-one.

lolli42:
Handling convertArrayNode() in https://github.com/TYPO3/Fluid/pull/763 - This is slightly easier nowadays with current PHP versions since arrays allow a trailing comma on last element.

lolli42:
second pr: https://github.com/TYPO3/Fluid/pull/764 - i'm closing here.

PR #447, opened 2019-04-08
[TASK] Add a note about the special variable layoutName

dogawaf:
Maybe the good way would be to have not any reserved variable names. But in the meantime, this one should be documented.


dogawaf:
Maybe the good way would be to have not any reserved variable names. But in the meantime, this one should be documented.

mbrodala:
Thanks for the note.

PR #450, opened 2019-05-07
[BUGFIX] Ensure escaping of escapable ExpressionNode

NamelessCoder:
Prevents a potential security issue when expression
nodes are used to output variables, in which case,
they would not be properly escaped.

The fix implements escaping interception for these
expression nodes.

https://nvd.nist.gov/vuln-metrics/cvss/v3-calculator?vector=AV:N/AC:H/PR:N/UI:R/S:C/C:L/I:L/A:N/E:F/RL:O/RC:C

### DIFF
--- a/examples/Resources/Private/Singles/Variables.html
+++ b/examples/Resources/Private/Singles/Variables.html
@@ -29,6 +29,7 @@ Direct access of numeric prefixed variable: {123numericprefix}
 
 <!-- Passing arguments to sections/partials -->
 <f:render section="Secondary" arguments="{
+    ternaryCheck: 1,
 	myVariable: 'Nice string',
 	array: {
 		baz: 42,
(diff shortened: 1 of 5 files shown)


mbrodala:
Micro-nitpick: the tag should be `[SECURITY]`.

NamelessCoder:
We've currently only got `BUGFIX`, `TASK`, `DOC`, `FEATURE` and `REMOVAL` as conventions. The same commit is already pushed to 7 other branches but master is protected ;)

mbrodala:
Sure thing, just a heads up that [this is a thing in TYPO3](https://docs.typo3.org/typo3cms/ContributionWorkflowGuide/Appendix/CommitMessage.html#summary-line-first-line). ;-)

NamelessCoder:
I will take that under advisement, completely adopting the commit guidelines, if nothing else then to have a link to point to. This should also go along with a CI implementation to prevent unwanted commit messages.

But for now, please ignore the subject line ;)

PR #451, opened 2019-05-10
[TASK] Refactor and and guard namespace extraction

NamelessCoder:
Refactors the namespace validation from occurring in
the TemplateProcessor, to occur instead when parsing
the template. Behavior remains unchanged.

Also guards several potentially empty matches of
namespaces from being iterated over, causing fatal
PHP errors.


src/Core/Parser/TemplateProcessor/NamespaceDetectionTemplateProcessor.php
```diff
@@ -103,18 +102,20 @@ public function registerNamespacesFromTemplateSource($templateSource)
...
-                $namespaceUrl = trim($set[2], '"\'');
-                if (strpos($namespaceUrl, 'http://typo3.org/ns/') === 0) {
-                    $namespaceUri = substr($namespaceUrl, 20);
-                    $namespacePhp = str_replace('/', '\\', $namespaceUri);
-                } elseif (!preg_match('/([^a-z0-9_\\\\]+)/i', $namespaceUrl)) {
-                    $namespacePhp = $namespaceUrl;
-                    $namespacePhp = preg_replace('/\\\\{2,}/', '\\', $namespacePhp);
-                } else {
-                    $namespacePhp = null;
+            if (!empty($namespaces)) {
```
mbrodala: This can be dropped, an empty array will not be iterated over anyways.

src/Core/Parser/TemplateProcessor/NamespaceDetectionTemplateProcessor.php
```diff
@@ -124,71 +125,36 @@ public function registerNamespacesFromTemplateSource($templateSource)
...
-                    $namespaceAttributesToRemove = [];
-                    foreach ($namespaces as $namespace) {
-                        if (!$viewHelperResolver->isNamespaceIgnored($namespace[1])) {
-                            $namespaceAttributesToRemove[] = preg_quote($namespace[1], '/') . '="' . preg_quote($namespace[2], '/') . '"';
-                        }
-                    }
-                    if (count($namespaceAttributesToRemove)) {
-                        $matchWithRemovedNamespaceAttributes = preg_replace('/(?:\\s*+xmlns:(?:' . implode('|', $namespaceAttributesToRemove) . ')\\s*+)++/', ' ', $matches[0]);
-                        $templateSource = str_replace($matches[0], $matchWithRemovedNamespaceAttributes, $templateSource);
+            } else if (!empty($namespaces)) {
```
mbrodala: Same here.

src/Core/Parser/TemplateProcessor/NamespaceDetectionTemplateProcessor.php
NamelessCoder: The point is two-fold: to skip empty arrays *and* skip `false`, while avoiding use of `is_array` (which admittedly is slightly opinionated but, that's my chosen way of making a potential future switch to `iterable` less painful).

src/Core/Parser/TemplateProcessor/NamespaceDetectionTemplateProcessor.php
```diff
@@ -125,70 +124,35 @@ public function registerNamespacesFromTemplateSource($templateSource)
...
-                        $matchWithRemovedNamespaceAttributes = preg_replace('/(?:\\s*+xmlns:(?:' . implode('|', $namespaceAttributesToRemove) . ')\\s*+)++/', ' ', $matches[0]);
-                        $templateSource = str_replace($matches[0], $matchWithRemovedNamespaceAttributes, $templateSource);
+                $namespaceAttributesToRemove = [];
+                foreach ($namespaces as $namespace) {
+                    if (!$viewHelperResolver->isNamespaceIgnored($namespace[1])) {
+                        $namespaceAttributesToRemove[] = preg_quote($namespace[1], '/') . '="' . preg_quote($namespace[2], '/') . '"';
                     }
                 }
+                if (count($namespaceAttributesToRemove)) {
+                    $matchWithRemovedNamespaceAttributes = preg_replace('/(?:\\s*+xmlns:(?:' . implode('|', $namespaceAttributesToRemove) . ')\\s*+)++/', ' ', $matches[0]);
```
liayn: why do we need \s*+ (non-greedy) operators here?

NamelessCoder: This expression has not changed since Fluid was created so it's not exactly related to this change. But, as far as I can determine, this part ensures that `" ` and `">` are both matched (two variations on how an attribute value may end: space or tag end after attr value).

liayn: Ah right, missed that it was there already. Then leave as is

NamelessCoder:
@mbrodala Could I have you do a quick re-review (reading is enough) and if OK, merge? I need to prepare a new release *today* if we want to solve (work around) the regular expression backtrack issues in CMS.

mbrodala:
Done that now. Given the complexity I'd wanted to at least do some tests but it's probably fine.

bwaidelich:
Hi guys,

> Behavior remains unchanged.

not 100% sure, but it seems that the behavior _did_ change in fact.

Assume:

```html
{namespace f*}{namespace bar}<foo:bar /><bar:foo></bar:foo><foo.bar:baz /><f:format.raw>foobar</f:format.raw>
```

Result before:
```html
<foo:bar /><bar:foo></bar:foo><foo.bar:baz />foobar
```
Result now:
```html
<foo:bar /><bar:foo></bar:foo><foo.bar:baz /><f:format.raw>foobar</f:format.raw>
```

And the reason is that *registered* namespaces took precedence over *ignored* namespaces before and no longer do now.

(see https://github.com/neos/flow-development-collection/issues/1761 for more details)

kdambekalns:
With 2.6.6 this seems to be back for us, see https://github.com/neos/flow-development-collection/issues/1761

kdambekalns:
… and it seems quite logical, looking at https://github.com/TYPO3/Fluid/commit/f8c0fb3753520f78022ddddd52f18466fc4888e1 that went into 2.6.6 as per https://github.com/TYPO3/Fluid/compare/2.6.5...2.6

PR #454, opened 2019-05-15
[BUGFIX] PR-453 Return text node with the full text instead

thomashohn:
Return text node with the full text instead instead of dying.

### DIFF
--- a/src/Core/Parser/TemplateParser.php
+++ b/src/Core/Parser/TemplateParser.php
@@ -588,6 +588,11 @@ class TemplateParser
     protected function textAndShorthandSyntaxHandler(ParsingState $state, $text, $context)
     {
         $sections = preg_split(Patterns::$SPLIT_PATTERN_SHORTHANDSYNTAX, $text, -1, PREG_SPLIT_DELIM_CAPTURE | PREG_SPLIT_NO_EMPTY);
+        if ($sections === false) {
+            // String $text was not possible to split; we must return a text node with the full text instead.
+            $this->textHandler($state, $text);
+            return;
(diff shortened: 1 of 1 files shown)


thomashohn:
It actually also depends/should be merged together with another patch that @NamelessCoder will create where he is fixing the namespacing issue PR-352 I believe - so just let it cook for a couple of days ;-)

src/Core/Parser/TemplateParser.php
```diff
@@ -588,6 +588,11 @@ public function unquoteString($quotedValue)
     protected function textAndShorthandSyntaxHandler(ParsingState $state, $text, $context)
     {
         $sections = preg_split(Patterns::$SPLIT_PATTERN_SHORTHANDSYNTAX, $text, -1, PREG_SPLIT_DELIM_CAPTURE | PREG_SPLIT_NO_EMPTY);
+        if ($sections === false) {
```
mbrodala: Would it make sense to also check `preg_last_error()` in this case?

mbrodala: Probably not because we cannot work with the result unless it's an array.

mbrodala:
Thanks for the change.

src/Core/Parser/TemplateParser.php
NamelessCoder: It would definitely make sense - if we had any logging capabilities. I considered outputting the failure reason along with the captured text but it's not really possible to safely add arbitrary HTML comment or output since this might happen in any inline syntax including an argument.

Issue #453, opened 2019-05-15
Method textAndShorthandSyntaxHandler dies with PHP 7.3.5


thomashohn:
The method  textAndShorthandSyntaxHandler dies with PHP 7.3.5 due to a change in the behaviour of preg_split - if it returns false it should pass through the text instead

mstaatz:
For me i have this issue "throwExceptionsForUnhandledNamespaces" in vendor/typo3fluid/fluid/src/Core/Parser/TemplateProcessor/NamespaceDetectionTemplateProcessor.php

and also with "textAndShorthandSyntaxHandler" in vendor/typo3fluid/fluid/src/Core/Parser/TemplateParser.php

there are a lot usecases like this:

mstaatz:
for me i have this issue "throwExceptionsForUnhandledNamespaces" in vendor/typo3fluid/fluid/src/Core/Parser/TemplateProcessor/NamespaceDetectionTemplateProcessor.php

and also with "textAndShorthandSyntaxHandler" in vendor/typo3fluid/fluid/src/Core/Parser/TemplateParser.php

there are a lot usecases like this:
`$sections = preg_split(Patterns::$SPLIT_PATTERN_SHORTHANDSYNTAX, $text, -1, PREG_SPLIT_DELIM_CAPTURE | PREG_SPLIT_NO_EMPTY);`

`foreach ($sections as $section) {`

this result in many error for the case that preg_split returns "false". foreach will not work with "false"

but this is an other thing

it seems PHP 7.3 uses a newer PCRE Library Version and it will not work the same way

i could fix this after debugging with:
`var_dump(array_flip(get_defined_constants(true)['pcre'])[preg_last_error()]);`

and set the backtrack_limit to a higher value:
`ini_set('pcre.backtrack_limit', 10000000000);`

there is a default of 1000000 .... checked this value on PHP 7.2.18 and 7.3.5

as i am not an Expert with RegEx i could not say what is going wrong .... this are only the results of my debugging

jonakieling:
Hi there,

we have stumbled upon the same issue and checked the template part that was causing it.
In our case we have inline Javascript which leads to the preg_split not finishing.

```
                    <div id="div-with-identation">
                        <script type="text/javascript">
                            function this_is_problematic(some_arg) {
                                # breaks because of the non-matching hashtag
                                # resulting in backtracking
				# backtracking gets catastrophic because of all the whitespaces
                                # at the start of the first line with the hashtag
                            };
[...]

PR #460, opened 2019-06-19
Catch some errors introduced with pcre2 in PHP 7.3

YetiCGN:
See #453


NamelessCoder:
Sorry to reject this one @YetiCGN - but converting a pcre failure to a hard exception is not the way forward. https://github.com/TYPO3/Fluid/pull/454 however is the more graceful way that simply outputs whatever pcre couldn't handle, as text.

Don't feel discouraged from making PRs in the future, the quality is fine, only the solution is too drastic this time!

YetiCGN:
No problem. I already encountered an exception with this so I just made the message more clear about what exactly went wrong. Also sort of a fail-fast approach.

But it's totally fine if you want to handle it in a different way.

PR #466, opened 2019-06-22
[TASK] Implement strict types where possible

manuelselbach:


NamelessCoder:
The build is now green :)

NamelessCoder:
Should now be compatible with API used in TYPO3 CMS with no changes required in their source code. New interfaces are expected for the majority of those features so we can likely deprecate instead of adding type strictness there.

NamelessCoder:
Assimilated by https://github.com/TYPO3/Fluid/pull/470 which will become 3.x

PR #467, opened 2019-07-16
[BUGFIX] Support overloaded methods in StandardVariableProvider

RinyVT:
The bugfix/feature which was introduced in #438 where the function `is_callable` was used, is also necessary in StandardVariableProvider.

### DIFF
--- a/tests/Unit/Core/Variables/StandardVariableProviderTest.php
+++ b/tests/Unit/Core/Variables/StandardVariableProviderTest.php
@@ -7,7 +7,6 @@ namespace TYPO3Fluid\Fluid\Tests\Unit\Core\Variables;
  */
 
 use TYPO3Fluid\Fluid\Core\Variables\StandardVariableProvider;
-use TYPO3Fluid\Fluid\Core\Variables\VariableExtractor;
 use TYPO3Fluid\Fluid\Tests\Unit\Core\Fixtures\ClassWithMagicGetter;
 use TYPO3Fluid\Fluid\Tests\Unit\Core\Fixtures\ClassWithProtectedGetter;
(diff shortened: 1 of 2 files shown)


mbrodala:
Thanks for the fix. Can you add another test in `StandardVariableProviderTest::getAccessorsForPathTestValues()` for this?

tests/Unit/Core/Variables/StandardVariableProviderTest.php
```diff
@@ -7,6 +7,9 @@
  */
 
 use TYPO3Fluid\Fluid\Core\Variables\StandardVariableProvider;
+use TYPO3Fluid\Fluid\Core\Variables\VariableExtractor;
```
mbrodala: This class is unused, please remove this.

RinyVT:
@mbrodala I overlooked the `getAccessorsForPathTestValues` function in your comment. I have added 2 new functions in the `StandardVariableProvider`, not sure if this is also okay.

tests/Unit/Core/Variables/StandardVariableProviderTest.php
RinyVT: I've removed the use statement

mbrodala:
@RinyVT Thanks a lot, looks good from my side.

NamelessCoder:
Thanks Riny! As a side note, in real world usage VariableProviders should not operate directly on objects since that removes the ability to store new variables. The root "container" should always be an array or array-access, or the VariableProvider should be read-only (e.g. JsonVariableProvider).

For use cases that require multiple sources of variables a ChainedVariableProvider exists which can for example use a JsonVariableProvider as basic source and add a mutable StandardVariableProvider on top so that ad-hoc variables can also be assigned.

PR #469, opened 2019-07-31
[BUGFIX] Fix contentAs feature when used on Layout level

smichaelsen:
`<f:render section="MySection" contentAs="content">some content</f:section>` does not work when called from a Layout because the `content` variable is not passed to the section in that case.


mbrodala:
Thanks for the report and fix. Would you be able to add a functional test case for this? You can have a look at the [existing tests](https://github.com/TYPO3/Fluid/blob/2d2d5db7268f57b9f8e83e0d3c474f7603b54cab/tests/Functional/ExamplesTest.php).

smichaelsen:
Thanks for your feedback, Mathias. I fear I can't add a test at the moment (deadlines and stuff). Would that block my PR from being merged?

mbrodala:
Of course not. I'll have a look at this.

mbrodala:
@NamelessCoder I've added a test for this now which looks basically correct but still fails. Can you have a look?

src/View/AbstractTemplateView.php
```diff
@@ -221,6 +221,9 @@ public function renderSection($sectionName, array $variables = [], $ignoreUnknow
         if ($this->getCurrentRenderingType() === self::RENDERING_LAYOUT) {
             // in case we render a layout right now, we will render a section inside a TEMPLATE.
             $renderingTypeOnNextLevel = self::RENDERING_TEMPLATE;
+            foreach ($variables as $key => $value) {
```
NamelessCoder: This is the dangerous line. It operates on the not-cloned RenderingContext retrieved from `getCurrentRenderingContext` so assigning variables here, bleeds the variables to other sections and the layout's subsequent code.

NamelessCoder:
```
        $renderingContext = $this->getCurrentRenderingContext();
        $renderingTypeOnNextLevel = $this->getCurrentRenderingType();
        if ($this->getCurrentRenderingType() === self::RENDERING_LAYOUT) {
            // in case we render a layout right now, we will render a section inside a TEMPLATE.
            $renderingTypeOnNextLevel = self::RENDERING_TEMPLATE;
            if (empty($variables)) {
                $variables = $renderingContext->getVariableProvider()->getAll();
            }
        }
        $variableProvider->getVariableProvider()->getScopeCopy($variables)
        $renderingContext = clone $renderingContext;
        $renderingContext->setVariableProvider($variableProvider);
```

Issue #471, opened 2019-08-03
Problematic regular expression in StandardVariableProvider for resolving sub variable references


FelHell:
There seems to be an issue with multiple variables in the arguments of a `<f:render partial="...">` tag. The following Fluid template demonstrates the problem:
```html
<f:variable name="entity" value="{text_lorem_ipsum: 'Lorem Ipsum Dolor Sit Amet'}" />
<f:variable name="key1" value="lorem" />
<f:variable name="key2" value="ipsum" />

Correct output:
<f:render partial="DummyPartial" arguments="{text: entity.text_lorem_ipsum}" />

Correct output:
<f:render partial="DummyPartial" arguments="{text: '{entity.text_lorem_{key2}}'}" />

No output (text is NULL):
<f:render partial="DummyPartial" arguments="{text: '{entity.text_{key1}_{key2}}'}" />
```
The `DummyPartial` simply outputs `{text}`.

I think that the regular expression doesn't work here for multiple variables:
https://github.com/TYPO3/Fluid/blob/865890c3cffaf8d5f6aa44295d42a84a103895ec/src/Core/Variables/StandardVariableProvider.php#L258

NamelessCoder:
Just for the record: this is a known limitation and it is unlikely to be solved by modifying the regular expression. Business logic around the extraction also has to change - and I really hesitate to make this truly recursive (our history with regular expression as solution is less than ideal...)

Also for the record: this will be a complete non-issue in Fluid 3.0 which will support any level and combination of dynamic variable access, including use of inline VH inside.

albe:
Yeah, this solves only one specific corner-case of variables within expressions. Something like `_{key1{key2}}` still wouldn't work (though arguably that is stretching things a bit anyway). But IMO the "solution" (making the inner match greedy) is minimally invasive and also improves the regex performance for common match cases. So it could still be considered, but ofc decision is up to @NamelessCoder

NamelessCoder:
Apologies to @FelHell for rejecting this. Don't let this discourage you from future pull requests!

Reasons given:

* Inability to extract multiple same-level variables is a known limitation of Fluid 2.x and earlier
* Changing this also implies a need to change a lot more regarding how variables are resolved
* It is a non-issue in the next major version of Fluid, 3.0

PR #472, opened 2019-08-03
[BUGFIX] Fix RegExp in resolveSubVariableReferences

FelHell:
Resolves: #471

### DIFF
--- a/src/Core/Variables/StandardVariableProvider.php
+++ b/src/Core/Variables/StandardVariableProvider.php
@@ -255,7 +255,7 @@ class StandardVariableProvider implements VariableProviderInterface
     protected function resolveSubVariableReferences($propertyPath)
     {
         if (strpos($propertyPath, '{') !== false) {
-            preg_match_all('/(\{.*\})/', $propertyPath, $matches);
+            preg_match_all('/(\{.*?\})/', $propertyPath, $matches);
             foreach ($matches[1] as $match) {
                 $subPropertyPath = substr($match, 1, -1);
                 $propertyPath = str_replace($match, $this->getByPath($subPropertyPath), $propertyPath);


src/Core/Variables/StandardVariableProvider.php
```diff
@@ -255,7 +255,7 @@ public function getAccessorsForPath($propertyPath)
     protected function resolveSubVariableReferences($propertyPath)
     {
         if (strpos($propertyPath, '{') !== false) {
-            preg_match_all('/(\{.*\})/', $propertyPath, $matches);
+            preg_match_all('/(\{.*?\})/', $propertyPath, $matches);
```
liayn: Would this also work for `<f:render partial="DummyPartial" arguments="{text: '{entity.text_{array1.some{key1}}_{key2}}'}" />` 
I think we can have arbitrary nesting here

albe: Probably not, but this likely also didn't work beforehand. The way Fluid handles nesting IIRC is by parsing out the inner part of an expression, then matching that again recursively.
So with the given example (simplified):
`... arguments="{text: '{entity.text_{array1.some{key1}}_{key2}}'}"` -argument syntax match-> `text: '{entity.text_{array1.some{key1}}_{key2}}'` -object/array syntax match-> `entity.text_{array1.some{key1}}_{key2}` -> and here it should fail with greedy as with non greedy matching, because the results are:
```
greedy: array(2) {
  [0]=>
  array(2) {
    [0]=>
    string(18) "{array1.some{key1}"
    [1]=>
    string(6) "{key2}"
  }
  [1]=>
  array(2) {
    [0]=>
    string(18) "{array1.some{key1}"
    [1]=>
    string(6) "{key2}"
  }
}
non-greedy: array(2) {
  [0]=>
  array(1) {
    [0]=>
    string(26) "{array1.some{key1}}_{key2}"
  }
  [1]=>
  array(1) {
    [0]=>
    string(26) "{array1.some{key1}}_{key2}"
  }
}
```
Those are both wrong and will not lead to proper evaluation of the expression. Fluid 3 should probably solve this mess once and for all 🤞

NamelessCoder:
Apologies to @FelHell for rejecting this. Don't let this discourage you from future pull requests!

Reasons given:

* Inability to extract multiple same-level variables is a known limitation of Fluid 2.x and earlier
* Changing this also implies a need to change a lot more regarding how variables are resolved
* It is a non-issue in the next major version of Fluid, 3.0

PR #475, opened 2019-09-09
[BUGFIX] Handle conflict with registered/ignored namespaces

NamelessCoder:
Fixes a regression in namespace resolving behavior,
wherein a conflicting ignored namespace (e.g. "f*")
incorrectly caused the registered namespace to be
ignored (e.g. "f:" would be ignored).

References: #451
References: https://github.com/neos/flow-development-collection/issues/1761
Signed-off-by: Claus Due <claus@namelesscoder.net>


tests/Functional/Cases/Parsing/NamespaceRegistrationTest.php
```diff
@@ -0,0 +1,37 @@
...
+    /**
+     * @return array
+     */
+    public function getTemplateCodeFixturesAndExpectations()
+    {
+        return [
+            'Ignoring namespaces without conflict with registered namespace' => [
+                '{namespace z*}{namespace bar}<zoo:bar /><bar:foo></bar:foo><zoo.bar:baz /><f:format.raw>foobar</f:format.raw>',
+                $this->variables,
+                ['<zoo.bar:baz />'],
```
NamelessCoder: Reading these test cases:

* First array is an array of (string) values that *must appear in output when executed*.
* Second array is an array of (string) values that *must NOT appear in output when executed*.

albe:
@NamelessCoder any information anywhere to be found about why this was reverted? I assume a regression from the fix to the regression was observed?
Any way we can help out to find a solution on how to proceed? Maybe we just need to adjust things on our end after all, but need to know what is the status.

Thanks a lot for your answer and effort in advance!

NamelessCoder:
Now where's that red-cheeks emote...

The patch was never supposed to be reverted but it was, due to an unfortunate series of events.

An unrelated regression was supposed to be reverted but 1) I prepared the PR to revert changes since 2.6.1 but did not notice this patch was included, 2) there was some time pressure to prepare a new release because of the regression, and finally 3) the unrelated revert was also missed during review.

Unfeasible magic method support was the cause of the regression and is completely unrelated to namespace validity.

Obviously I will restore this commit and apologise for the incorrect revert!

(PS: you may want to check if you have extensive use of magic methods without explicit getters since this is the support that had to be abandoned. Versions 2.6.3, 2.6.4 and 2.6.5 are somewhat unpredictable when magic methods are present on objects used as template variables)

albe:
Thanks a lot for the explanation! Nothing to be ashamed of, things happen! Glad it's just a mistake that can be undone again :)

kdambekalns:
Hey @NamelessCoder – there you go: 🤭 

😎 Thanks for clearing that up, and as @albe said, don't worry.

Issue #476, opened 2019-09-13
Wrong type annotation in NamespaceDetectionTemplateProcessor


andreasfernandez:
The property `NamespaceDetectionTemplateProcessor::$localNamespaces`is annotated as `array()` which breaks with newer releases of phpdocumentor/reflection-docblock used in TYPO3 v10:

> A type is missing in an array expression

xdebug reveals that the exception is thrown due to this annotation.

NamelessCoder:
Merged.

PR #477, opened 2019-09-13
[BUGFIX] Fix broken annotation of property (#476)

andreasfernandez:
The annotation of the property `NamespaceDetectionTemplateProcessor::$localNamespaces` is changed to being `@var array` to fix issues with the recent version of `phpdocumentor/reflection-docblock`.


mbrodala:
Thanks for the fix!

lexwebmaker:
Hello. Do you know when a version with this fix will be released?

PR #478, opened 2019-09-13
[BUGFIX] Prevent sprintf warning in cached ternary with modulo

NamelessCoder:
This patch primarily fixes a bug in TernaryExpressionNode
in which an expression that contains a modulo (%) would
be concatenated with the format string instead of inserted
as parameter, causing the parameter count for sprintf to
be incorrect. By adding the expression as format argument
the parameter count for sprintf will always be consistent.

Further, the patch also fixes a precursor problem that meant
functional tests did not test cached executions and adjusts
two other functional test to remove code duplication.

And allows functional tests to pass a string instead of an
array of expected strings, which makes the test match the
entire output instead of a substring, making the "not expected"
argument redundant when "expected" is a string. Cleans up
the ternary expression node test accordingly.


lolli42:
This is superseded by https://github.com/TYPO3/Fluid/pull/619

Issue #479, opened 2019-09-26
Support for overloaded methods breaks property access of instances with magical __call


benwick:
This commit is very problematic: https://github.com/TYPO3/Fluid/commit/d8ccbcddc5c01ad3e6d9e74d3ddab919f8a44f20

It changed the test for a getter accessor from `method_exists` to `is_callable`. The problem is: `is_callable` returns always `true` for class instances with a magical `__call` method. So, if you try to access a public property of a class instance with `_call`, it will always fail trying to access the value with a getter method...

Example: In some DCE templates I access properties from the TyposcriptFrontendController like this: `{TSFE.metaCharset}`.

The result is this: 
![image](https://user-images.githubusercontent.com/2118579/65696945-6bc1b680-e07a-11e9-945f-fba8d5bfe682.png)

benwick:
One solution could be: Make this condition (https://github.com/TYPO3/Fluid/blob/master/src/Core/Variables/StandardVariableProvider.php#L353) the first one instead of the last...

NamelessCoder:
Hi @benwick,

The reason for having the "can we call a getter?" check first:

* Historically, the main data source for Fluid templates has been model objects with getters.
* Which use getter methods to extract values.
* Which is why this check is done first.

The second check is for array access since this is the second-most frequent way a variable can be extracted. Last comes the property check, since this is the least frequent use case (and on a side note, TYPO3 CMS is moving away from these).

Moving the condition for public property to the top means we break another use case - if a getter method exists and a public property exists which both match the name, we would be changing that priority.

Another way could be to add an additional check here: https://github.com/TYPO3/Fluid/blob/master/src/Core/Variables/StandardVariableProvider.php#L347 to check `&& !property_exists($subject, $propertyName)` so a (virtual) getter method will not be called if a public property exists. I suppose we can live with one additional function call although it adds it the most frequent use case.

(PS: overloading is, generally speaking, bad practice - as far as possible you should not use it as a solution in your own code!)

NamelessCoder:
Looks like we can't even do `!property_exists()` since this doesn't respect accessibility. In any case, it would be best if we can come up with a solution that doesn't change priority of current checks.

PR #480, opened 2019-10-09
[BUGFIX] Use magic getter method as last attempt

NamelessCoder:
Prevents issues with greedy magic methods being
called when a (public) property of the same name
exists in the class.

Close: #479


janslu:
Just wanted to chime in and say that this patch fixed my problem with accessing TSFE public properties in due elements. I'd love to have this build in a release.

src/Core/Variables/StandardVariableProvider.php
```diff
@@ -23,6 +23,14 @@ class StandardVariableProvider implements VariableProviderInterface
      */
     protected $variables = [];
 
+    /**
+     * Internal array of ['className' => ['getFoo', 'getBar'], ...]
+     * storing all names of getter methods for a given class.
+     *
+     * @var array
+     */
+    protected static $gettersByClassName = [];
```
mbrodala: Won't this increase the risk of memory exhaustion?

NamelessCoder: It improves the performance by sacrificing memory, sure - but the alternative is to call `get_class_methods` (and copying/releasing that memory instead) so many times that it's probably better to store a couple of items here (in common use cases there will be few actual classes which are accessed in templates, e.g. models and special configs).

Part of the problem is that accessors are detected every time you *evaluate* an ObjectAccessorNode - that means once per variable access even inside loops and that's where it gets really costly.

Side note: I am also working on an optimisation that stores the accessors used by a certain ObjectAccessorNode instead of detecting them each time the accessor is read. I'm doing this because I profiled my way to determining that the accessor detection can be quite heavy and causes a lot of repeated function calls that return the same values. A bit like this use case in that storing these return values at the cost of memory has a pretty significant performance benefit ;)

mbrodala: OK, thanks for the insights. :+1:

src/Core/Variables/StandardVariableProvider.php
```diff
@@ -380,10 +398,44 @@ protected function isExtractableThroughAsserter($subject, $propertyName)
...
+     *   coalesce expression to determine if a getter method exists *AND* is public *AND*
+     *   is not going to be handled by __call.
+     *
+     * For ease of use (to avoid repeating get_class calls) the method returns the class
+     * name of the instance as returned by get_class($instance).
+     *
+     * @param object $instance
+     * @return string
+     */
+    protected function studyObject($instance)
```
mbrodala: I'd name this `populateGettersByClassName()` instead to increase the readability of the code. I understand that you don't directly return the map to reduce memory usage but right now the relation of this method and the static property is unclear without analyzing the code.

NamelessCoder: Good choice.

NamelessCoder:
Thanks @mbrodala :)

Issue #485, opened 2019-10-22
Define & document order or accessors


ohader:

NamelessCoder:
Merged, closing.

Issue #483, opened 2019-10-21
Protected property prevents using magic __call getter


philipmasser:
This commit 308f1af1a7526497f35b755d15ee87eef7fb837b adds support for (public) properties in StandardVariableProvider. However, it doesn't check the accessability. 

My project broke because of this. I happen to have a protected property `$name`. For the actual access, I rely on `__call`.

However, since 308f1af1a7526497f35b755d15ee87eef7fb837b I get a critical error `Cannot access protected property [...]Group::$name`

NamelessCoder:
Hi @philipmasser,

This problem is fixed by https://github.com/TYPO3/Fluid/commit/b4ba01c67a56ad24d2a164c47fa4646217584ad4 which is in 2.6.5 which was released right around the same time you wrote this issue ;)

philipmasser:
Hi,

thanks for your reply.

But the commit you mentioned is just a small code cleanup of the problematic commit. The problem still persists. In fact, it only occurred after I upgradet to 2.6.5!

Testcase:

```
class Test extends AbstractEntity {
    protected $name;
    public function __construct() {
        $this->name = 'Hello, World!';
    }
    public function __call($name, $arguments) {
        if ($name == 'getName') return $this->name;
        return '';
    }
}
```

When accessing `{obj.name}` in a fluid template where `obj` is an instance of class `Test`, I get an critical error *Cannot access protected property* as of Fluid 2.6.5.

I think a check is necessary for the case of `ACCESSOR_PUBLICPROPERTY` if the property is, well, public.

I hope you can reconsider this, as for me, the issue is not closed

My other option would be to rewrite my complete domain/repository logic

philipmasser:
For now I found a workaround: use a magic __getter:

```
public function __get($name) {
    if (property_exists($name)) return $this->$name;
}
```

NamelessCoder:
That is indeed a different problem, but may not be one we can fix - whichever way I think about it, we end with a bad solution:

* We can't merely change the order of these checks because `__call` is so promiscuous it says "yep, I will call anything!".
* Checking for property being public requires use of reflection which is inherently slow (and violates the visibility constraint in order to work).
* Removing `__call` (check for `is_callable` initially) support to restore previous behavior would break your use case.

[...]

PR #486, opened 2019-10-22
[TASK] Add tests in order to ensure accessor orders

ohader:
Related: #485


ohader:
Before continuing on this change, precessing order really should be defined and documented first (e.g. public properties, then getters, then `__call`, ...)

tests/Functional/Cases/Rendering/DataAccessorTest.php
```diff
@@ -0,0 +1,207 @@
...
+                    'publicValue' => 'publicValue', // @todo Probably regression?!
+                ],
+            ],
+            // accessing using everything
+            [
+                $this->createObjectWithEverything(),
+                [
+                    'privateValue' => 'privateValue@getPrivateValue()',
+                    'protectedValue' => 'protectedValue@getProtectedValue()',
+                    'publicValue' => 'publicValue@getPublicValue()', // @todo Probably regression?!
```
NamelessCoder: Getter takes priority and did before, also for public properties - so this probably is not a regression.

tests/Functional/Cases/Rendering/DataAccessorTest.php
```diff
@@ -0,0 +1,207 @@
...
+            ],
+            // accessing using magic method __call()
+            [
+                $this->createObjectWithMagicCall(),
+                [
+                    'privateValue' => null,
+                    'protectedValue' => null,
+                    'publicValue' => 'publicValue@__call("getPublicValue")',
+                ],
+                'Cannot access private property class@anonymous::$privateValue' // @todo Probably regression?!
```
NamelessCoder: This is the pain point. If the property exists, public or not, the `__call` method is not consulted. See https://github.com/TYPO3/Fluid/issues/483 - TL;DR we likely can't do this without Reflection and I *really* don't want to start using Reflection for extracting properties. Issue has three workarounds; `__call` support was introduced recently and has no truly established convention (my recommendation is to avoid it).

tests/Functional/Cases/Rendering/DataAccessorTest.php
```diff
@@ -0,0 +1,207 @@
...
+                    'publicValue' => 'publicValue',
+                ],
+            ],
+            // accessing using getters
+            [
+                $this->createObjectWithGetter(),
+                [
+                    'privateValue' => 'privateValue@getPrivateValue()',
+                    'protectedValue' => 'protectedValue@getProtectedValue()',
+                    'publicValue' => 'publicValue@getPublicValue()', // @todo Probably regression?!
```
NamelessCoder: See below; probably not a regression.

tests/Functional/Cases/Rendering/DataAccessorTest.php
```diff
@@ -0,0 +1,207 @@
...
+            ],
+            // accessing using properties directly
+            [
+                $this->createObjectWithProperties(),
+                [
+                    'privateValue' => null,
+                    'protectedValue' => null,
+                    'publicValue' => 'publicValue',
+                ],
+                'Cannot access private property class@anonymous::$privateValue', // @todo Probably regression?!
```
NamelessCoder: To my knowledge, not a regression. Attempting to access private or protected properties without getter methods has always triggered a PHP error (we just never tested for it).

ohader:
Alright, I've adjusted test-cases in order to keep `v2.6.1` behavior with `__call()` support on top... Accessor order is ensured using functional tests. Real test fixture classes are used now instead of using anonymous classes...

src/Core/Variables/StandardVariableProvider.php
```diff
@@ -356,20 +356,27 @@ protected function detectAccessor($subject, $propertyName)
...
+            $lowerCasePropertyName = strtolower($propertyName);
+            $getter = 'get' . $lowerCasePropertyName;
             if (static::$gettersByClassName[$className][$getter] ?? false) {
                 return self::ACCESSOR_GETTER;
             }
             if ($this->isExtractableThroughAsserter($subject, $propertyName)) {
                 return self::ACCESSOR_ASSERTER;
             }
-            if (property_exists($subject, $propertyName)) {
+            $propertyNames = array_keys(get_object_vars($subject));
```
NamelessCoder: I would really like to avoid this one - the code here executes thousands and thousands of times in a normal set of Fluid templates, so if we every time extract array keys and get object vars for the same object, that will add a lot of pressure. Perhaps it can be done by the method that populates the array of getters per class name.

src/Core/Variables/StandardVariableProvider.php
```diff
@@ -356,20 +356,27 @@ protected function detectAccessor($subject, $propertyName)
...
+            // __get() magic methods are considered as public property
+            if (in_array($propertyName, $propertyNames, true) || method_exists($subject, '__get')) {
```
NamelessCoder: Technically, adding support for `__get` changes the behavior again because now, a class which implements `__get` will never be able to support `__call`. There is also the question if (virtual) public properties should take priority over getter methods, which would also change the behavior.

NamelessCoder:
Correction, it is `is_callable` that returns `true` for any name if `__call` exists.

This is a tough one. I think the code is moving in the wrong direction and have a proposal how to reverse that.

I'll try to summarise chronology and cause:

* Last year, the public method deprecation trait was added to TYPO3
* This trait allows changing a public method to protected and still be able to call it (and log the call)
* But has introduced the problem that `is_callable` always returns `true` for any method on any class implementing the trait.
* Among other classes, the trait is implemented on TypoScriptFrontendController
* Then certain previously public methods were made protected or private and the only way to call them externally is via `__call`

Side effects:

* `is_callable` is no longer trustworthy for any method on those classes.
* `method_exists` is not of much help because it includes protected/private methods, whose visibility has now changed.
* Checking if a method truly exists thus now requires a combination of `get_class_methods` and `method_exists` plus checking if `__call` is defined.

This caused the following:

* Because the TSFE class is one of those implementing the trait and this class is frequently assigned to Fluid templates, a report was received asking to implement `is_callable` (at which time I was not aware it always returns `true` if `__call` exists).
* This caused accessing public properties on TSFE to no longer work in Fluid because 1) getters have and always had priority, and 2) `is_callable` now reported that any theoretical getter method exists.
* Following this, a number of commits were made attempting to restore the previous behavior introduced by this situation.

My evaluation:

I have to come to the conclusion that using `__call` to change public methods to protected and still be able to call them, and log deprecation, is not ideal:

* Because the visibilty of a method was changed without really changing it.
* Because commonplace methods like `is_callable` can no longer guard against incorrect method calls.

I lean towards making `method_exists` the check Fluid uses and ignore presence of `__call`. This would restore the 2.6.1 behavior - and not attempt to introduce support for overloaded methods.

And ideally TYPO3 wouldn't use overloaded methods to violate accessibility of methods in order to log a deprecation. As an alternative I suggest removing runtime logging of these deprecations and using static analysers to determine if any code in a site is attempting to access a property that will in the future be made protected. Utilities like `phpstan` are capable of such custom checks. That avoids problems with Fluid and every other code that attempts to extract properties dynamically (like ObjectAccess).

NamelessCoder:
https://github.com/TYPO3/Fluid/pull/489 will revert the changes but I'd still like to merge all (relevant) tests from this PR. Expect some upcoming refactoring ;)

Issue #484, opened 2019-10-22
Property with special name not accessible


nicodh:
In the ExtensionBuilder we have a property "nameSpaceName" which is needed when generating the PHP code for Classes etc. Since 53eb2aa5 this property is not rendered any more, before that commit it is fine...
We see it in the functional tests and when testing locally:
Mismatch: 
Line 1:namespace FIXTURE\TestExtension\Tests\Unit\Controller;
Line 1:namespace \Tests\Unit\Controller;
in File /home/travis/build/FriendsOfTYPO3/extension_builder/Tests/Fixtures/TestExtensions/test_extension/Tests/Unit/Controller/MainControllerTest.php

So the variable in 
`
namespace {extension.nameSpaceName}\Tests\Unit\Controller;
`
is empty.

nicodh:
My fault: This was a case sensitive/insensitive problem which did'nt show up before. But the variables are parsed correct.

PR #489, opened 2019-10-24
[BUGFIX] Roll back magic method support

NamelessCoder:
Undoes the initial added support for `__call` by switching to `is_callable`
due to unforeseen side effects (method returns `true` for any method
name and causes every theoretical getter method to "exist").

References: #486, #485, #484, #483, #479, #437


src/Core/Variables/StandardVariableProvider.php
```diff
@@ -299,20 +291,15 @@ protected function extractSingleValue($subject, $propertyName, $accessor = null)
      */
     protected function canExtractWithAccessor($subject, $propertyName, $accessor)
     {
+        $class = is_object($subject) ? get_class($subject) : false;
```
mbrodala: Micro nitpick: this should be `$className`.

Issue #405, opened 2018-08-23
Update requirement of PHP to 7.2+


NamelessCoder:
This is a RFC to gather opinions about upgrading the PHP minimum version of Fluid to 7.2, in order to gain access to nullable types and nail down the entire API with strict return/argument types.

Currently Fluid supports PHP >= 5.5 so this is a rather dramatic upgrade. It would however in my opinion not make much sense to upgrade the requirement unless it is to 7.2 because of nullable types. Without those, there would still be (too many) places where a return type could still not be declared.

It might also be reasonable to compromise and *not* declare strict types in some areas of the API, e.g. ViewHelper methods which are guaranteed to be implemented in hundreds or even thousands of classes in production right now.

This change would obviously be breaking, thus the RFC. In particular I am interested in hearing if certain parts of the API (aside from VH methods as already mentioned) your particular Fluid-using framework would like to exempt.

And I am equally interested in hearing possible arguments for why PHP 7.0 for example would be sufficient.

Thanks in advance for any and all input!

kitsunet:
No problem, please ping whenever you seem fit :)
We are currently at 7.1 and will probably update with next major in about a year. I think it's fine though, we can stay at the lastest < 7.2 version until then.
Generally I would like to see type hints. We started to add them in the framework as well. I currently don't know any particular place besides VH arguments that might be problematic. Everything else can probably be adapted.

YetiCGN:
I've noticed you have set the minimum required PHP version in the [develop branch](https://github.com/TYPO3/Fluid/blob/6815505dde6042a5521de8bec228e5b91bf4baa5/composer.json#L7) already. Might I suggest you use a more constrained version range? Right now you're declaring compatibility with PHP 8.0, 9.0 and above - even though you probably haven't tested this package on those platforms. The version constraint `^7.2.0` is much more appropriate, since it at least sets the compatible version to `<= 8.0.0`. 

You can always increase the supported PHP version, but if you try to decrease it in a later version, people will simply not get any updates but be stuck on an old version of TYPO3/Fluid (or `typo3/fluid-engine`, then) which won't work with their PHP version either.

tl;dr: Please don't use version constraints without a sensible upper bound.

NamelessCoder:
True and acknowledged, @YetiCGN!

PR #490, opened 2019-10-29
[TASK] Declare PHP ext-json as dependency

NamelessCoder:
Since Fluid provides a JSON-based VariableProvider
it technically requires ext-json.


NamelessCoder:
Probably shouldn't be merged until Fluid 3.0, could theoretically cause a broken install which shouldn't happen in a minor/bugfix version.

See also #405 which likewise should happen for Fluid 3.0

lolli42:
Hey. I'd argue the JSONVariableProvider is a relatively rare usecase.
Imho a hard requirement is not justified, even though ext-json is one of those PHP extensions that is available in 99% of cases by default anyways. TYPO3 CMS has a hard requirement to it anyways.
Still, Fluid only needs it in relatively rare cases.
I updated the PR to make it a suggestion instead in v2.

lolli42 (approved):

sbuerk (approved):

PR #491, opened 2019-11-04
[BUGFIX] Handle conflict with registered/ignored namespaces (#475)

NamelessCoder:
Restores incorrect revert https://github.com/TYPO3/Fluid/commit/2b5ffd5ef66a557b4bf01fd74d65c2df6fec1587
which was unintentionally reverted as part of #489.

Fixes a regression in namespace resolving behavior,
wherein a conflicting ignored namespace (e.g. "f*")
incorrectly caused the registered namespace to be
ignored (e.g. "f:" would be ignored).

References: #451, #475, #489
References: https://github.com/neos/flow-development-collection/issues/1761

### DIFF
--- a/tests/Functional/Cases/Parsing/NamespaceRegistrationTest.php
+++ b/tests/Functional/Cases/Parsing/NamespaceRegistrationTest.php
@@ -1,0 +1,37 @@
+<?php
+namespace TYPO3Fluid\Fluid\Tests\Functional\Cases\Parsing;
+
+use TYPO3Fluid\Fluid\Tests\Functional\BaseFunctionalTestCase;
+
+/**
+ * Class NamespaceRegistrationTest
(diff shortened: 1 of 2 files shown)


tests/Functional/Cases/Parsing/NamespaceRegistrationTest.php
```diff
@@ -0,0 +1,37 @@
...
+     * @var array
+     */
+    protected $variables = [];
+
+    /**
+     * @return array
+     */
+    public function getTemplateCodeFixturesAndExpectations()
+    {
+        return [
```
mbrodala: I know it's just a revert but no matter how long I look at this test, I simply don't get the intention. Especially that the output which is expected and which is not expected is basically the same.

Issue #449, opened 2019-05-02
Different results of template parser for arrays on different operating systems


dlubitz:
I had issues with shorthand viewhelpers. The arguments I passed as an array where handled as a string at the end. I dug deeper into the issue and found the root cause in the regex `\TYPO3Fluid\Fluid\Core\Parser\Patterns::$SPLIT_PATTERN_SHORTHANDSYNTAX_ARRAY_PARTS` for splitting the arrays recusive. 

My viewhelper code: 
```
{f:translate(id: 'some.id.to.translate', package: 'Vendor.Package', arguments: {0: some.value})}
``` 
which leads to the exeption: 
```
The argument "arguments" was registered with type "array", but is of type "string" in view helper
```

**I could reproduce** this error **just on a FreeBSD** server and could find related bug reports.

Related issues: 
- https://forge.typo3.org/issues/87958
- https://forge.typo3.org/issues/87573

---
The [regex pattern](https://github.com/TYPO3/Fluid/blob/dbe74a61ba1d67f7d957e714ab42273511813987/src/Core/Parser/Patterns.php#L231) fetches there for any reason `VariableIdentifier` and `Subarray`. Which leads to the issue described above, because the Subarray gets handled as ObjectAccessorNode.

https://github.com/TYPO3/Fluid/blob/bcd0ff44dd3c4f9d0f37097b0237db6319c2b181/src/Core/Parser/TemplateParser.php#L688-L698

### Test code
```
$string = 'id: \'some.id.to.translate\', package: \'Vendor.Package\', arguments: {0: some.value}';
preg_match_all(\TYPO3Fluid\Fluid\Core\Parser\Patterns::$SPLIT_PATTERN_SHORTHANDSYNTAX_ARRAY_PARTS, $string, $matches, PREG_SET_ORDER);
print_r($matches);
```
### Expected result:
```
Array
(
    [0] => Array
        (
            [0] => id: 'some.id.to.translate'
            [ArrayPart] => id: 'some.id.to.translate'
            [1] => id: 'some.id.to.translate'
            [Key] => id
            [2] => id
            [QuotedString] => 'some.id.to.translate'
            [3] => 'some.id.to.translate'
        )

    [1] => Array
        (
            [0] => package: 'Vendor.Package'
            [ArrayPart] => package: 'Vendor.Package'
            [1] => package: 'Vendor.Package'
            [Key] => package
            [2] => package
            [QuotedString] => 'Vendor.Package'
[...]

PR #493, opened 2019-11-07
[BUGFIX] Build subarray if found in regex, regardless of other matches

smichaelsen:
To solve #449, we can change the order of result handling. Tests were added, which test for the existing behavior.

### DIFF
--- a/src/Core/Parser/TemplateParser.php
+++ b/src/Core/Parser/TemplateParser.php
@@ -698,7 +698,9 @@ class TemplateParser
         if (preg_match_all(Patterns::$SPLIT_PATTERN_SHORTHANDSYNTAX_ARRAY_PARTS, $arrayText, $matches, PREG_SET_ORDER)) {
             foreach ($matches as $singleMatch) {
                 $arrayKey = $this->unquoteString($singleMatch['Key']);
-                if (!empty($singleMatch['VariableIdentifier'])) {
+                if (array_key_exists('Subarray', $singleMatch) && !empty($singleMatch['Subarray'])) {
+                    $arrayToBuild[$arrayKey] = new ArrayNode($this->recursiveArrayHandler($singleMatch['Subarray']));
+                } elseif (!empty($singleMatch['VariableIdentifier'])) {
                     $arrayToBuild[$arrayKey] = new ObjectAccessorNode($singleMatch['VariableIdentifier']);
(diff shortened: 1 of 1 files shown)


smichaelsen:
> @dlubitz: This would solve the parser output in this case, but not the issue itself. But if it's the only way ATM to solve this, I would also go with that "hacky" solution.

The underlying problem can most likely not be fixed within fluid itself, because it seems to be somewhere in the PHP environment. As matter of fact there are those environments out there where the marching works differently, so I think Fluid should be able to cope with that case.

src/Core/Parser/TemplateParser.php
```diff
@@ -698,16 +698,16 @@ protected function recursiveArrayHandler($arrayText)
         if (preg_match_all(Patterns::$SPLIT_PATTERN_SHORTHANDSYNTAX_ARRAY_PARTS, $arrayText, $matches, PREG_SET_ORDER)) {
             foreach ($matches as $singleMatch) {
                 $arrayKey = $this->unquoteString($singleMatch['Key']);
-                if (!empty($singleMatch['VariableIdentifier'])) {
+                if (array_key_exists('Subarray', $singleMatch) && !empty($singleMatch['Subarray'])){
```
NamelessCoder: I'm not sure this changes anything - and if it does, it merely hides the fact that the array was not detected correctly by PCRE. `$singleMatch['Subarray']` will still be set but will be null or an empty array, which is the core symptom of the problem. So this condition will still not trigger - and if it triggers with anything other than an array, it leads to new PHP errors (constructor of ArrayNode does not accept non-arrays).

smichaelsen: > it merely hides the fact that the array was not detected correctly by PCRE.
Yes, that is the point, see my comment above.

In our tests `VariableIdentifier` or `QuotedString` were set *additionally* to `Subarray` and that's why the wrong if/elseif-branch was executed. Changing the order did fix that. I agree that the check for the Subarray is not solid enough, an `is_array()` should be added.

albe:
@NamelessCoder friendly ping regarding reopening and merging this as of your comment on #506 :)

NamelessCoder:
Merged since I can't provoke any errors associated with the change - but I still have this nagging sensation that this could break something. I'd appreciate if you could do an extremely careful round of testing on `dev-master` on both affected and unaffected systems, before I go ahead and backport + create releases.

NamelessCoder:
Version 2.6.9 has been released, if you are affected by this issue and prefer to use tagged release versions, you can now update `typo3fluid/fluid` to fix the issue.

PR #498, opened 2019-11-14
[TASK] Remove PHP 7.4 notices and deprecations

brotkrueml:
The access on array elements with curly brackets is deprecated in PHP 7.4:
https://wiki.php.net/rfc/deprecate_curly_braces_array_access

Also a notice is thrown when accessing an array value that is of type bool, null, etc:
https://github.com/php/php-src/pull/4386

### DIFF
--- a/src/View/AbstractTemplateView.php
+++ b/src/View/AbstractTemplateView.php
@@ -377,7 +377,7 @@ abstract class AbstractTemplateView extends AbstractView
     {
         $currentRendering = end($this->renderingStack);
         $renderingContext = $this->getCurrentRenderingContext();
-        $parsedTemplate = $currentRendering['parsedTemplate'] ? $currentRendering['parsedTemplate'] : $renderingContext->getTemplateCompiler()->getCurrentlyProcessingState();
+        $parsedTemplate = isset($currentRendering['parsedTemplate']) ? $currentRendering['parsedTemplate'] : $renderingContext->getTemplateCompiler()->getCurrentlyProcessingState();
         if ($parsedTemplate) {
             return $parsedTemplate;
         }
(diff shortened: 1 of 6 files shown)


src/View/AbstractTemplateView.php
```diff
@@ -377,7 +377,7 @@ protected function getCurrentParsedTemplate()
     {
         $currentRendering = end($this->renderingStack);
         $renderingContext = $this->getCurrentRenderingContext();
-        $parsedTemplate = $currentRendering['parsedTemplate'] ? $currentRendering['parsedTemplate'] : $renderingContext->getTemplateCompiler()->getCurrentlyProcessingState();
+        $parsedTemplate = isset($currentRendering['parsedTemplate']) ? $currentRendering['parsedTemplate'] : $renderingContext->getTemplateCompiler()->getCurrentlyProcessingState();
```
mbrodala: This should probably be `!empty()` instead to be closer to the previous behavior.

src/View/AbstractTemplateView.php
```diff
@@ -405,6 +405,6 @@ function($parent, TemplatePaths $paths) use ($controllerName, $actionName, $rend
     protected function getCurrentRenderingContext()
     {
         $currentRendering = end($this->renderingStack);
-        return $currentRendering['renderingContext'] ? $currentRendering['renderingContext'] : $this->baseRenderingContext;
+        return isset($currentRendering['renderingContext']) ? $currentRendering['renderingContext'] : $this->baseRenderingContext;
```
mbrodala: Same here.

NamelessCoder:
Thanks for the patch and updates, @brotkrueml - but I have to close this since it's no longer required. Another contributor created https://github.com/TYPO3/Fluid/pull/499 almost precisely at the same time as your PR. I chose the other even though you came first, and gave a few reasons in the other pull request.

PR #499, opened 2019-11-15
[FEATURE] Make fluid PHP 7.4 compatible

alexanderschnitzler:
Which is mainly not using curly braces to get characters
of a string and checking for valid arrays before accessing
them.

See https://wiki.php.net/rfc/notice-for-non-valid-array-container
for a further explanation.

To stay compatible with PHP 5 `isset` is used instead of a null
coalesce operator:

`isset($foo['bar']) ? $foo['bar'] : null;`

vs.

`$foo['bar'] ?? null;`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ cache:
 php:
   - 7.2
   - 7.3
+  - 7.4snapshot
 
 before_install:
   - composer self-update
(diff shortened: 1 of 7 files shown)


mbrodala:
Hm, this is basically the same as #498 ...

alexanderschnitzler:
True. I didn't know someone else is working on PHP 7.4 compatibility yet.

NamelessCoder:
Merging this instead of #498 which came first, chronologically, since:

* This PR appears more complete (at least one additional empty check applied)
* This PR adds 7.4 to testing matrix
* This PR is by a first-time contributor

NamelessCoder:
Re-labelled as bugfix since our current PHP requirement is (too) loose and already allows PHP 7.4, and the fact that these changes are so minor.

PR #504, opened 2020-01-06
[BUGFIX] Guard against filemtime on missing file

NamelessCoder:
Guard TemplatePaths->createIdentifierForFile against
attempts to use filemtime() on a file which does not
exist. Return a zero timestamp in this case.


NamelessCoder:
Thanks for the fast merge, @mbrodala :)

Issue #448, opened 2019-04-12
The ViewHelper "<f:fluid>" could not be resolved


vaxul:
The [documentation contains an example code](https://github.com/TYPO3/Fluid/blob/master/doc/FLUID_VIEWHELPERS.md#registeringimporting-viewhelpers) with a `<f:fluid>` tag, which couldn't work, since the viewhelper doesn't exists.

> PHP Fatal error:  Uncaught TYPO3Fluid\Fluid\Core\Parser\Exception: The ViewHelper "<f:fluid>" could not be resolved.
> Based on your spelling, the system would load the class "TYPO3Fluid\Fluid\ViewHelpers\FluidViewHelper", however this class does not exist. 

Seem I have stick to:
```html
<html data-namespace-typo3-fluid="true" xmlns:f="http://typo3.org/ns/TYPO3/Fluid/ViewHelpers">
</html>
```

mbrodala:
Korrekt, there is no `f:fluid` in Fluid. This must be an arbitrary element, just like `html` in your example. The `data-namespace-typo3-fluid` is only used to automatically remove this container from the output.

Can you make a PR to fix this? :-)

vaxul:
> Can you make a PR to fix this?
Sure, to fix the docs or create a f:fluid view helper? I think both is reasonable. 🤔

mbrodala:
No, please fix the docs. Adding a viewhelper without function would not be reasonable. ;-)

PR #505, opened 2020-01-08
[DOC] Remove wrong namespace registration examples

NamelessCoder:
Avoids use of pseudo-ViewHelper "f:fluid" in favor
of traditional way using "html" tag.

Close: #448


NamelessCoder:
@mbrodala Would you mind giving this a quick look? Just read-check required since it only affects docs.

mbrodala:
Never used `<f:fluid>` anywhere so this is a nobrainer.

PR #509, opened 2020-01-29
fixes #353

### LINKED ISSUE #353: [BUG] Multiple level references not resolved
Multiple same level references are not resolved.

Example:

```
  <f:variable name="array" value="{0: {0: 'foo-0.0', 1: 'foo-0.1'}, 1: {0: 'bar-0.1', 1: 'bar-0.1'}}" />
  <f:variable name="index1" value="0" />
  <f:variable name="index2" value="1" />

  Example 1 - Does not work: {array.{index1}.{index2}}

  <f:variable name="subitems" value="{array.{index1}}"/>

  Example 2 - Works: {subitems.{index2}}
```

Example 1 should also return the given index of the multidimensional array.

smoebody:
* modifies regex for subvariable matching
* adds test

### DIFF
--- a/src/Core/Variables/VariableExtractor.php
+++ b/src/Core/Variables/VariableExtractor.php
@@ -120,7 +120,7 @@ class VariableExtractor
     protected function resolveSubVariableReferences($subject, $propertyPath)
     {
         if (strpos($propertyPath, '{') !== false) {
-            preg_match_all('/(\{.*\})/', $propertyPath, $matches);
+            preg_match_all('/(\{[^}]+\})/', $propertyPath, $matches);
             foreach ($matches[1] as $match) {
                 $subPropertyPath = substr($match, 1, -1);
                 $propertyPath = str_replace($match, $this->getByPath($subject, $subPropertyPath), $propertyPath);
--- a/tests/Unit/Core/Variables/VariableExtractorTest.php
+++ b/tests/Unit/Core/Variables/VariableExtractorTest.php
@@ -50,6 +50,7 @@ class VariableExtractorTest extends UnitTestCase
             [['user' => $namedUser], 'user.invalid', null],
             [['foodynamicbar' => 'test', 'dyn' => 'dynamic'], 'foo{dyn}bar', 'test'],
             [['foo' => ['dynamic' => ['bar' => 'test']], 'dyn' => 'dynamic'], 'foo.{dyn}.bar', 'test'],
+            [['foo' => ['dynamic1' => ['dynamic2' => ['bar' => 'test']]], 'dyn1' => 'dynamic1', 'dyn2' => 'dynamic2'], 'foo.{dyn1}.{dyn2}.bar', 'test'],
             [['user' => $namedUser], 'user.hasAccessor', true],
             [['user' => $namedUser], 'user.isAccessor', true],
             [['user' => $unnamedUser], 'user.hasAccessor', false],


NamelessCoder:
Hi @smoebody,

Please see https://github.com/TYPO3/Fluid/pull/472 for an explanation why this patch cannot be accepted. Same-level references are unlikely to ever be supported on Fluid 2.x but Fluid 3.x will support them no problem.

Same as I wrote in the other rejected pull request:

> Don't let this discourage you from future pull requests!

;)

smoebody:
Hi @NamelessCoder,

I dont understand, why you reject this pull-request. Having this fixed in v3 is not a valid reason IMHO. There is definitely a need for this right now and since fluid is used as template-engine in Typo3 and even Typo3v9 requires "typo3fluid/fluid": "^2.6.8", which means it is unlikely that even the newest Typo3 will see support for multiple subvariables.

Your arguments are
> Inability to extract multiple same-level variables is a known limitation of Fluid 2.x and earlier
Knowing a limitation might help suggesting workarounds but should not prevent removing this limitations in the first place. Especially when they are fixed that easy.
> Changing this also implies a need to change a lot more regarding how variables are resolved
I could not find anything that would not work. All tests pass, i even added a new test for this use-case. If there are more things to take care of, please point me to, so i can have a look at it.
> It is a non-issue in the next major version of Fluid, 3.0
This does not help when - as in my case - fluid is a third-party-dependency of a framework, such as typo3. Since fluid3 is a major release it is likely that there is no backward-compatibility to fluid2 and therefore will not make it into the cms.

Also this change breaks nothing. It is therefore backward-compatible and according to semver this qualifies for a [minor upgrade](https://semver.org/#spec-item-7):
> Minor version Y (x.Y.z | x > 0) MUST be incremented if new, backwards compatible functionality is introduced to the public API. It MUST be incremented if any public API functionality is marked as deprecated. It MAY be incremented if substantial new functionality or improvements are introduced within the private code. It MAY include patch level changes. Patch version MUST be reset to 0 when minor version is incremented.

Please would you reconsider your decision, I am certainly not the only one who needs this.

Regards
Ulf

NamelessCoder:
Hi Ulf,

This is - unfortunately - very far from being as simple a fix as it appears.

It's not exactly easy to explain, mainly because there's a significant mess in the interplay between `VariableExtractor` (which is deprecated) and `StandardVariableProvider` (which has copied the logic from `VariableExtractor` as part of the deprecation). The reason this was deprecated is a while back in history:

https://github.com/TYPO3/Fluid/pull/308

This happened because we received a bug report from the Neos team, because they use a special variable extraction logic, implemented a custom VariableProvider which did this, but were unable to override the `VariableExtractor` which meant that *their variable extraction only worked in uncompilable templates*. The original error description can be found here:

https://github.com/TYPO3/Fluid/issues/255

The deprecation solution involved short-circuiting the `VariableExtractor` so it calls `StandardVariableProvider->getByPath()` if the subject is an instance of this class, thus triggering instead the logic moved from VariableExtractor. Important to note is that this is why the `VariableExtractorTest` actually works with this change - the subject we feed to the `VariableExtractor` in this case is *a pure array*.

That's the historical background for why `VariableExtractor` is deprecated and the reason why `VariableExtractorTest` doesn't report any issues if you change this regular expression there.

Knowing this, we can infer that the right solution is to change the expression not in the `VariableExtractor` but rather in the `StandardVariableProvider`. And this is where things begin to break: if you implement it there, an additional test is triggered via `ObjectAccessorNodeTest` (again, sorry for the mess!):

```php
[
['foo' => ['bar' => 'test'], 'dynamic' => ['sub' => 'bar'], 'baz' => 'sub'], // data set
'foo.{dynamic.{baz}}', // expression
'test' // asserted expectation
],
```

And this is the test which fails if you change the regular expression as your patch suggests.

In short:

* The tests don't fail with your patch because the tested subject is a pure array which means the short-circuiting to `StandardVariableProvider` does not happen.
* And the `VariableExtractorTest` does not contain the test set quoted above.
* Therefore, fixing A breaks B.

You can confirm this by adding the test case above to the `VariableExtractorTest` which results in an error.

The next part of the reasoning is that `StandardVariableProvider` (or a subclass hereof) is always the root of any variable expression in Fluid. This means that the fix in this patch *actually doesn't fix the problem for real environments, it only fixes it for the test case because of the two reasons above*.

Even further detail: the reason why the `.*` expression works for recursively nested variables is that is is "greedy but docile" - this is why changing it to `[^}]+` won't work (it turns the expression into a "pure greedy" one). And if the expression is docile, which it has to be if we are to avoid breaking changes, it is unable to match multiple same-level dynamic variable parts.

Pheeeeew. As you can tell by now, it's definitely not an easy fix.

So the final situation is that if we change this in `VariableExtractor` it has no impact on real-life installs anyway. If we change it in `StandardVariableProvider` then we break something else, and if we break something else, the next patch level for Fluid would have to be a major revision - and this problem already is dead and gone in Fluid 3.0. In the end, this means the patch becomes redundant.

I honestly am sorry to reject this, but even explaining it like I did above, demanded more than an hour of testing and writing. That's how not-simple this really is. I personally don't think there is any sensible solution to this situation (greedy-docile vs. greedy will continue to come back and bite us, is my prediction). I may be wrong about this - but given the above explanation it should now be clear why the patch is rejected. You are obviously welcome to try and pursue a different solution but I should warn you that it probably will require a significant amount of time and definitely isn't an easy fix; and is likely to constitute a breaking change which would mean I once again would reject it because Fluid 3.0 simply isn't affected by this problem of docile-greedy vs. greedy - this new major version completely drops the use of PCRE and instead linearly parses such expressions in a way that supports both of these structures.

NamelessCoder:
I've added a PR that would add the test that fails, in order to make it clearer and hopefully help you out if/when you decide to further challenge yourself to try and add support for multiple same-level sub-variable references in Fluid 2.x.

https://github.com/TYPO3/Fluid/pull/510

smoebody:
Hi @NamelessCoder,

thanks for taking the time to explain this in more detail. I now see your point, also having the new test shows the breaking change.

I will see if i can find something to overcome the limitation, but from what you explained i can see that this is bigger than i thought. Maybe a fresh perspective helps but more likely it won't.

Sorry to have you bothered with my persistence.

PR #510, opened 2020-01-30
[TASK] Add two missing test cases

NamelessCoder:
This patch adds two missing test cases for VariableExtractor
and StandardVariableProvider. The absence of these tests
has caused two pull requests to be rejected even though the
patches' tests ran successfully.

To prevent future pitfalls, the test case that would fail if a
developer attempts to make multiple same-level sub-variable
references work, are added in the appropriate places.

References: #509, #472


NamelessCoder:
@mbrodala Sorry to keep pinging you like this... but would you mind reading this one and merging if you're fine with it? Should be a formality since it only touches test cases and adds assertions for current behavior.

mbrodala:
The test succeeds and also looks correct from reading. :+1:

PR #519, opened 2020-05-09
[TASK] Add Travis builds on osx and windows

NamelessCoder:
Adds the "osx" and "windows" environments to the Travis
build script to trigger building on these in addition to "linux".

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,11 @@ dist: trusty
 
 language: php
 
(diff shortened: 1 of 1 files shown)


NamelessCoder:
Hmm, unfortunately it seems this isn't possible to achieve with PHP projects on Travis - perhaps at some point in the future it will be possible!

PR #523, opened 2020-09-06
[TASK] Enable testing on PHP 7.4

NamelessCoder:
Enable 7.4 in matrix and convert incompatible syntax occurrences.

PR #524, opened 2020-09-06
[BUGFIX] Fix sub-variable in string arguments detected as arrays

NamelessCoder:
Fixes an edge case where a sub-variable reference in
a string tag argument would be detected as an array
due to an incorrect condition which switched the
sequencer to array sequencing mode if an inline
syntax open character was detected and arrays are
allowed.

Example:

<f:c foo="{var.{sub}}" />

Before:

foo=['var', ObjectAccessor('sub')]

After:

foo=ObjectAccessor('var.', ObjectAccessor('sub'))


NamelessCoder:
Requires https://github.com/TYPO3/Fluid/pull/523 which fixes the failing tests, can be rebased after that patch and tests will be green.

NamelessCoder:
Bug can be confirmed by TYPO3 CMS 11-dev when editing any record that contains a media reference which allows crop/scale. A call to `f:translate` will cause an error where `id` is an array instead of a resolved variable with sub-variable reference.

PR #525, opened 2020-10-09
[TASK] Replace method_exists by is_callable

kabarakh:
In the method detectAccessor of the class VariableExtractor is_callable is used, but in the method detectAccessor of the class StandardVariableProvider method_exits ist used. So when trying to use a magic call to access a dynamic generated method, the call fails. By using is_callable, the __call method is recognized and used


mbrodala:
Thanks for the contribution, unfortunately the topic is highly complex and cannot be fixed that easily. See #467 and all related changes/reverts for the previous attempt.

PR #526, opened 2020-10-24
[TASK] Move argument validation/conversion from ViewHelperNode to TemplateParser

NamelessCoder:
This change moves the validation of arguments and conversion
of boolean arguments to boolean nodes, away from the
ViewHelperNode class and into the TemplateParser class.

This has a couple of benefits:

1. It means the Parser is now aware of the argument definitions
   and their types while parsing arguments, which saves a sub-
   loop and therefore increases performance.
2. It makes the ViewHelperNode constructor less heavy.
3. It brings the behavior of the template parsing logic in Fluid 2.x
   closer to the behavior in Fluid 3.x.


mbrodala (approved):
Looks good just from reading. I see some points which could be de-duplicated but that'd be too much for a single change.

Issue #512, opened 2020-02-18
Formatting issues in ViewHelper reference - pragmatic suggestion


sypets:
The topic of Fluid VH reference reST vs. markdown etc. has been discussed at length. We still have the automatically generated typo3fluid VH which looks pretty ugly, because the rendering toolchain expects reStructuredText at some point and the source was written as Markdown. Last status was, we cannot easily fix this at this point (unless someone is willing to work on the scripts. I don't know how difficult or complicated this might be). 

Now I came across this [if-VH page](https://docs.typo3.org/other/typo3/view-helper-reference/9.5/en-us/typo3fluid/fluid/latest/If.html) which looks partly quite ok, when rendered on docs.typo3.org.

So, can we use a reasonable subset of markup that looks ok - whether it is interpreted as reST or Markdown? Incidentally, I found this resource [Markdown and reStructuredText](https://gist.github.com/dupuy/1855764) a while ago.

-----

For **headers**, we an use underlining which is ok in both worlds, see https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#headers

```
header
=====
```

Also, instead of headers we can use bold (`**bold**`) which is also ok in both worlds

There is no really good solution for **links** (but we usually don't use links in the VHs)

For **lists** we can use * list item 1 etc. which is ok (reST is just really picky, you MUST have an empty line before and after). 

```
* list item 1
* list item 2
```

For code-blocks there is a nifty solution in the mentioned resource:

> you can combine the formatting for both of them by ending one paragraph with a double colon :: (for reStructuredText) and indenting the next one by four or more spaces (for Markdown) to make it appear in monospaced font without flow or word-wrapping:

example:

```
some example code::

    <f:if condition="{whatever} == 'notever'">
```

sypets:
This is the mentioned if-ViewHelper page which is already rendered partly ok: 

* on docs.typo3.org: https://docs.typo3.org/other/typo3/view-helper-reference/9.5/en-us/typo3fluid/fluid/latest/If.html
* the source: https://github.com/TYPO3/Fluid/blob/master/src/ViewHelpers/IfViewHelper.php

![example3](https://user-images.githubusercontent.com/13206455/74746376-a3b5da00-5265-11ea-8a7f-6c5cead3d93a.png)

mbrodala:
Interesting approach to improve the current state. :-)

Explicitly customizing the code block highlighting (code language) won't be possible with this, right?

Markdown uses this:

```
[...]

PR #516, opened 2020-03-13
[TASK] Migrate phpdoc VH markup to common syntax

sypets:
This is based on the PR #436 by @DanielSiepmann and the suggestions in issue #512. 

A summary of the used markup is in my gist: https://gist.github.com/sypets/360a62ce0685521049a65be366edbbef

Following changes were performed:

- removed admonitions (note, caution)
- removed textroles, e.g. :xml:. Use just double-backticks for inline code.
- removed :ref:
- If the text was in a note, I moved it to the top (if it is important enough to warrant a note, I think it should exist before the examples)

PR #527, opened 2020-10-27
[TASK] Migrate phpdoc VH markup to common syntax II

sypets:
I cleaned up my workspace and accidentally removed my fork. Did not find another way to restore branch in original PR (see https://github.com/isaacs/github/issues/168). 

This is the same commit as in #516

-------------

This is based on the PR #436 by @DanielSiepmann and the suggestions in issue #512. 

A summary of the used markup is in my gist: https://gist.github.com/sypets/360a62ce0685521049a65be366edbbef

Following changes were performed:

- removed admonitions (note, caution)
- removed textroles, e.g. :xml:. Use just double-backticks for inline code.
- removed :ref:
- If the text was in a note, I moved it to the top (if it is important enough to warrant a note, I think it should exist before the examples)

Resolves: #512


mbrodala (approved):

NamelessCoder:
I wish you would have avoided merging this PR before discussing it with me. The intention was to not merge this until a new major version was pending for release, because it changes the way documentation generators have to handle the markup - and now these generators will have to compare versions against a bugfix release. In the meantime I've almost decided to compromise on this and allow the markup change in a minor version - but it was never intended to be merged to become part of a bugfix version.

Basically, my blocking comment https://github.com/TYPO3/Fluid/pull/436#issuecomment-544430942 was never resolved.

It's not a huge thing (just a minor annoyance for doc generators) but in the future please make sure to discuss with me before merging pull requests that either have been sitting around for a long time, or are based on older pull requests that have been sitting around for a long time. There's usually a reason why they haven't been merged.

mbrodala:
I'm sorry, I wasn't aware of this background since no comment was added here. It wouldn't be a huge deal to revert this change and push it as new PR again. Would this be an option?

NamelessCoder:
Let's do a micro-survey before we decide to revert or not. I know this would cause trouble for the VH doc package that I maintain (https://github.com/NamelessCoder/fluid-documentation-generator) but if I'm the only one I'd say we just keep it - I can then implement a format detection or version number condition to support the new format.

Questions for @maddy2101 and @kitsunet about their respective doc generating implementations:

* If we keep this patch will we cause trouble for your documentation generators?
* If there is trouble, could you accept having to add detection or a version number check to determine which markup format to handle?
* Is it a major task for either of you to implement support for the changed markup format (one that you'd rather postpone)?

sypets:
Sorry, about the mixup. That was not intentional - the reason I created a new PR was because I removed my fork a while ago (which deleted the relation to the branch). 

My PR was an attempt to make markup work with docs rendering on docs.typo3.org - but if this is not possible currently - I think it might be good idea to revisit (and agree on) the entire concept before putting more effort into this. 

Current problems:

- Writing code in a formatting language (.rst or markdown does not really matter here) inside source code comments is tricky - no proper syntax highlighting or error checking by IDEs etc. - the result are various formatting problems
- no unified agreed on format between TYPO3 core and Fluid
- VH rendering workflow not documented (AFAIK)
- no "Edit me on GitHub" capability - contributors often create PR in the "automatically generated" repository

If the documentation is not inside the comments but as separate documentation files - this might solve part of these problems.

PR #528, opened 2020-10-31
[BUGFIX] Restore support for data- prefixed arguments

NamelessCoder:
Due to a regression in https://github.com/TYPO3/Fluid/pull/419,
a TagBasedViewHelper which was called with one or more
data- prefixed arguments would discard the prefixed arguments
and the output would not contain the attribute.

This patch fixes that regression by moving the handling of
unregistered arguments to the initialize() function in order
to make sure the additional attributes are handled *after*
the reset() method has been called on TagBuilder.

References: #419

### DIFF
--- a/tests/Functional/Cases/TagBasedTest.php
+++ b/tests/Functional/Cases/TagBasedTest.php
@@ -1,0 +1,48 @@
+<?php
+namespace TYPO3Fluid\Fluid\Tests\Functional\Cases;
+
+use TYPO3Fluid\Fluid\Core\ViewHelper\ViewHelperInterface;
+use TYPO3Fluid\Fluid\Core\ViewHelper\ViewHelperInvoker;
+use TYPO3Fluid\Fluid\Tests\Functional\Fixtures\ViewHelpers\TagBasedTestViewHelper;
+use TYPO3Fluid\Fluid\Tests\Unit\Core\Rendering\RenderingContextFixture;
+use TYPO3Fluid\Fluid\Tests\UnitTestCase;
+
+class TagBasedTest extends UnitTestCase
+{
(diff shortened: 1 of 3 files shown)


mbrodala:
@NamelessCoder Can you rebase this one? I guess #530 will need to be adjusted if merged after this one.

NamelessCoder:
@mbrodala Done, but please wait for Travis to complete - I wasn't able to run unit tests locally and I'm a bit busy on other projects. Yep, #530 will need an adjustment afterwards!

tests/Functional/Cases/TagBasedTest.php
```diff
@@ -0,0 +1,48 @@
...
+    {
+        $invoker = new ViewHelperInvoker();
+        $viewHelper = new TagBasedTestViewHelper();
+        $arguments = [
+            'data-foo' => 'bar',
+        ];
+        $result = $invoker->invoke($viewHelper, $arguments, new RenderingContextFixture());
+        $this->assertSame('<div data-foo="bar" />', $result);
+    }
+}
```
mbrodala, line 48: We should add one additional test here which passes in *both* `'data-foo'` and `'data' => ['foo' => ...]` to clearly document the expected result for such a conflict.

mbrodala (changes requested):
Should I add this test?

NamelessCoder:
> Should I add this test?

That would be great, thanks!

mbrodala (approved):

mbrodala:
@NamelessCoder I'd merge this now if you don't see anything missing since it predates #530

NamelessCoder:
@mbrodala All good from my perspective!

Issue #534, opened 2021-04-21
[FEATURE] Add "aria" argument to AbstractTagBasedViewHelper


o-ba:
Since accessibility is a very important topic in the web, the Fluid framework should support its developers as much as possible in building user-friendly and accessible applications. A first step is to describing elements with `aria` (Accessible Rich Internet Applications) attributes. Further reading: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA.

To ease the use of these attributes, the same functionality as for the `data` attributes should also be implemented for the `aria` attributes. Especially since there are some limitations for using `additionalAttributes`. See for example: https://forge.typo3.org/issues/93963

mbrodala:
There's already a PR pending in #530 for this ...

o-ba:
Yes, sorry. I just searched for an existing issues, not for existing PRs.

How would you like to proceed, since #535 does not only allow to pass the `aria-` attribute, but also allows to define them with a new VH argument and supports them in `TagBuilder->addAttribute()`. Having a dedicated argument is required, see the TYPO3 forge [issue](https://forge.typo3.org/issues/93963). I furthermore already applied the same test coverage as for the `data-` attribute functionality.

mbrodala:
As discussed I'd vote for including #530 first since it handles a part of the topic. Then #535 can be added on top focussing on the `aria` argument alone.

o-ba:
Yes, fine for me.

PR #536, opened 2021-04-28
[FEATURE] Introduce aria argument in AbstractTagBasedViewHelper

o-ba:
To support developers in building accessible applications,
the AbstractTagBasedViewHelper now provides a new
argument aria. It therefore can be used to add additional
aria attributes to the HTML element.

### DIFF
--- a/src/Core/ViewHelper/TagBuilder.php
+++ b/src/Core/ViewHelper/TagBuilder.php
@@ -194,9 +194,11 @@ class TagBuilder
         if ($escapeSpecialCharacters) {
             $attributeName = htmlspecialchars($attributeName);
         }
-        if ($attributeName === 'data' && (is_array($attributeValue) || $attributeValue instanceof \Traversable)) {
+        if (($attributeName === 'data' || $attributeName === 'aria')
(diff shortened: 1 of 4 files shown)


o-ba:
Closes #534

src/Core/ViewHelper/TagBuilder.php
```diff
@@ -194,9 +194,11 @@ public function addAttribute($attributeName, $attributeValue, $escapeSpecialChar
         if ($escapeSpecialCharacters) {
             $attributeName = htmlspecialchars($attributeName);
         }
-        if ($attributeName === 'data' && (is_array($attributeValue) || $attributeValue instanceof \Traversable)) {
+        if (($attributeName === 'data' || $attributeName === 'aria')
```
mbrodala, line 197: We can use `in_array()` here to shorten this.

mbrodala (changes requested):

src/Core/ViewHelper/TagBuilder.php
o-ba: Done

mbrodala (approved):

PR #537, opened 2021-05-04
[BUGFIX] Fix eager rendering of RenderViewHelper children

maechler:
The `RenderViewHelper` currently eagerly renders its children, even if they are not used. This can lead to unexpected behaviour using Layout-Sections with default content. Consider the following setup:

**Layout**
```html
<f:render section="Main">
  <f:debug>Debug from Layout</f:debug>
  Output from Layout
</f:render>
```

**Template**
```html
<f:layout name="Default"/>
<f:section name="Main">
  <f:debug>Debug from Template</f:debug>
  Output from Template
</f:section>
```

**Output (before this PR)**
```html
Debug from Layout
Debug from Template
Output from Template
```

**Expected Output (after this PR)**
```html
Debug from Template
Output from Template
```

Even though Fluid only outputs the content from the Template, the Layout's section gets rendered as well shown with the DebugViewHelper. In my opinion this should not happen and even if it were not for this layout bug, I believe the children should not be rendered eagerly for performance reasons.


NamelessCoder:
Change itself looks fine to me, but we can't yet force use of null coalesce since technically, PHP >= 5.5 is still supported. A weak-type ternary condition, or one that uses isset() would be equally acceptable as alternative.

maechler:
@NamelessCoder Thanks for your fast response! I updated the PR and made it even more lazy in the case `$arguments['default']` is set. If you think it was more readable before, just let me know.

NamelessCoder:
Change still looks good - although I'd like to see some added test cases in https://github.com/TYPO3/Fluid/blob/master/tests/Unit/ViewHelpers/RenderViewHelperTest.php which make assertions that render-children won't happen in the specific cases that don't need it.

maechler:
@NamelessCoder I updated the unit test to make sure that `renderChildren` is at most called once.

maechler:
@NamelessCoder Is this PR ready to be merged?

lolli42:
Sorry for the long delay and many thanks for your work and patience.

I rebased the patch and adapted cgl details we established meanwhile.

Codewise, this should be ready.

I'll finally go ahead with a merge now.

PR #539, opened 2021-05-07
[BUGFIX] Reset arguments when using cached templates

### LINKED ISSUE #538: Fluid 3: partial arguments are persistent for all following calls
If i use a partial and define an argument, this argument is also set for all following partial calls.
```
    <f:render partial="MyPartial" /> <!-- {classes} is empty in partial -->
    <f:render partial="MyPartial" arguments="{classes: 'foo'}" />  <!-- {classes} is 'foo' in partial -->
    <f:render partial="MyPartial" arguments="{bar: 1}" /> <!-- {classes} is 'foo' and {bar} is 1 in partial -->
    <f:render partial="MyPartial" /> <!-- {classes} is 'foo' and {bar} is 1 in partial -->
```

Problem does not exist when using atoms instead of `f:render`.

sascha-egerer:
The TemplateParser creates a cache for parsed templates.
The arguemnts must be reset when using a cached verion
of a template.

Resolves #538

### DIFF
--- a/src/Component/AbstractComponent.php
+++ b/src/Component/AbstractComponent.php
@@ -293,4 +293,17 @@ abstract class AbstractComponent implements ComponentInterface
         }
         return (string) $value;
     }
+
+    public function __clone()
+    {
+        if ($this->arguments !== null) {
+            $this->arguments = clone $this->arguments;
+        }
+
+        if (0 < count($this->children)) {
+            $this->children = array_map(function ($child) {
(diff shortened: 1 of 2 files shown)


sascha-egerer:
Not sure if this is sufficient. There could also be a problem with children of the template which are also stored in the object.

sascha-egerer:
Component instances and all child instances are now cloned when coming from cache.

PR #541, opened 2021-05-07
[BUGFIX] Evaluate components before using them as atom argument

### LINKED ISSUE #540: Fluid 3: Parameter of Atom does not support variables
If i pass a variable as argument to an Atom the value is always null.

```
    <my:atom name="bar"></my:atom>

    <f:variable name="foo">bar</f:variable>
    <my:atom name="{foo}"></my:atom>
```

The first one works (`{name}` is `bar` in `atom`) but in the second example `{name}` is always null

sascha-egerer:
If a variable of type object accessor is used as an argument value
for an atom, this variable must be evaluated before it is passed to
the atom so it is evaluated with the correct rendering context.

Resolves #540

### DIFF
--- a/src/Core/Parser/SyntaxTree/AtomNode.php
+++ b/src/Core/Parser/SyntaxTree/AtomNode.php
@@ -21,7 +21,7 @@ class AtomNode extends AbstractComponent
     protected $escapeOutput = false;
 
     protected $file;
-    
+
     public function setName(?string $name): ComponentInterface
     {
         $this->name = $name;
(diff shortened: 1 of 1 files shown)


src/Core/Parser/SyntaxTree/AtomNode.php
```diff
@@ -38,7 +38,14 @@ public function evaluate(RenderingContextInterface $renderingContext)
     {
         $atom = clone $renderingContext->getTemplateParser()->parseFile($this->file);
         $arguments = clone $atom->getArguments();
-        $arguments->assignAll((array) $this->getArguments()->getAllRaw() + $renderingContext->getVariableProvider()->getAll())->setRenderingContext($renderingContext);
+        $evaluatedArguments = [];
```
NamelessCoder, line 41: I think you can get the same result by:

```php
$arguments->assignAll($this->getArguments()->getArrayCopy() + $renderingContext->getVariableProvider()->getAll())->setRenderingContext($renderingContext);
```

The `getArrayCopy` method will evaluate all arguments against the assigned RenderingContext before returning a plain old array; this is how it differes from `getAllRaw()`.

NamelessCoder (changes requested):

src/Core/Parser/SyntaxTree/AtomNode.php
sascha-egerer: Ok, I'll test that. 
I had also an issue where an argument containing `{foo.{bar}.baz}` did not return the value but an array with parts of that string as values.

sascha-egerer:
I think this issue has been resolved by #552 in a better way

sascha-egerer:
Oh, no, i was wrong here. Still an issue-

PR #544, opened 2021-05-18
Add Argument viewhelper

### LINKED ISSUE #427: Feature: slots

sascha-egerer:
Add a new `f:argument` viewhelper to be able to add
arguments to a node by using `f:argument` as child node.

Usage example:
```
<my:node>
	<f:argument name="content">Hello World</f:argument>
</my:node>
```

The sequencer does now put validation errors into a stack
so we are able to revalidate a node before it is closed but
children have been parsed. So we can still show the initial
problem if e.g. a required property was not defined.

Resolves #427

### DIFF
--- a/src/Core/Parser/Sequencer.php
+++ b/src/Core/Parser/Sequencer.php
@@ -122,6 +122,11 @@ class Sequencer
      */
     protected $nodeStack = [];
 
+    /**
+     * @var array<string, array<SequencingException>>
+     */
+    protected $nodeSequencingExceptionStack = [];
+
     public function __construct(
         RenderingContextInterface $renderingContext,
         Contexts $contexts,
@@ -191,13 +196,14 @@ class Sequencer
                     try {
                         $childNode = $this->sequenceTagNode();
                     } catch (Exception $exception) {
+                        if (!$exception instanceof SequencingException) {
+                            $exception = $this->createErrorAtPosition(
+                                $exception->getMessage(),
+                                $exception->getCode()
+                            );
+                        }
                         $childNode = new TextNode(
-                            $this->renderingContext->getErrorHandler()->handleParserError(
-                                $this->createErrorAtPosition(
-                                    $exception->getMessage(),
-                                    $exception->getCode()
-                                )
-                            )
+                            $this->renderingContext->getErrorHandler()->handleParserError($exception)
                         );
                     }
                     $this->splitter->switch($this->contexts->root);
@@ -343,6 +349,7 @@ class Sequencer
         $closing = false;
         $escapingEnabledBackup = $this->escapingEnabled;
         $viewHelperNode = null;
+        $argumentValidationError = null;
 
         $this->splitter->switch($this->contexts->tag);
         $this->sequence->next();
@@ -438,7 +445,15 @@ class Sequencer
 
                     if (!$closing || $selfClosing) {
                         $viewHelperNode = $viewHelperNode ?? $this->resolver->createViewHelperInstance($namespace, (string) $method);
-                        $viewHelperNode->onOpen($this->renderingContext)->getArguments()->validate();
+                        try {
+                            $viewHelperNode->onOpen($this->renderingContext)->getArguments()->validate();
+                        } catch (Exception $e) {
+                            $argumentValidationError = $this->createErrorAtPosition(
+                                $e->getMessage(),
+                                1621337984
+                            );
+                            $this->addSequencingExceptionToStack($viewHelperNode, $argumentValidationError);
+                        }
                     } else {
                         // $closing will be true and $selfClosing false; add to stack, continue with children.
                         $viewHelperNode = array_pop($this->nodeStack);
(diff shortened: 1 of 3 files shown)


mbrodala:
Can you add a few tests for this?

src/ViewHelpers/ArgumentViewHelper.php
```diff
@@ -0,0 +1,58 @@
...
+use TYPO3Fluid\Fluid\Component\EmbeddedComponentInterface;
+use TYPO3Fluid\Fluid\Component\SequencingComponentInterface;
+use TYPO3Fluid\Fluid\Core\Parser\Sequencer;
+use TYPO3Fluid\Fluid\Core\Rendering\RenderingContextInterface;
+use TYPO3Fluid\Fluid\Core\ViewHelper\AbstractViewHelper;
+
+/**
+ * Argument assigning ViewHelper
+ *
+ * Assigns an argument for a parent ViewHelper call when
```
NamelessCoder, line 19: Drop "when the parent ViewHelper supports it." because every ViewHelper will support this method.

I would also reword to "Assigns an argument to the parent node" since this works for every node that supports arguments (Atom, ViewHelper).

src/ViewHelpers/ArgumentViewHelper.php
```diff
@@ -0,0 +1,58 @@
...
+{
+    protected $escapeOutput = false;
+
+    /**
+     * @return void
+     */
+    public function initializeArguments()
+    {
+        parent::initializeArguments();
+        $this->registerArgument('name', 'string', 'Name of the parameter', true);
```
NamelessCoder, line 50: Argument "value" also needs to be registered (as optional) with description saying that "If not provided, argument value is taken from tag content or inline-passed child").

src/Core/Parser/Sequencer.php
```diff
@@ -438,7 +445,15 @@ protected function sequenceTagNode(): ?ComponentInterface
 
                     if (!$closing || $selfClosing) {
                         $viewHelperNode = $viewHelperNode ?? $this->resolver->createViewHelperInstance($namespace, (string) $method);
-                        $viewHelperNode->onOpen($this->renderingContext)->getArguments()->validate();
+                        try {
+                            $viewHelperNode->onOpen($this->renderingContext)->getArguments()->validate();
+                        } catch (Exception $e) {
+                            $argumentValidationError = $this->createErrorAtPosition(
+                                $e->getMessage(),
+                                1621337984
```
NamelessCoder, line 453: Better resolve the exception code from `$e->getCode()` instead.

src/Core/Parser/Sequencer.php
```diff
@@ -486,6 +501,36 @@ protected function sequenceTagNode(): ?ComponentInterface
...
+                    $nodeHash = spl_object_hash($viewHelperNode);
+                    // Check if we already have a validation error on the sequencing error stack
+                    foreach ($this->nodeSequencingExceptionStack[$nodeHash] ?? [] as $sequencingException) {
+                        if ($sequencingException->getCode() === 1621337984) {
+                            $argumentValidationError = $sequencingException;
+                            break;
+                        }
+                    }
+
+                    try {
```
NamelessCoder, line 514: This try/catch, including the throw, should only happen if `$argumentValidationError !== null` since we don't want to re-validate arguments on every valid node - only those that previously added an error to the stack.

Then you can also skip the condition below and directly throw the error:

```php
throw ($argumentValidationError ?? $this->createErrorAtPosition(...));
```

...and better resolve the exception code from `$e->getCode()` instead of hardcoding it here.

src/Core/Parser/Sequencer.php
```diff
@@ -191,13 +196,14 @@ public function sequence(): ComponentInterface
                     try {
                         $childNode = $this->sequenceTagNode();
                     } catch (Exception $exception) {
+                        if (!$exception instanceof SequencingException) {
```
NamelessCoder, line 199: This changed block seems redundant - or am I overlooking something? If redundant, please revert this change block.

src/Core/Parser/Sequencer.php
```diff
@@ -122,6 +122,11 @@ class Sequencer
      */
     protected $nodeStack = [];
 
+    /**
+     * @var array<string, array<SequencingException>>
```
NamelessCoder, line 126: `@var SequencingException[]` is preferred.

NamelessCoder (changes requested):

src/Core/Parser/Sequencer.php
ohader: `SequencingException[]` means `list<SequencingException>`, means `array<int, SequencingException>` - which does not reflect using `$nodeHash` as string.

NamelessCoder:
@ohader It doesn't actually change the argument *definitions*, it merely defines an argument to pass to the atom that is rendered. These two things are equivalent:

```xml
<my:node arg="Value" />
<!-- or -->
<my:node>
    <f:argument name="arg" value="Value" />
</my:node>
```

The intention is to provide a way to pass larger argument values, such as a big block of text that would be horrible to put into a tag argument. You could also compare it to this:

```xml
<f:variable name="text">
Big, big text with "quotes" and other stuff that doesn't play well as tag attribute value.
</f:variable>

<my:node arg="{text}" />
```

Basically just a quality-of-life type of feature.

Side note; for Atoms we distinguish between a _parameter_ (which is an ArgumentDefinition) and an _argument_ (which is the value assigned to a parameter and passed to the Atom).

ohader:
> Side note; for Atoms we distinguish between a _parameter_ (which is an ArgumentDefinition) and an _argument_ (which is the value assigned to a parameter and passed to the Atom).

I hope you do see the problem here and the confusion it creates... 😉

NamelessCoder:
> I hope you do see the problem here and the confusion it creates... 😉

Mnjaaah, yes and no. If we didn't make this distinction then you couldn't tell an argument value from an argument definition (which then caused the confusion leading to your question today). If we do make the distinction then of course people have to be aware that there is actually a distinction. _Damned if we do and damned if we don't_ :)

BTW I'd personally recommend to make this distinction also when speaking about PHP itself, we tend to just use the name "argument" there for both the definition and the runtime value. It's just easier to talk about if there's a distinction.

mbrodala:
@NamelessCoder I think what @ohader was hinting at is that "argument definition" / "argument" would maybe be clearer than "parameter" / "argument". At least that's how I see it and it would allow for using "parameter definition" / "parameter" interchangeably just fine.

lolli42:
Hey.

PR #548, opened 2021-05-28
[BUGFIX] Circular dependency RenderingContext<->AbstractTemplateView

lolli42:
The constructor of AbstractTemplateView instantiates RenderingContext,
handing itself over as constructor argument.
However, there is no point the RenderingContext knows it's view, and
$this->view is actually completly unused. There is also no getter or
anything accessing the property.
The patch resolves this circular dependency by dropping
the required constructor argument in RenderingContext.

### DIFF
--- a/src/Core/Rendering/RenderingContext.php
+++ b/src/Core/Rendering/RenderingContext.php
@@ -26,7 +26,6 @@ use TYPO3Fluid\Fluid\Core\ViewHelper\ViewHelperInvoker;
 use TYPO3Fluid\Fluid\Core\ViewHelper\ViewHelperResolver;
 use TYPO3Fluid\Fluid\Core\ViewHelper\ViewHelperVariableContainer;
 use TYPO3Fluid\Fluid\View\TemplatePaths;
-use TYPO3Fluid\Fluid\View\ViewInterface;
 
 /**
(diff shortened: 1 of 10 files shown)


mbrodala (approved):

mbrodala:
Apparently the `$view` parameter was added a long time ago but the view was afterwards never exposed:

https://github.com/TYPO3/Fluid/commit/e3837b5980152f87f3a8883dbb0ed2d0c9a51afd

So it is safe to drop this IMO.

NamelessCoder:
For this type of signature-modifying pull request, I would prefer if you left it open for a while or pinged me in order to discuss it before merging. Please compare to this PR that was opened 3.5 years ago:

https://github.com/TYPO3/Fluid/pull/340

This older PR contained additional changes that would have been relatively important:

* Changed return signature for ViewHelperVariableContainer->getView to possibly-null
* Implemented very specific error in f:render if someone misconfigured a custom RenderingContext implementation that caused a view instance to not be present

Although these two changes are only tangentially related, and they can (and should) be done as followups now, if I had the opportunity to review this before merging I would have pointed to the old PR and suggested either to process that one instead, or implement the same changes in this PR in order to avoid a followup.

Or if you had pinged me even before creating this PR I could have pointed to the existing one and saved you some work ;)

Issue #551, opened 2021-06-07
Fluid 3: Recursive object accessor does not work if used as argument


sascha-egerer:
The recursive object accessor does not work if used as parameter / inline.

This does work if used as root node in template:
```
{foo.{subArrayKeyValue}}

or

<f:variable name="myvariable">{foo.{subArrayKeyValue}}</f:variable>
<my:viewhelper id="{myvariable}">
```

This does not work:
```
<my:viewhelper id="{foo.{subArrayKeyValue}}">
```

Error:
```
Array to string conversion
 /xxx/src/Core/Parser/SyntaxTree/EscapingNode.php:48
 /xxx/src/Component/AbstractComponent.php:269
 /xxx/src/ViewHelpers/AliasViewHelper.php:62
 /xxx/src/Component/AbstractComponent.php:269
 /xxx/src/Component/AbstractComponent.php:247
 /xxx/src/Core/Parser/SyntaxTree/EntryNode.php:58
 /xxx/src/Core/Rendering/FluidRenderer.php:103
 /xxx/src/View/AbstractTemplateView.php:120
 /xxx/tests/Functional/BaseFunctionalTestCase.php:97
```

PR #552, opened 2021-06-07
[BUGFIX] Fix recursive object access in parameters

sascha-egerer:
References #551


sascha-egerer:
I've only added unit test for the issue. There is no fix yet

lolli42:
Hey.

PR #554, opened 2021-06-14
[FEATURE] Show useful error message on array to string conversion

sascha-egerer:
If an array is used in a template as variable, the EscapingNode
tries to cast it to a string which does not work but throws an error.
To make the error message easier to understand this case is now
catched and an Exception with the content of the array is shown
instead of only the stacktrace.

### DIFF
--- a/src/Core/Parser/SyntaxTree/EscapingNode.php
+++ b/src/Core/Parser/SyntaxTree/EscapingNode.php
@@ -9,6 +9,7 @@ namespace TYPO3Fluid\Fluid\Core\Parser\SyntaxTree;
 
 use TYPO3Fluid\Fluid\Component\AbstractComponent;
 use TYPO3Fluid\Fluid\Component\ComponentInterface;
+use TYPO3Fluid\Fluid\Core\Parser\Exception;
 use TYPO3Fluid\Fluid\Core\Rendering\RenderingContextInterface;
 
 /**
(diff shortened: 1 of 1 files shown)


NamelessCoder:
We can't safely do this; there are a couple of reasons why:

* Most importantly, this breaks the case of passing an array as content argument for a VH that takes an array as argument and has `escapeChildren = true`
* It also causes exposure of object properties if the array contains objects and error handling is set to output the error message as part of the rendered template (exception caught and output as string).
* It is definitely breaking in cases where error handling is set to not catch the exception and output as part of rendered template.

I understand the motivation for doing this but in the long run I think we are best off (if nothing else then in terms of safety) by letting an array output as string `Array` instead of dumping the properties within the array.

sascha-egerer:
> Most importantly, this breaks the case of passing an array as content argument for a VH that takes an array as argument and has escapeChildren = true
Hm i don't think so. The problem is, that the array will be casted to a string wich ends up in a PHP error. 

> It also causes exposure of object properties if the array contains objects and error handling is set to output the error message as part of the rendered template (exception caught and output as string).
Thats true and i totally agree.

> It is definitely breaking in cases where error handling is set to not catch the exception and output as part of rendered template.
True.

My intention was to have at least a bit of information where the error occurred. It would be a way better to have the template file / line number as pointer where the error is. Something like "can not cast array to string" does not really help in a big set of html files.

I'll rework this patch.

lolli42:
Hey.

PR #555, opened 2021-06-29
[BUGFIX] Create child node copy on evaluation

sascha-egerer:
When a child node is evaluated a copy of the node and
it's arguments must be used. If no copy is used things
like f:for viewhelper will not work as expected as the
nodes do always get the arguments of the first iteration.

### DIFF
--- a/src/Component/AbstractComponent.php
+++ b/src/Component/AbstractComponent.php
@@ -266,7 +266,9 @@ abstract class AbstractComponent implements ComponentInterface
             if ($childNode instanceof EmbeddedComponentInterface) {
                 continue;
             }
-            $evaluatedNodes[] = $childNode->evaluate($renderingContext);
+            $childNodeCopy = clone $childNode;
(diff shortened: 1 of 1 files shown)


sascha-egerer:
Test have to be added

src/Component/AbstractComponent.php
```diff
@@ -266,7 +266,9 @@ protected function evaluateChildNodes(RenderingContextInterface $renderingContex
             if ($childNode instanceof EmbeddedComponentInterface) {
                 continue;
             }
-            $evaluatedNodes[] = $childNode->evaluate($renderingContext);
+            $childNodeCopy = clone $childNode;
```
NamelessCoder, line 269: Might the problem be solved by `$evaluatedNodes[] = $childNode->evaluate(clone $renderingContext);` instead? It shouldn't be necessary to clone the node and the arguments every time - it should be enough to clone the context when evaluating (since the composition of the node and arguments doesn't change; only the context changes).

lolli42:
Hey.

PR #558, opened 2021-08-30
[DOCS] Provide segment about Header/FooterAssets

helsner:
This PR is overtaken from https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-ViewHelper/pull/39 as there are overwrite conflicts as it seems.

Also general notice:
The docs generation for this vh is broken on 9.5 or 10.4 docs version
on master it works fine.
see:
https://docs.typo3.org/other/typo3/view-helper-reference/10.4/en-us/typo3fluid/fluid/latest/Section.html

### DIFF
--- a/src/ViewHelpers/SectionViewHelper.php
+++ b/src/ViewHelpers/SectionViewHelper.php
@@ -62,6 +62,27 @@ use TYPO3Fluid\Fluid\Core\ViewHelper\Traits\ParserRuntimeOnly;
  *     [...]
  *     (depending on the value of {menu})
  *
+ *
+ * Special sections HeaderAssets/FooterAssets
+ * ------------------------------------------
+ *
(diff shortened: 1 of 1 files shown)


NamelessCoder:
Hi @helsner,

Thanks for the pull request - but this information is specific only to TYPO3 CMS and does not belong in the Fluid library. I'm sorry to say this because you already spent time making the same proposal to TYPO3 CMS docs, but this pull request cannot be accepted. Basically, Fluid provides the ability to render a section but TYPO3 CMS contains the feature that renders a specific named section and uses the output as assets.

You might consider making the documentation changes in the docs dealing with the `FLUiDTEMPLATE` content object type and/or the Extbase controller class. These are the two places where TYPO3 CMS attempts to render the assets sections (which also means that those are the only two places where this feature is supported, further indicating this documentation doesn't belong as global documentation for `f:section`).

helsner:
Sounds valid - thx for your answer!

Issue #566, opened 2021-10-21
Broken PHP 8.0 build


lolli42:
See my test run with enabled PHP 8.0 unit tests:
https://github.com/lolli42/Fluid/runs/3956462672?check_suite_focus=true

PR #567, opened 2021-10-22
[TASK] Change expected values to match used php version behaviour

sbuerk:
Some tests has been expecting results on how php <8.0
behaves. These expectations were hard coded. These tests
had further tested raw php execution as a second match.

This patch sets the hardcoded expected values to values
based on the current php version used to run the
testsuite.

Testes with php8.0 enabled build, which demonstrates that
tests working with it. Still have issue with clover file
as current phpunit version do not support it.

See: https://github.com/sbuerk/Fluid/runs/3979490519?check_suite_focus=true

Fixes: #566

### DIFF
--- a/tests/Functional/Cases/Conditions/BasicConditionsTest.php
+++ b/tests/Functional/Cases/Conditions/BasicConditionsTest.php
@@ -19,11 +19,14 @@ class BasicConditionsTest extends BaseConditionalFunctionalTestCase
             ['1 != 2', true],
             ['1 == 2', false],
             ['1 === 1', true],
-            ['\'foo\' == 0', true],
-            ['1.1 >= \'foo\'', true],
+            // expected value based on php versions behaviour
+            ['\'foo\' == 0', (PHP_VERSION_ID < 80000 ? true : false)],
+            // expected value based on php versions behaviour
+            ['1.1 >= \'foo\'', (PHP_VERSION_ID < 80000 ? true : false)],
             ['\'String containing word \"false\" in text\'', true],
             ['\'  FALSE  \'', true],
-            ['\'foo\' > 0', false],
+            // expected value based on php versions behaviour
+            ['\'foo\' > 0', (PHP_VERSION_ID < 80000 ? false : true)],
             ['FALSE', false],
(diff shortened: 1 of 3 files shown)


vertexvaar:
The PHP RFC that changed the behavior in PHP itself is called ["Saner string to number comparisons"](https://wiki.php.net/rfc/string_to_number_comparison), which pretty much sums up what the intention of the PHP developers is. I strongly agree to the change in PHP and i hope fluid will reflect that.
Since this branch forms the start of a new major version, it would be actually acceptable to change the behavior of fluid. 
My opinion is, that fluid should behave like PHP 8.0, even when executed with PHP 7.4, for the following reasons:

* Developers will get used to the behavior of PHP 8 eventually and will expect the same from fluid
* It does not change the expected behavior based on the actually used PHP version
* As the RFC is named, the PHP 8 comparison result is actually saner

Just my $0.02 :wink:

NamelessCoder:
> My opinion is, that fluid should behave like PHP 8.0, even when executed with PHP 7.4

This makes perfect sense to me! We just have to consider if that constitutes a breaking change.

sbuerk:
What also be considered - we are talking here about tests about absolutly rare edge cases which are do not make really sense.

The other thing is, this issue prevent the goal of activating php8.0 testing (and finding eventually more). And further php 8.1 testing.

And the most important point is, that it is **already breaking** as it works differently if executed with php8.0 (which may happen with Typo3 v11 as it offically supports php8.0).

So, I would suggest following @NamelessCoder
- apply this PR to change tests to match current behaviour for php 8.0, so testing for 8.0 may be activated or further issues detected (=> and in the end to 8.1).

- Think about it and if you want to restore it to the "wrong" behavoir you can fix it and revert the test adjustments.

bmack:
I like the approach in checking for the PHP version in the tests. 👍

bmack:
Ping @NamelessCoder @mbrodala anything that needs to be considered? Would be helpful to have proper PHP 8.0/8.1 tests finally...

mbrodala (approved):
Generally OK from my POV since it indeed is about edge-cases.

PR #578, opened 2021-11-15
[task] simplify GroupedFor ViewHelper by always using VariableExtractor

MajPay:
This also adds the possibility of using propertyPaths when grouping
items within a multidimensional array.

Unit tests passed


mbrodala:
Can you add an additional unit test which demonstrates the desired change of behavior? This also ensures that your change will never be dropped accidentally on refactorings.

maddy2101 (approved):

lolli42 (approved):

lolli42:
Thanks Philipp and sorry for the delay.
I moved the tests to functional tests in a pre-patch, and added a test to verify you change with this patch.
Looks good and is merged now.

PR #579, opened 2021-11-15
Feature/docker

MajPay:
Adding tools for development - Just a proposal/draft.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,0 +1,10 @@
+version: "3.9"
+
+services:
+    php74:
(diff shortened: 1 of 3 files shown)


lolli42:
As far as I can see, the main change of this patch has been added as https://github.com/TYPO3/Fluid/pull/578 to the system already, while the docker related changes are side-changes that shouldn't be merge to the repository. I hope it is ok to close here for now and keep https://github.com/TYPO3/Fluid/pull/578 as the main patch.

PR #581, opened 2021-12-22
[BUGFIX] Allow action-only templates to work

bmack:
The current templatePaths functionality is highly limited,
in order to run Fluid with an "ADR" approach where each
Action represents a single PHP class (single responsibility),
Fluid always expected a Controller in order to build the path.

This change addresses the issue when a TemplatePaths object
is built with an empty controller (""), to resolve the paths
correctly when trying to locate the proper file path.

### DIFF
--- a/src/View/TemplatePaths.php
+++ b/src/View/TemplatePaths.php
@@ -565,7 +565,7 @@ class TemplatePaths
             return 'source_' . sha1($this->templateSource) . '_' . $controller . '_' . $action . '_' . $this->getFormat();
         }
         $templatePathAndFilename = $this->resolveTemplateFileForControllerAndActionAndFormat($controller, $action);
-        $prefix = $controller . '_action_' . $action;
+        $prefix = ltrim($controller . '_action_' . $action, '_');
         return $this->createIdentifierForFile($templatePathAndFilename, $prefix);
     }
 
--- a/src/View/TemplatePaths.php
+++ b/src/View/TemplatePaths.php
@@ -258,9 +258,11 @@ class TemplatePaths
         $controller = str_replace('\\', '/', $controller);
         $action = ucfirst($action);
         $identifier = $controller . '/' . $action . '.' . $format;
+        $identifier = ltrim($identifier, '/');
         if (!array_key_exists($identifier, $this->resolvedFiles['templates'])) {
             $templateRootPaths = $this->getTemplateRootPaths();
             foreach ([$controller . '/' . $action, $action] as $possibleRelativePath) {
+                $possibleRelativePath = ltrim($possibleRelativePath, '/');
                 try {
                     return $this->resolvedFiles['templates'][$identifier] = $this->resolveFileInPaths($templateRootPaths, $possibleRelativePath, $format);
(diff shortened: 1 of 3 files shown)


mbrodala:
I kind of understand just from reading, but can we have a few tests for this, as usual at least to avoid regressions?

bmack:
@mbrodala added unit test for covering both "resolveTemplateFileForControllerAndActionAndFormat" and "getTemplateIdentifier"

bmack:
And another one for templates with controllers + actions

src/View/TemplatePaths.php
```diff
@@ -258,9 +258,11 @@ public function resolveTemplateFileForControllerAndActionAndFormat($controller,
         $controller = str_replace('\\', '/', $controller);
         $action = ucfirst($action);
         $identifier = $controller . '/' . $action . '.' . $format;
+        $identifier = ltrim($identifier, '/');
```
mbrodala, line 261: Is there a specific reason for the separate line here but combined line below?

bmack: Thanks for the fast review. No it isn't. But I can adapt this of course. Which variant do you prefer?

mbrodala: IMO this can be merged here, there is no real need for debugging interim values here.

bmack: Thanks, I took care of that

tests/Unit/View/TemplatePathsTest.php
```diff
@@ -281,4 +279,34 @@ public function testGetTemplateIdentifierReturnsSourceChecksumWithControllerAndA
         $instance->setTemplateSource('foobar');
         $this->assertEquals('source_8843d7f92416211de9ebb963ff4ce28125932878_DummyController_dummyAction_html', $instance->getTemplateIdentifier('DummyController', 'dummyAction'));
     }
+
+    /**
+     * @@test
```
mbrodala, line 284:
```suggestion
     * @test
```

tests/Unit/View/TemplatePathsTest.php
```diff
@@ -281,4 +279,34 @@ public function testGetTemplateIdentifierReturnsSourceChecksumWithControllerAndA
...
+        $baseTemplatePath = __DIR__ . '/Fixtures';
+        $subject->setTemplateRootPaths([$baseTemplatePath]);
+        $foundFixture = $subject->resolveTemplateFileForControllerAndActionAndFormat('ARandomController', 'TestTemplate');
+        self::assertEquals($baseTemplatePath . '/ARandomController/TestTemplate.html', $foundFixture);
+        $identifier = $subject->getTemplateIdentifier('ARandomController', 'TestTemplate');
+        self::assertStringStartsWith('ARandomController_action_TestTemplate_', $identifier);
+    }
+
+    /**
+     * @@test
```
mbrodala, line 299:
```suggestion
     * @test
```

mbrodala (changes requested):

mbrodala (approved):

Issue #582, opened 2022-01-14
Setting CompileWithContentArgumentAndRenderStatic::$contentArgumentName emits PHP Fatal Error


sbuerk:
Setting CompileWithContentArgumentAndRenderStatic::$contentArgumentName emits PHP Fatal Error

ViewHelper trait 'CompileWithContentArgumentAndRenderStatic' provides
the magic to set the value of the first optional registered argument
as return value/content of the renderChildrenClosure, if this is provided
in the fluid template. 

This magic should be explicitly settable through the $contentArgumentName
property, which the trait defines and has checks in correpsonding render
methods for it. Because a class using a trait cannot override/redefine a
property which is defined in the trait, this workflow does not work as
described in the comment of this property.

This was detected not working through a core patch which tried to set this
property explicitly, which is stated as better practice in the comment of
the property: https://review.typo3.org/c/Packages/TYPO3.CMS/+/73014

Also this php demostration shows that this is not depending on a php version,
as this generally not works and is not meant to work from php side: https://3v4l.org/r6GqN

Using the constructor or another method is not conveniant enough, thus not
usable in core.

Example error message:

```
PHP Fatal error:  TYPO3Fluid\Fluid\Tests\Functional\Fixtures\ViewHelpers\CompileWithContentArgumentAndRenderStaticExplicitSetArgumentNameForContentViewHelper
and TYPO3Fluid\Fluid\Core\ViewHelper\Traits\CompileWithContentArgumentAndRenderStatic define the same property ($contentArgumentName) in the composition of 
TYPO3Fluid\Fluid\Tests\Functional\Fixtures\ViewHelpers\CompileWithContentArgumentAndRenderStaticExplicitSetArgumentNameForContentViewHelper.

However, the definition differs and is considered incompatible. Class was composed in /home/runner/work/Fluid/Fluid/tests/Functional/Fixtures/ViewHelpers/CompileWithContentArgumentAndRenderStaticExplicitSetArgumentNameForContentViewHelper.php on line 25
```

PR #583, opened 2022-01-14
[TASK] Change property docblock of $contentArgumentName how to set it

sbuerk:
This pull-request contains two commits:

- Adding tests to cover this trait and show that implementation is broken
- Provide the change to document how to set this property correctly

As this is needed for changes in the Typo3 core, it would be nice if
a new patchlevel release could be done after approving and merging this,
so a pending core change can be reviewed and processed:
https://review.typo3.org/c/Packages/TYPO3.CMS/+/73014

Fixes: #582

### DIFF
--- a/src/Core/ViewHelper/Traits/CompileWithContentArgumentAndRenderStatic.php
+++ b/src/Core/ViewHelper/Traits/CompileWithContentArgumentAndRenderStatic.php
@@ -25,11 +25,31 @@ trait CompileWithContentArgumentAndRenderStatic
      * registered in `initializeArguments` of the ViewHelper
      * will be used.
      *
-     * Note: it is significantly better practice to define
+     * Note: it is significantly better practice defining
      * this property in your ViewHelper class and so fix it
      * to one particular argument instead of resolving,
      * especially when your ViewHelper is called multiple
      * times within an uncompiled template!
+     *
+     * This property cannot be directly set in consuming
+     * ViewHelper, instead set the property in ViewHelper
+     * constructor '__construct()', for example with
+     * $this->contentArgumentName = 'explicitlyToUseArgumentName';
+     *
+     * Another possible way would be to override the method
+     * 'resolveContentArgumentName()' and return the name.
+     *
+     * public function resolveContentArgumentName()
+     * {
+     *     return 'explicitlyToUseArgumentName';
+     * }
+     *
+     * Note: Setting this through 'initializeArguments()' will
+     *       not work as expected, and other methods should be
+     *       avoided to override this.
+     *
+     * Following test ViewHelpers are tested and demonstrates
+     * that the setting posibillities works.
      *
      * @var string
      */
@@ -66,7 +86,7 @@ trait CompileWithContentArgumentAndRenderStatic
         ViewHelperNode $node,
         TemplateCompiler $compiler
     ) {
-        list ($initialization, $execution) = ViewHelperCompiler::getInstance()->compileWithCallToStaticMethod(
+        list($initialization, $execution) = ViewHelperCompiler::getInstance()->compileWithCallToStaticMethod(
             $this,
             $argumentsName,
             $closureName,
@@ -128,7 +148,7 @@ trait CompileWithContentArgumentAndRenderStatic
             }
             throw new Exception(
                 sprintf('Attempting to compile %s failed. Chosen compile method requires that ViewHelper has ' .
-                'at least one registered and optional argument', __CLASS__)
+                    'at least one registered and optional argument', __CLASS__)
             );
         }
         return $this->contentArgumentName;
--- a/tests/Functional/Cases/CompileWithContentArgumentAndRenderStaticTest.php
+++ b/tests/Functional/Cases/CompileWithContentArgumentAndRenderStaticTest.php
@@ -1,0 +1,150 @@
+<?php
+
+namespace TYPO3Fluid\Fluid\Tests\Functional\Cases;
+
+use TYPO3Fluid\Fluid\Core\Cache\FluidCacheInterface;
+use TYPO3Fluid\Fluid\Core\Cache\SimpleFileCache;
+use TYPO3Fluid\Fluid\Tests\Functional\BaseFunctionalTestCase;
+
+class CompileWithContentArgumentAndRenderStaticTest extends BaseFunctionalTestCase
+{
+    /**
+     * If your test case requires a cache, override this
+     * method and return an instance.
+     *
+     * @return FluidCacheInterface
+     */
+    protected function getCache()
+    {
+        return new SimpleFileCache(sys_get_temp_dir());
+    }
+
+    public function getTemplateCodeFixturesAndExpectations(): array
+    {
+        return [
+            // with trait but without contentArgumentProperty set in viewhelper and having optional argument
+            'children content but no argument value' => [
+                '<test:compileWithContentArgumentAndRenderStaticUseFirstRegisteredOptionalArgumentAsRenderChildren>mustBeRenderingChildrenClosure</test:compileWithContentArgumentAndRenderStaticUseFirstRegisteredOptionalArgumentAsRenderChildren>',
+                [],
+                [
+                    '"arguments[firstOptionalArgument]": null',
+                    '"renderChildrenClosure": "mustBeRenderingChildrenClosure"',
+                ],
+                [],
+                null,
+                true
+            ],
+            'children content and argument value' => [
+                '<test:compileWithContentArgumentAndRenderStaticUseFirstRegisteredOptionalArgumentAsRenderChildren firstOptionalArgument="firstOptionalArgument">mustBeRenderingChildrenClosure</test:compileWithContentArgumentAndRenderStaticUseFirstRegisteredOptionalArgumentAsRenderChildren>',
+                [],
+                [
+                    '"arguments[firstOptionalArgument]": "firstOptionalArgument"',
+                    '"renderChildrenClosure": "firstOptionalArgument"',
+                ],
+                [],
+                null,
+                true
+            ],
+            // with trait but without contentArgumentProperty set in viewhelper and having first optional argument as second argument
+            'children content but no argument value [optional is second argument]' => [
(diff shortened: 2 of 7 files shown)


lolli42 (approved):

src/Core/ViewHelper/Traits/CompileWithContentArgumentAndRenderStatic.php
```diff
@@ -118,7 +121,7 @@ protected function buildRenderChildrenClosure()
      */
     public function resolveContentArgumentName()
     {
-        if (empty($this->contentArgumentName)) {
+        if (!property_exists($this, 'contentArgumentName') || empty($this->contentArgumentName)) {
```
lolli42, line 124: maybe we should split !property_exists() and empty() here, to avoid writing to a non-existing property below - if the VH does not have the property, we could fall back to a second property that is exclusively used within the trait, or just do the magic each time.
Note accessing a non-existing class property will log a deprecation note with php 8.2

sbuerk: Done. Please re-review if it's okay now.

src/Core/ViewHelper/Traits/CompileWithContentArgumentAndRenderStatic.php
```diff
@@ -66,7 +75,7 @@ public function compile(
         ViewHelperNode $node,
         TemplateCompiler $compiler
     ) {
-        list ($initialization, $execution) = ViewHelperCompiler::getInstance()->compileWithCallToStaticMethod(
+        [$initialization, $execution] = ViewHelperCompiler::getInstance()->compileWithCallToStaticMethod(
```
mbrodala, line 78: Since [Fluid still supports PHP >= 5.5](https://github.com/TYPO3/Fluid/blob/3d8499ae81e1fda39d236072f3ef8a7ef35433b1/composer.json#L6) we cannot do this.

src/Core/ViewHelper/Traits/CompileWithContentArgumentAndRenderStatic.php
```diff
@@ -118,19 +127,30 @@ protected function buildRenderChildrenClosure()
...
     {
-        if (empty($this->contentArgumentName)) {
-            $registeredArguments = call_user_func_array([$this, 'prepareArguments'], []);
-            foreach ($registeredArguments as $registeredArgument) {
-                if (!$registeredArgument->isRequired()) {
-                    $this->contentArgumentName = $registeredArgument->getName();
-                    return $this->contentArgumentName;
-                }
+        // early return already defined argument name
+        if ($this->useContentArgumentName !== null) {
```
mbrodala, line 131: Would a local `static` variable work, too? :thinking:

mbrodala (changes requested):

src/Core/ViewHelper/Traits/CompileWithContentArgumentAndRenderStatic.php
sbuerk: How came that ? This was a non intended change and not done activly oO. Will revert this.

src/Core/ViewHelper/Traits/CompileWithContentArgumentAndRenderStatic.php
sbuerk: mhhhh .....

sbuerk: I thing static would introduce some nasty flaws, see: https://3v4l.org/r8C8D

mbrodala: That's ... interesting. It surely makes sense somewhere but yes, too much risk here.

NamelessCoder:
TBF, I don't really like the direction this is going - the added property that's nearly the same as the existing property, plus the involvement of a property that may or may not exist. May I suggest instead that we adjust the phpdoc and mention that overriding the `resolveContentArgumentName` method is the way to go? That fits better with composition logic and traits, at least IMHO. Plus it's testable in implementers without also invoking the trait's methods.

It also seems excessive to me to add three full VH fixtures with the only purpose of testing the behavior of this trait.

sbuerk:
> TBF, I don't really like the direction this is going - the added property that's nearly the same as the existing property, plus the involvement of a property that may or may not exist. May I suggest instead that we adjust the phpdoc and mention that overriding the `resolveContentArgumentName` method is the way to go? That fits better with composition logic and traits, at least IMHO. Plus it's testable in implementers without also invoking the trait's methods.

Yes, right. I have reverted my changes in my commit and added the changed docblock of the
property to describe how to correctly set this, and removed the wrong statement to simply
copy it. Generally setting it in __constructor seems to work, at least in tests, thus adding a 4th ViewHelper fixture.

> It also seems excessive to me to add three full VH fixtures with the only purpose of testing the behavior of this trait.

It is far away from excessig. At first, there was not any coverage of this trait and for this
behavior, so the first two fixture ViewHelpers are there to demonstrate the normal resolving
behaviour which already worked but was not tested explicitly.

The 3rd or better to say 3rd and 4th are there to test/demonstrates the behaviour using the constructor or overriding the resolveContentArgumentName method and that both of them works.

Personally I would add a fifth, demonstrating that setting it in 'initializeArguments()' for example wouldn't work - but left that out, as you was already against it.

Anyway the 4th fixture should stay as they are testing behiouver which *must* be covered and should not be left out. And doing the different constalltions are only possible with different ViewHelpers. Using fixture ViewHelpers decouples this from shipped viewhelpers as it tests the trait behaviour, not a special ViewHelper - which was the goal of these tests.

src/Core/ViewHelper/Traits/CompileWithContentArgumentAndRenderStatic.php
```diff
@@ -25,12 +25,34 @@ trait CompileWithContentArgumentAndRenderStatic
...
      * especially when your ViewHelper is called multiple
      * times within an uncompiled template!
      *
+     * This property cannot be directly set in consuming
+     * ViewHelper, instead set the property in ViewHelper
+     * constructor '__construct()', for example with
+     * $this->contentArgumentName = 'explicitlyToUseArgumentName';
+     *
+     * Another possible way would be to override the method
+     * 'resolveContentArgumentName()' and returning the name.
```
mbrodala, line 40:
```suggestion
     * 'resolveContentArgumentName()' and return the name.
```

src/Core/ViewHelper/Traits/CompileWithContentArgumentAndRenderStatic.php
```diff
@@ -25,12 +25,34 @@ trait CompileWithContentArgumentAndRenderStatic
...
+     *
+     * public function resolveContentArgumentName()
+     * {
+     *     return 'explicitlyToUseArgumentName';
+     * }
+     *
+     * Note: Setting this through 'initializeArguments()' do
```
mbrodala, line 47:
```suggestion
     * Note: Setting this through 'initializeArguments()' will
```

src/Core/ViewHelper/Traits/CompileWithContentArgumentAndRenderStatic.php
```diff
@@ -25,12 +25,34 @@ trait CompileWithContentArgumentAndRenderStatic
...
+     *       not work as it may be expected, and the other
```
mbrodala, line 48:
```suggestion
     *       not work as expected, and other
```

src/Core/ViewHelper/Traits/CompileWithContentArgumentAndRenderStatic.php
```diff
@@ -25,12 +25,34 @@ trait CompileWithContentArgumentAndRenderStatic
...
+     *       methods should be avoided to override for this.
```
mbrodala, line 49:
```suggestion
     *       methods should be avoided to override this.
```

src/Core/ViewHelper/Traits/CompileWithContentArgumentAndRenderStatic.php
```diff
@@ -25,12 +25,34 @@ trait CompileWithContentArgumentAndRenderStatic
...
+     *
+     * Following test ViewHelpers are tested and demonstrates
+     * that the setting posibillities works.
+     * @see CompileWithContentArgumentAndRenderStaticExplicitSetArgumentNameForContentInConstructorViewHelper
+     * @see CompileWithContentArgumentAndRenderStaticExplicitSetArgumentNameForContentOverriddenResolveContentArgumentNameMethodViewHelper
```
mbrodala, line 54: This should be dropped, since it doesn't add value to the consumer and will get outdated over time.

mbrodala (changes requested):

src/Core/ViewHelper/Traits/CompileWithContentArgumentAndRenderStatic.php
sbuerk: done

src/Core/ViewHelper/Traits/CompileWithContentArgumentAndRenderStatic.php
sbuerk: done

src/Core/ViewHelper/Traits/CompileWithContentArgumentAndRenderStatic.php
sbuerk: done

src/Core/ViewHelper/Traits/CompileWithContentArgumentAndRenderStatic.php
sbuerk: done

lolli42 (approved):

lolli42:
ping. could we get this done?

mbrodala (approved):
Perfectly fine from my POV.

NamelessCoder (approved):

NamelessCoder:
Thanks for taking care of this, @sbuerk - and for being receptive to the implementation input during review ;)

PR #586, opened 2022-02-20
[TASK] Add TemplateAwareViewInterface

lolli42:
Method render() of AbstractTemplateView has an
optional argument to provide a template name.
This is not reflected in ViewInterface since
ViewInterface does not know about templates.

The patch adds an optional TemplateAwareViewInterface
to reflect this argument in 'template aware'
views like TemplateView.

### DIFF
--- a/src/View/TemplateAwareViewInterface.php
+++ b/src/View/TemplateAwareViewInterface.php
@@ -1,0 +1,23 @@
+<?php
+declare(strict_types=1);
+namespace TYPO3Fluid\Fluid\View;
+
+/*
+ * This file belongs to the package "TYPO3 Fluid".
+ * See LICENSE.txt that was shipped with this package.
+ */
+
+/**
+ * Optional addition to ViewInterface if the view deals with template files.
+ *
+ * @api
+ */
+interface TemplateAwareViewInterface
+{
+    /**
+     * @param string A template name to render, e.g. "Main/Index"
+     * @return string The rendered view
+     * @api
+     */
+    public function render(string $templateName = '');
+}
--- a/src/View/AbstractTemplateView.php
+++ b/src/View/AbstractTemplateView.php
@@ -22,7 +22,7 @@ use TYPO3Fluid\Fluid\ViewHelpers\SectionViewHelper;
  *
  * Contains the fundamental methods which any Fluid based template view needs.
  */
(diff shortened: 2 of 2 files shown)


lolli42:
This especially helps static code analyzers and IDE's to understand the render argument, see
https://review.typo3.org/c/Packages/TYPO3.CMS/+/73611

sbuerk (approved):

oliverklee (approved):

src/View/TemplateAwareViewInterface.php
```diff
@@ -0,0 +1,23 @@
...
+ * @api
+ */
+interface TemplateAwareViewInterface
+{
+    /**
+     * @param string A template name to render, e.g. "Main/Index"
+     * @return string The rendered view
+     * @api
+     */
+    public function render(string $templateName = '');
```
mbrodala, line 22: Hm, I'd rather prefer a `renderTemplate()` or `renderTemplateName()`. This would avoid the implicit override of `ViewInterface::render()` and make for a clearer distinction/API.

lolli42: That wouldn't reflect the current given situation in AbstractTemplateView / TemplateView. Do you really want to change this, and add a second method to TemplateView that has an identical signature?

mbrodala: I see what you mean. This by itself is already a violation of the interface IMO even if PHP permits this.

Can you add more context about the issue you are trying to solve? E.g. why type-hinting to `AbstractTemplateView` instead of `ViewInterface` is not an option.

lolli42: Well, in general one should code against interfaces, not implementations. The TYPO3 core type hints ViewInterface at a couple of places, and that's the way to go. It's just that render('Foo') is then not understood.
The patch simply adds a second interface to this, making this argument visible. This is fully PHP compatible and fine with liskov: When an interface declares the method, a second interface can add optional arguments.
I see no reason to not pick this patch.

mbrodala: Make sense then, especially if it helps making at least the runtime invocation of `render()` with an argument explicit.

mbrodala (approved):

PR #607, opened 2022-05-23
[TASK] Stop exporting ViewHelperBaseTestcase

### LINKED ISSUE #562: ViewHelperBaseTestcase not usable as base class for testing viewhelpers
I have an own extension where I want to test some viewhelpers with the ViewHelperBaseTestcase.
Sadly this class is in the "tests" directory which means it is not autoloaded when requesting it from external extensions.

Secondly, the ViewHelperBaseTestcase has a dependency to RenderingContextFixture which is excluded to be exported via `.gitattributes`. This means I need to copy all the files to my own extension which is really not what I want.

lolli42:
The class is an abstract for unit testing VH's. Exporting
it does not make sense since it has a direct dependency to
RenderingContextFixture is setup(), which is not exported.

The patch excludes the file from packaging again, which
also allows Fluid to refactor internally without being
breaking.

Resolves #562

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -6,4 +6,3 @@ tests/Functional/ export-ignore
 tests/Unit/*/ export-ignore
 tests/Unit/ViewHelpers/ -export-ignore
 tests/Unit/ViewHelpers/* export-ignore
-tests/Unit/ViewHelpers/ViewHelperBaseTestcase.php -export-ignore


simonschaufi:
@lolli42 you can exclude all tests, not just this one

mbrodala:
I agree, it does not make any sense to export tests in a library package.

lolli42:
Hmm. This one is obvious and easy to argue.

Let's do the other ones in a dedicated patch? UnitTestCase for instance is useless since it's empty. But I'm currently streamlining the tests anyways, so let's wait with the other files a bit, bring this one in, and see what is left when the test-setup is more clean?

mbrodala (approved):

mbrodala:
Perfectly fine with me. :+1:

PR #609, opened 2022-05-23
[TASK] Simplify ExamplesTest output assertion

lolli42:
Preparing phpunit 9, the closure output assertion
needs to be streamlined a bit to continue to work.


lolli42:
This should be safe: Changing an expectation leads to a test fail. I'll merge.

PR #611, opened 2022-05-23
[TASK] More strict phpunit.xml.dist

lolli42:
Fail more often, report more errors and
reference phpunit 9.5 xsi.


lolli42:
Guess it's fine to bring this in: It's rather obvious and test-related "only".

PR #612, opened 2022-05-23
[TASK] Add a standard .editorconfig

lolli42:
This should align IDE's to casual defaults.
Also tune .gitattributes a bit.


lolli42:
This should be fine for now. In case we want to derive with other file types, we can easily adapt later.

PR #613, opened 2022-05-23
[TASK] Add phpstan level 4, add baseline, add CI

lolli42:
Have a default phpstan config with PHP 7.2.
Level 4 seems to be suitable for now. Add a
baseline file and enable phpstan in CI.

With PHP 7.2:
composer req --dev phpstan/phpstan
vendor/bin/phpstan --generate-baseline

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -12,3 +12,6 @@ trim_trailing_whitespace = true
 
 [*.{yaml,yml}]
 indent_size = 2
+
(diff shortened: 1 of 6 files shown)


lolli42:
I'll go ahead an merge this: We should be able to reduce baseline quite a bit with a couple of dedicated patches that don't affect b/w compat.

PR #614, opened 2022-05-23
[TASK] Add phpstan/phpstan-phpunit and regenerate baseline

lolli42:
Using PHP 7.2:
composer req --dev phpstan/phpstan-phpunit
./vendor/bin/phpstan --generate-baseline


lolli42:
This should be fine to go to make phpstan understand some phpunit quirks.
We'll reduce baseline with next patches.

PR #616, opened 2022-05-25
[TASK] Revamp examples

lolli42:
* Mimic a real project having a src/ sub folder
  for classes, including a composer autoload-dev
* Encapsulate base helper in a class
* Exclude examples/ folder from packaging
* Update README.md regarding examples
* Cache dir is now in system temp (/tmp/ on unix)


README.md
```diff
@@ -15,13 +15,17 @@ Installation
...
-
-> Tip: you can execute all examples in the same run by calling
->
-> `find examples/ -maxdepth 1 -name \*.php -exec php {} \;`
+The library source comes with a set of example scripts to study and play around with.
+They are PHP entry scripts which render templates, their partials and layouts.
+
+```bash
+$ git clone git@github.com:TYPO3/Fluid.git
+$ composer update
```
mbrodala, line 23:
```suggestion
$ composer install
```

mbrodala (changes requested):

README.md
lolli42: No. Update is fine since there is no composer.lock. Recent composer versions even warn about this, using 'install' when there is no .lock.

mbrodala: As discussed personally I see `install` as the safer option in all cases but given that we have a library package here always updating all deps to their latest version should be OK.

mbrodala (approved):

PR #619, opened 2022-07-28
[BUGFIX] Correct cached ternary expressions

lolli42:
Ternary expressions are broken with cached templates.
Even with simple things like this:

$view->assign('myVariable', false);
{myVariable ? 'yes' : 'no'}

This evaluates correctly to 'no' on first call with cold
caches, but to 'yes' when a cached template is used
on subsequent calls.

This came up when the functional tests have recently been
fixed to correctly test cached behavior.

The fix is to evaluate() the "left" / "check" side just
like the non-cached variant does.

### DIFF
--- a/src/Core/Parser/SyntaxTree/Expression/TernaryExpressionNode.php
+++ b/src/Core/Parser/SyntaxTree/Expression/TernaryExpressionNode.php
@@ -9,7 +9,6 @@ namespace TYPO3Fluid\Fluid\Core\Parser\SyntaxTree\Expression;
 
 use TYPO3Fluid\Fluid\Core\Compiler\TemplateCompiler;
 use TYPO3Fluid\Fluid\Core\Parser\BooleanParser;
-use TYPO3Fluid\Fluid\Core\Parser\SyntaxTree\BooleanNode;
 use TYPO3Fluid\Fluid\Core\Rendering\RenderingContextInterface;
 
 /**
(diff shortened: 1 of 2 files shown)


sbuerk (approved):

src/Core/Parser/SyntaxTree/Expression/TernaryExpressionNode.php
```diff
@@ -123,23 +122,27 @@ public function compile(TemplateCompiler $templateCompiler)
...
         $initializationPhpCode .= sprintf(
             '%s = function($context, $renderingContext) {
-				if (%s::convertToBoolean(' . $compiledExpression . ', $renderingContext) === TRUE) {
-					return %s::getTemplateVariableOrValueItself(%s, $renderingContext);
-				} else {
-					return %s::getTemplateVariableOrValueItself(%s, $renderingContext);
-				}
+                $check = %s;
+                $parser = new \TYPO3Fluid\Fluid\Core\Parser\BooleanParser();
+                $checkResult = $parser->evaluate($check, $context);
```
lolli42, line 138: This is the main change: We're not using ->compile() anymore for the time being, but simply parse $check again:

It seems the code returned by ->compile() has at least some quoting issues.

We switch to full evaluation here for the time being to fix the cache-related bugs first.

Looking at ->compile() and maybe using it here again to optimize evaluation is another task, but fixing the main issues of cached-behavior has priority for now.

lolli42:
This got a "this looks sensible to me" from Claus, too (via PM)

PR #634, opened 2022-08-01
[TASK] Remove "Fluid 3.x" compat classes

lolli42:
A couple of compatibility classes have been added
to Fluid 2.x aiming for 3.x compatibility. Fluid 3.x
however won't make it as such.

The patch removes these classes in 2.x again. This is
not considered breaking since Fluid 3 has never been
tagged.


NamelessCoder:
Hi @lolli42,

Can you elaborate a bit about which decisions have been made here? Does this mean that all of the work I did for Fluid 3.x is getting deleted and if so, what is the reasoning behind that decision?

TIA

PR #636, opened 2022-08-02
[TASK] Refactor for VH tests to functionals

maddy2101:
In order to provide reliable testing scenarios for ViewHelpers,
we prepare real world application of passing templates into
the Fluid Parser and comparing the result over isolated
variable passage into VH Helper methods.

With the refactoring also caching can now be tested.

Most f:for tests have been refactored to functional tests
with this patch.


maddy2101:
I left some of the tests, that aim at the internal behavior, because those need the mocked setup.

tests/Functional/ViewHelpers/ForViewHelperTest.php
```diff
@@ -0,0 +1,130 @@
...
+            '<f:for each="{value}" as="item" reverse="true">{item} </f:for>',
+            ['value' => $value],
+            '3 2 1 0 ',
+        ];
+        $value = new \ArrayObject(['key1' => 'value1', 'key2' => 'value2']);
+        yield 'object gets traversed' => [
+            '<f:for each="{value}" as="item">{item} </f:for>',
+            ['value' => $value],
+            'value1 value2 ',
+        ];
```
s2b, line 83: A test case could be added that combines `reverse` and associative arrays.

lolli42 (approved):

PR #643, opened 2022-08-05
[TASK] Refactor switch/case/defaultCase VH tests to functionals

s2b:
In order to provide reliable testing scenarios for ViewHelpers,
we prepare real world application of passing templates into
the Fluid Parser and comparing the result over isolated
variable passage into VH Helper methods.

With the refactoring also caching can now be tested.

All f:switch/f:case/f:defaultCase tests have been refactored to functional tests
with this patch.


lolli42:
Nice. So this one reveals another cache or state issue.

lolli42 (approved):

PR #654, opened 2022-08-12
[BUGFIX] ParserRuntimeOnly trait compile() returns string

lolli42:
ViewHelperInterface->compile() is annotated to return
string. ParserRuntimeOnly violates this and returns
null instead. This leads to different behavior in
consuming ViewHelpers between non-compiled and compiled
versions.

The change adapts ParserRuntimeOnly->compile() to
return an empty string and changes NodeConverter to
deal with it. A string cast is added to stay compatible
in case some other compile() method returns null as well.

### DIFF
--- a/src/Core/ViewHelper/Traits/ParserRuntimeOnly.php
+++ b/src/Core/ViewHelper/Traits/ParserRuntimeOnly.php
@@ -29,6 +29,6 @@ trait ParserRuntimeOnly
      */
     public function compile($argumentsName, $closureName, &$initializationPhpCode, ViewHelperNode $node, TemplateCompiler $compiler)
     {
-        return null;
+        return '';
     }
 }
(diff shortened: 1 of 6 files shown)


src/Core/ViewHelper/Traits/ParserRuntimeOnly.php
```diff
@@ -29,6 +29,6 @@ public function render()
      */
```
sbuerk, line 29: Should we not add at least the return type as docblock here ?

```
@return string
```

sbuerk (changes requested):

src/Core/ViewHelper/Traits/ParserRuntimeOnly.php
lolli42: done

PR #663, opened 2022-09-08
[BUGFIX] Support multiple references in variable names

dablach:
A small ajustement to the used regex and repeated matching calls make it possible to support multipe reference in variable names. This allowes to use for example `{foo}bar{baz}` as a variable name.

An exception is thrown if resolved variables would introduce new references. Otherwise variables would get resolved, that are not visible to the author of the template, which looks a bit off security-wise.

### DIFF
--- a/src/Core/Variables/StandardVariableProvider.php
+++ b/src/Core/Variables/StandardVariableProvider.php
@@ -7,6 +7,8 @@
 
 namespace TYPO3Fluid\Fluid\Core\Variables;
 
+use TYPO3Fluid\Fluid\Core\Exception;
+
 /**
  * Class StandardVariableProvider
  */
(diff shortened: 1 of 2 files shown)


src/Core/Variables/StandardVariableProvider.php
```diff
@@ -251,10 +253,15 @@ public function getAccessorsForPath($propertyPath)
...
-            preg_match_all('/(\{.*\})/', $propertyPath, $matches);
-            foreach ($matches[1] as $match) {
-                $subPropertyPath = substr($match, 1, -1);
-                $propertyPath = str_replace($match, $this->getByPath($subPropertyPath), $propertyPath);
+            while (preg_match_all('/(\{[^{}]*\})/', $propertyPath, $matches) > 0) {
+                foreach ($matches[1] as $match) {
+                    $subPropertyPath = substr($match, 1, -1);
+                    $replacement = $this->getByPath($subPropertyPath);
+                    if (strpos($replacement, '{') !== false) {
+                        throw new Exception('Variables referenced in variable names must not introduce new references.');
```
lolli42, line 261: could we add a test to show such an invalid construct?

lolli42:
This looks good at first glance, but I'm a bit unsure on the full impact of this change and would love to hear further opinions on this.

lolli42:
Closing this in favor of https://github.com/TYPO3/Fluid/pull/738

PR #671, opened 2022-12-09
[FEATURE] Allow setting keys of objects

spthiel:
Add an additional option to set the name to a period seperated path for modifying of inner values of arrays/objects for the ability to expand objects during fluid template validation

An example usecase for this would be when rendering a partial for a range input and allowing the passing of min and max as seperate parameters to additionalParameters to the f:render to then merge them inside the partial like this

Template.html
```html
<f:render partial="RangeInput" min="0" max="100" additionalAttributes="{data-target: "rangeinput"}"
```

RangeInput.html
```html
<f:variable name="additionalAttributes.min" value="{min}" />
<f:variable name="additionalAttributes.max" value="{max}" />
<f:form.textfield type="range" additionalAttributes="{additionalAttributes}"/>
```

This change allows for the parent to be null so in above example the additionalAttributes key is optional and if not passed the resulting additionalAttributes array would just look like ["min" => 0, "max" => 100]


spthiel:
Saw today it's included in #425

spthiel:
Reopen as #425 was closed for the scope being too large

spthiel:
I definitely want to avoid anything that mutates data outside fluid (i.e. php objects) the goal of this is exclusively to allow for setting of further keys on fluid template variables

spthiel:
I can accept array union as a solution to this issue

Issue #691, opened 2023-03-22
Error in 2.7.3 at flow:package:rescan


patriceckhart:
Version: 2.7.3

When I run the `flow flow:package:rescan` I get the following error. The error does not occur with version 2.7.2.

Error: 
Warning: Undefined variable $argv in /data/neos/Packages/Libraries/typo3fluid/fluid/src/Tools/console.php line 22

fcool:
Sadly this is no bug in typo3fluid but a bug in Flow's Object-Management.

It seems like enabling the Object-Management for `typo3fluid` is not necessary at all. At least in one project I testet it was sufficient to disable it for `typo3fluid` like that in Flow's Settings.yaml:
```yaml
Neos:
  Flow:
    object:
      includeClasses:
        'typo3fluid.fluid': []
```

paavo:
Mh, why is this already closed? 🤔

fcool:
Because this is not to fix by `typo3fluid/fluid` - it has to be fixed in `neos/flow-development-collection`

The only thin `fluid` "could" fix is the PSR-4 violation, but that would not solve the error reported here ;)

lolli42:
Would it help flow to merge the content from console.php back into bin/fluid and remove console.php again?

kdambekalns:
> Would it help flow to merge the content from console.php back into bin/fluid and remove console.php again?

Yes, that would help. Flow tries to parse `console.php` as a class file, and fails (since that file directly "runs" when being included, something we don't expect in a folder supposed to contain classes.) So if the file disappeared, that would solve our issue.

kdambekalns:
Our workaround would (for now) be https://github.com/neos/flow-development-collection/pull/2994 – with the drawback of breaking XSD schema generation (but I have no clue how many people use that.)

lolli42:
Thanks for your work Christoph!

I agree having global code in src/ is kinda ugly anyways, and it's better to just keep that stuff in bin/fluid.

The detail with the "class name" is imho a bug in Neos/Flow's class compiler and it would be great if that could be fixed/relaxed somehow. I'll of course keep it in the patch, but can't guarantee similar things don't ever pop up again.

I tend to merge the PR soon and will tag a 2.7.4 right away. HTH.

lolli42:
Tagged 2.7.4

paavo:
> Because this is not to fix by `typo3fluid/fluid` - it has to be fixed in `neos/flow-development-collection`
> 
[...]

PR #693, opened 2023-03-22
[TASK] Support Neos/Flow class compiler

christophlehmann:
This moves back logic from console.php to bin/fluid in order to support Neos/Flow class compiler and resolves the warning:

Undefined variable $argv in Packages/Libraries/typo3fluid/fluid/src/Tools/console.php line 22

The Heredoc update also help the class compiler because it interpretes 'name' in the line

class name

as a class name which leads to the exception:

The name of the class "name" is not the same as the filename which is "ConsoleRunner.php".

Resolves: #691


christophlehmann:
Would be nice to see it merged, but could fully understand if not and that both changes should better be fixed in Neos/Flow.

lolli42 (approved):

Issue #455, opened 2019-05-21
Allow trailing comma in arrays


mbrodala:
A *very* common error in Fluid templates is a trailing comma after the last array item:

```html
<f:render partial="MyPartial" arguments="{
  first: 1,
  second: 2,
}"/>
```

This leads to the whole array being interpreted as string instead and is hard to track.

For this reason we really should allow trailing commas in arrays. Even [IE9 agrees](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas#Browser_compatibility).

NamelessCoder:
The sequencer (symbol-and-capture yielding, context-switching tokenisation) strategy I am writing as a replacement for the current regular expression based parsing will support this. The reason for the current lack of support is that the separator isn't optional.

https://github.com/NamelessCoder/Fluid/commits/feature/sequencer

The following will also be supported:

```
{{foo: bar} | v:h()} // pass ['foo' => 'bar'] as child node.
{{foo, bar} | v:h()} // pass ['foo' => $foo, 'bar' => $bar] as child node
{[foo, bar] | v:h()} // pass [0 => $foo, 1 => $bar] as child node
```

With the same array syntax supported in arguments. Inline pass used in example because that ability is *also* new :)

mbrodala:
@NamelessCoder Sounds really useful. Just to be clear, the following would be valid as well then?

```
{{foo: bar,} | v:h()}
{{foo, bar,} | v:h()}
{[foo, bar,] | v:h()}
// Plain argument syntax
{v:h({foo: bar,})}
{v:h({foo, bar,})}
{v:h([foo, bar,])}
```

NamelessCoder:
Not exactly - here's a more complete version:

```
{v:h(foo)} // call v:h with foo=$foo
{v:h(foo, bar)} // call v:h with foo=$foo and bar=$bar
                        // VH arguments support ECMA literal property shorthand only
{v:h(array: [foo, bar,])} // supported, last comma ignored
{v:h(array: {foo, bar,})} // supported, last comma ignored
{v:h(array: {foo: x, bar: y,})} // supported, last comma ignored
{v:h(array: ["x"="y", "z"= "m"])} // supported
{v:h(array: ["{dynamic}"="y"])} // supported
{v:h(array: ["{dynamic}": "y"])} // supported
{v:h(array: ["{v:h()}": "value"])} // supported
```

mbrodala:
[...]

PR #696, opened 2023-04-19
[BUGFIX] Allow comma with trailing whitespace

s2b:
With this change, the syntax to define arrays as well as the inline syntax for viewhelpers allow trailing whitespace characters after the last comma. This is now possible:

Array syntax:

```xml
<f:variable name="myArray" value="{
    0: 'first item',
    1: 'second item',
}" />
```

Inline viewhelper syntax:

```xml
{f:if(
    condition: myVariable,
    then: 'yes',
    else: 'no',
)}
```


s2b:
Fixes: #455

lolli42 (approved):

lolli42:
This is super helpful. Love it!
Looked through the code, played around a bit. Looks good to me!

minapok (approved):

Issue #697, opened 2023-04-26
Remove usages of $self in compiled templates


IndyIndyIndy:
The variable $self is used in some places, when compiling the templates, but never actually used und just clutters up the cache.

PR #698, opened 2023-04-26
[TASK] Remove usages of $self in compiled templates

IndyIndyIndy:
Resolves: #697


lolli42:
Strictly speaking, this *may* be breaking since VH's compile() methods may use defined $self to create template php code. Removing that will break this.
However, I did not find any usages of $self anywhere ...

s2b (approved):
As mentioned by lolli, this might be breaking, but I can't think of a use case and it is easily revertible.

lolli42 (approved):

Issue #701, opened 2023-04-28
Switch to static functions in closures


IndyIndyIndy:
Another little optimization, that can be done. From what I see, none of the closure functions use $this,

lolli42:
Hey.

I don't think we can do this at the moment. The compiled template code currently creates tons of closures, and some of them aren't needed at all, or could be substituted with a more simple `static fn () => '';`. We'll have disentangle the code generation further to really use that. See also my current headaches with if VH in https://github.com/TYPO3/Fluid/issues/771 and https://github.com/TYPO3/Fluid/issues/769

I'll close here for now since the PR won't make it as such.

PR #702, opened 2023-04-28
[TASK] Switch to static functions in closures

IndyIndyIndy:
Resolves: #701


lolli42:
This is again potentially breaking since it will explode when you access $this.
I'd say this is more critical than the $self case and would say we shouldn't do it in 2.x.

lolli42:
@IndyIndyIndy I'm not 100% sure about impact. What is your assessment? Maybe it's better to reduce this change to those places that are "safe" and don't affect API?

lolli42:
Rejecting for now. See issue.

PR #719, opened 2023-05-01
[FEATURE] Add runtime cache for object access in fluid

s2b:
This change reduces PHP function calls when accessing multiple instances of the same class in a fluid template. The current implementation checks the class structure with each property access for the availability of getter or asserter methods, which can be cached after the first check and re-used for all subsequent accesses of the same property.

Due to the way variable providers are implemented, this cache can also be re-used in partials, layouts and sections.

### DIFF
--- a/src/Core/Variables/StandardVariableProvider.php
+++ b/src/Core/Variables/StandardVariableProvider.php
@@ -12,17 +12,19 @@ namespace TYPO3Fluid\Fluid\Core\Variables;
  */
 class StandardVariableProvider implements VariableProviderInterface
 {
-    const ACCESSOR_ARRAY = 'array';
-    const ACCESSOR_GETTER = 'getter';
-    const ACCESSOR_ASSERTER = 'asserter';
-    const ACCESSOR_PUBLICPROPERTY = 'public';
-
     /**
      * Variables stored in context
      *
      * @var mixed
      */
     protected $variables = [];
+
+    /**
+     * Runtime cache to speed up object access through fluid
+     *
+     * @var array
+     */
+    protected $objectAccessorCache = [];
 
     /**
      * Variables, if any, with which to initialize this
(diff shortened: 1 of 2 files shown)


s2b:
I think that there could be more potential for optimization here, however I'm not sure if I understand the interconnection between the parsing stage and the rendering stage correctly (`ObjectAccessorNode::getAccessors()`). There seem to be cases where Fluid needs to re-check if the accessor actually is correct (see `extractSingleValue()`, and specifically `canExtractWithAccessor()`). Also, `extractWithAccessor()` re-checks some things that should already be clear.

lolli42:
Measured this with TYPO3 v12 backend Template module "Active TypoScript" and bootstrap_package TypoScript:
This is roughly 20ms seconds quicker, which is not much, extractSingleValue() still kills 1/4 of overall runtime.
I think we should see if we could further optimize this, probably by having the cache lookup in extractSingleValue() already, to skip both the extractWithAccessor() and the detectAccessor() call.
What do you think?

s2b:
After the other performance improvements in StandardVariableProvider I'm not sure if these changes are still worth it. The introduced complexity with the cache may outweigh the small remaining performance gains.

lolli42:
I did some further measurements: When we switch from call_user_func_array() to direct method calls, inline extract() into getByPath() and optimize a bit, we have improvements that are very close performance wise. Adding the cache afterwards then does not improve performance much anymore.

Instead, I think, we should try to pre-optimize the variable lookup in the compiled templates to reduce cycles in the method, next. For instance, we can explode a dotted string early and write the array into compiled templates directly, to then feed it to a method that takes the exploded array directly. Furthermore, we know up-front if the path-string contains sub references and can optimize that away as well.

As such, I'll close this PR now since the performance improvement of the lookup cache isn't high enough and we should better further optimize callers instead.

Issue #584, opened 2022-01-24
StandardVariableProvider type issue


ohader:
https://github.com/TYPO3/Fluid/blob/ae2d3d7ac201046275418c876fb82b63759678b1/src/Core/Variables/StandardVariableProvider.php#L261

`$this->getByPath` can be `null` leading to the following exception in TYPO3 v12 using PHP 8.1

> **#1476107295 TYPO3\CMS\Core\Error\Exception**
PHP Runtime Deprecation Notice: str_replace(): Passing null to parameter #2 ($replace) of type array|string is deprecated in vendor/typo3fluid/fluid/src/Core/Variables/StandardVariableProvider.php line 261

ohader:
Can be reproduced by using a sub-reference to a variable that is not defined in a Fluid template, e.g. `{{someUndefinedVariable}}`

helhum:
While at it, the following should be fixed as well:

https://github.com/TYPO3/Fluid/blob/ae2d3d7ac201046275418c876fb82b63759678b1/src/Core/Variables/StandardVariableProvider.php#L258

```diff
- preg_match_all('/(\{.*\})/', $propertyPath, $matches);
+ preg_match_all('/(\{[^}]*\})/', $propertyPath, $matches);
```

which fixes `{foo.{bar}.{baz}}`

mbrodala:
@helhum Can you create a separate ticket for this? Thanks.

stigfaerch:
> Can be reproduced by using a sub-reference to a variable that is not defined in a Fluid template, e.g. `{{someUndefinedVariable}}`

Any workarounds?
I have this:
{settings.maxImageWidth.colPos.{colPos}.full}

stigfaerch:
Btw - this is also the case for TYPO3 v11 and PHP 8.1

lolli42:
Fixed the deprecation warning now.
For the multi-sub-reference isssue Helmut mentioned: This is covered by #353 already, and we have https://github.com/TYPO3/Fluid/pull/663 and https://github.com/smoebody/Fluid/commit/51b0a926a7991c0353f3fc80df054bde10881c44
which both try to fix this. Let's continue with #353 for this.

PR #736, opened 2023-05-04
[BUGFIX] Avoid null to str_replace() in StandardVariableProvider

lolli42:
The getByPath() call in resolveSubVariableReferences() may return null if the variable can not be resolved. This leads to a PHP 8.1 deprecation when null is
given to str_replace().

Fix this and cover with tests.

Resolves: #584


Sebobo:
Sadly the type hint changes broke our FluidAdaptor for Neos.Flow https://github.com/neos/flow-development-collection/issues/3037 :/ 
These changes are really tricky.

PR #764, opened 2023-05-10
[TASK] Simplify arguments in compiled templates

lolli42:
NodeConverter->convertViewHelperNode() is
optimized to render more dense array definitions
of the argument array.


lolli42:
messed up ... followed in https://github.com/TYPO3/Fluid/pull/766

PR #787, opened 2023-05-13
[TASK] Move convert code from ViewHelperNode to VH

lolli42:
This aims to solve a class structure issue that
makes various things more complicated than they
should be, and prevents fixing various issues
with compiled templates.

When a template is parsed, view helpers are
represented by the ViewHelperNode in the
syntax tree.

At compile time, ViewHelperNode creates the PHP
representation of a view helper in convert().
convert() now creates the "outer" stuff - it
takes care of VH argument representation and
of embedding compiled children. For the "inner"
stuff, the specific ViewHelper compile()
method is called.

Various ViewHelpers thus override compile():
For instance, f:comment returns an empty string,
since it does not want anything to be rendered.

ViewHelperNode convert() however still compiles
children and adds them to the output, even though
this is not needed. This isn't great for f:comment, but it's a huge issue with f:if, which takes care
of children on it's own, but ViewHelperNode happily adds them a *second* time, which can be massive
with bigger f:then or f:else bodies.

There have been attempts to change this already:
The StopCompilingChildrenException can be thrown
in compile() in VH's, but then arguments
aren't compiled, either.

The true reason for these headaches is that the
"outer" stuff done in ViewHelperNode->convert()
does not belong there: Single VH's should
have full control over their compile representation, without ViewHelperNode->convert() assuming things.

The patch does exactly this: Move the convert()
body from ViewHelperNode to a new method in
AbstractViewHelper, which allows us to not only
solve the f:if / f:comment and friends issues
by overriding that method again, but also allows
us to phase out StopCompilingChildrenException
again.

Strictly speaking, this patch is breaking for two
reasons: First, we're changing ViewHelperInterface, and second, this fails when an existing VH currently has a method called "convert()" already.

We however consider this patch *not* being breaking *in practice* for these reasons, which is why we will add to next v2 minor:

First, we did not find a single Flow or TYPO3 related case where VH classes fully implement ViewHelperInterface on their own and do not extend AbstractViewHelper
directly or indirectly. We're now declaring
ViewHelperInterface @internal along the way and say that VH's *must* extend AbstractViewHelper. This should be ok and there are various internal ViewHelperInterface methods already.

Second, we could not find any public VH class with a method called "convert()" in neither Flow, nor
TYPO3 extensions. It seems this risk is very low.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -41,11 +41,6 @@ parameters:
 			path: src/Core/Parser/SyntaxTree/ViewHelperNode.php
 
 		-
-			message: "#^Else branch is unreachable because previous condition is always true\\.$#"
-			count: 1
-			path: src/Core/Parser/SyntaxTree/ViewHelperNode.php
-
-		-
 			message: "#^Binary operation \"\\+\" between non\\-empty\\-string and 0 results in an error\\.$#"
 			count: 1
 			path: src/Core/Parser/TemplateParser.php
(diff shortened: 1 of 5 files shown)


sbuerk (approved):
Strictly spoken, breaking. But commit message explains it very well. D`accord with beeing practical non-breaking and having it in a minor release.

PR #800, opened 2023-06-06
[TASK] Add expression node casting from and to json

kitzberger:
Handy shortcut to convert JSON without using any viewhelper (such as `f:format.json()` or `v:format.json.decode()`):

```
A json string converted to an array: {myJson as array}
An array converted to a json string: {myArray as json}
```

### DIFF
--- a/src/Core/Parser/SyntaxTree/Expression/CastingExpressionNode.php
+++ b/src/Core/Parser/SyntaxTree/Expression/CastingExpressionNode.php
@@ -22,7 +22,7 @@ class CastingExpressionNode extends AbstractExpressionNode
      * @var array
      */
     protected static $validTypes = [
-        'integer', 'boolean', 'string', 'float', 'array', 'DateTime'
+        'integer', 'boolean', 'string', 'float', 'array', 'DateTime', 'json'
     ];
 
     /**
@@ -88,6 +88,8 @@ class CastingExpressionNode extends AbstractExpressionNode
             $value = self::convertToDateTime($variable);
         } elseif ($type === 'array') {
             $value = (array)self::convertToArray($variable);
+        } elseif ($type === 'json') {
+            $value = (string)self::convertToJson($variable);
         }
         return $value;
     }
@@ -113,8 +115,14 @@ class CastingExpressionNode extends AbstractExpressionNode
         if (is_array($variable)) {
             return $variable;
         }
-        if (is_string($variable) && strpos($variable, ',')) {
-            return array_map('trim', explode(',', $variable));
+        if (is_string($variable)) {
+            // rough check on whether or not it's a json string
+            if (preg_match('/^[\[\{].*[\]\}]$/', trim($variable))) {
+                return json_decode($variable, true, 512, JSON_THROW_ON_ERROR);
+            }
+            if (strpos($variable, ',')) {
+                return array_map('trim', explode(',', $variable));
+            }
         }
         if ($variable instanceof \Iterator) {
             $array = [];
@@ -130,5 +138,14 @@ class CastingExpressionNode extends AbstractExpressionNode
             return [];
         }
         return [$variable];
+    }
+
+    /**
+     * @param mixed $variable
+     * @return string
+     */
+    protected static function convertToJson($variable)
+    {
+        return json_encode($variable, JSON_THROW_ON_ERROR);
     }
 }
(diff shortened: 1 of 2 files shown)


src/Core/Parser/SyntaxTree/Expression/CastingExpressionNode.php
```diff
@@ -113,8 +115,16 @@ protected static function convertToArray($variable)
         if (is_array($variable)) {
             return $variable;
         }
-        if (is_string($variable) && strpos($variable, ',')) {
-            return array_map('trim', explode(',', $variable));
+        if (is_string($variable)) {
+            if (preg_match('/^[\[\{].*[\]\}]$/', $variable)) {
```
mbrodala, line 119: I fear this check must be dropped since simple literals like `'null'`, `1` are also valid JSON, not only lists and objects/maps.

mbrodala: Geh, the point here is conversion to array, so it's perfectly fine like this. :facepalm:

sbuerk, line 119: I guess a `trim()` should be used here. The regexp would not match if there is one or more spaces at the beginning or the end. `json_decode()` properly handles this.

```
$myjsonstring = ' {"k1": "v1"} ';
```

would not be matched. See: https://3v4l.org/XWMi2

sbuerk (changes requested):

src/Core/Parser/SyntaxTree/Expression/CastingExpressionNode.php
sbuerk: Thanks. You should also fix the cgl issues, see PHP 8.1 checks.

kitzberger: Just done that ;-)

sbuerk: Don't want to bother you, but maybe we can add at least one test for the space/trim addition ?

kitzberger: Sure, done.

src/Core/Parser/SyntaxTree/Expression/CastingExpressionNode.php
```diff
@@ -113,8 +115,17 @@ protected static function convertToArray($variable)
         if (is_array($variable)) {
             return $variable;
         }
-        if (is_string($variable) && strpos($variable, ',')) {
-            return array_map('trim', explode(',', $variable));
+        if (is_string($variable)) {
+            // rough check on whether or not it's a json string
+            if (preg_match('/^[\[\{].*[\]\}]$/', trim($variable))) {
+                $array = @json_decode($variable, true);
```
s2b, line 121: Any particular reason why you use `@` here? If you wanted to cover this:

> If depth is outside the allowed range, a [ValueError](https://www.php.net/manual/en/class.valueerror.php) is thrown as of PHP 8.0.0, while previously, an error of level E_WARNING was raised.

I would rather add a try/catch here since Fluid requires PHP 8.1 now. This makes it more explicit. But I think an exception would also be fine in this case.

s2b: Also, do you have any particular reason why you use `JSON_THROW_ON_ERROR` for encode, but not for decode?

s2b (changes requested):
First of all, I think this is a great addition. :) I just have a few more questions/suggestions.

src/Core/Parser/SyntaxTree/Expression/CastingExpressionNode.php
kitzberger: @s2b, no reason, just thought it might be nice to not let the frontend explode when there's an issue with the provided string. But you're right I guess ;-)

I've changed that and added a test for an invalid JSON too.

s2b (approved):

NamelessCoder:
Problem 1: JSON encoding options cannot be assumed in a way that would work for every possible use case. At  best, a "one size kind of fits most" can be achieved.
Problem 2: "json" isn't a valid PHP type and the CastingExpressionNode is intended to deal explicitly with casting to PHP types internally.

Suggestion: this shouldn't be the responsibility of the casting expression node. Rather, it should be a dedicated ViewHelper which allows passing 1) options for various encoding/escaping parameters for `json_encode` and 2) option to declare if the output of a decode should be an object or an associative array.

PR #805, opened 2023-06-16
[BUGFIX] Ensure path order with duplicates

s2b:


sbuerk:
@s2b Can you please:

* Add some more description to the pull-request
* Add some commit message describing it
* or at least add and reference an issue

There is nothing which can be read why what is expected
and why this should be considered as bug. Especially why
the chosen solution is the **correct** way o fix it. Or
better to say why the behaviour should be the expected
behaviour.

Just from the code, I'm not quite sure if everyone would
follow what you postulated by code here and why it is fixed
the way you do it.

Not sure if this this is really a bug or more a feature.

It's debatable if a later re-announcing a already added
path should change the order. It may be your use-case,
but is it really a valid expectation in all cases ?

So let's add some public information to it and not doing
a silenced and hidden slip-in fix.

s2b:
Sorry, this was meant as a draft to check if this fixes https://forge.typo3.org/issues/81099. I‘ll add more information later.

s2b:
This has been fixed in TYPO3 now, so this is no longer relevant

PR #806, opened 2023-06-18
[BUGFIX] Ensure correct `ViewHelperNode` in `ViewHelper`

### LINKED ISSUE #804: Fluid 2.9.x has trouble with StaticCacheable ViewHelpers and cached partials/templates
When working with EXT:solr (dev-main) we've noticed a strange behaviour which I want to demonstrate quickly.

I have
* A template file (Results.html)
* Calling a partial file (Pagination.html)
* Using a simple ViewHelper (static cacheable) that just returns the given argument "page".

When calling the page with the template + partial for the first time, everything is fine, the second time (when Partials and Templates are cached), the ViewHelper always outputs the same value.

This does not happen in 2.8.x but only in 2.9.x.

Results.html:
```
<html xmlns="http://www.w3.org/1999/xhtml" lang="en"
	xmlns:f="http://typo3.org/ns/TYPO3/Fluid/ViewHelpers"
	xmlns:s="http://typo3.org/ns/ApacheSolrForTypo3/Solr/ViewHelpers"
	data-namespace-typo3-fluid="true"
>

<f:render partial="Result/Pagination" arguments="{allPageNumbers: '{1:1, 2:2, 3:3, 4:4, 5:5}', previousPageNumber: 1, nextPageNumber: 3}"/>

</html>
```

Pagination.html:
```
<html xmlns="http://www.w3.org/1999/xhtml" lang="en"
	xmlns:f="http://typo3.org/ns/TYPO3/Fluid/ViewHelpers"
	xmlns:s="http://typo3.org/ns/ApacheSolrForTypo3/Solr/ViewHelpers"
	data-namespace-typo3-fluid="true"
>

<ul class="pagination-test">
	<li class="page-item">{s:compilable(page: previousPageNumber)}- {previousPageNumber}</li>

	<f:for each="{allPageNumbers}" as="page">
		<li class="page-item">{s:compilable(page: '{page}')} - {page}</li>
	</f:for>

	<li class="page-item">{s:compilable(page: '{nextPageNumber}')} - {nextPageNumber}</li>
</ul>

</html>
```

CompilableViewHelper.php:
```
<?php
declare(strict_types=1);

namespace ApacheSolrForTypo3\Solr\ViewHelpers;

use TYPO3Fluid\Fluid\Core\Rendering\RenderingContextInterface;
use TYPO3Fluid\Fluid\Core\ViewHelper\Traits\CompileWithRenderStatic;
use TYPO3Fluid\Fluid\Core\ViewHelper\AbstractViewHelper;

class CompilableViewHelper extends AbstractViewHelper
{
    use CompileWithRenderStatic;

    /**
     * @inheritdoc
     */
    public function initializeArguments(): void
    {
        parent::initializeArguments();
        $this->registerArgument('page', 'int', 'The page', false);
    }

    public static function renderStatic(
        array $arguments,
        \Closure $renderChildrenClosure,
        RenderingContextInterface $renderingContext,
    ) {
        $page = $arguments['page'] ?? 1;
        return (string)$page;
    }
}
```

The output on the first run:
```
<ul class="pagination-test">
<li class="page-item">1- 1</li>
	<li class="page-item">1 - 1</li>
	<li class="page-item">2 - 2</li>
	<li class="page-item">3 - 3</li>
	<li class="page-item">4 - 4</li>
	<li class="page-item">5 - 5</li>
<li class="page-item">3 - 3</li>
</ul>
```

The output on all subsequent runs:
```
<ul class="pagination-test">
<li class="page-item">3- 1</li>
	<li class="page-item">3 - 1</li>
	<li class="page-item">3 - 2</li>
	<li class="page-item">3 - 3</li>
	<li class="page-item">3 - 4</li>
	<li class="page-item">3 - 5</li>
<li class="page-item">3 - 3</li>
</ul>
```

sbuerk:
Fluid have been designed without DependencyInjection
and therefore shared classes, for example ViewHelper,
in mind.

Using systems can implement custom ViewHelperResolvers
to determine how ViewHelper are resolved and created,
using custom configuration. For example, the TYPO3 CMS
implements such a custom ViewHelperResolver providing
the ability to instanciate them throug Symfony DI.

This allows shared ViewHelper instances now and may
become more likely, especial for custom ViewHelper
using the `CompileWithRenderStatic` trait.

ViewHelper and ViewHelperNodes are connected through
cycling references. This has not been a problem until
the execution code for `convert()` have been moved
from ViewHelperNode to the AbstractViewHelper (#787)
to avoid creating unnessesary children codes, which may
be duplicated and not needed. However, that enforced
the code to work with the `ViewHelperNode` set in the
`ViewHelper`.

Using DI and shared `ViewHelper` now leads to the state,
that the wrong `ViewHelperNode` instance is referenced
in the shared `ViewHelper`. That compiled template code
retrieving the wrong argument in some circumstances, for
example in a for-loop when the viewhelper are additionaly
used after the for-loop.

Moving the `convert()` code from the node to the ViewHelper
is valid, but `shared ViewHelper instances` are valid too.

This change now updates the ViewHelperNode reference in
the ViewHelper before dispatching to the ViewHelper.
We do not have any evidence yet, but the same thing is
done in the `ViewHelperNode->evaluate()` method.

This is baked by a regression test. A cross check is
added, but skipped for now as it makes the test green
even without the fix. That points to another issue,
test setup related or other, and have to be tackled
in a dedicated change. The skipped test is added to
the phpstan baseline.

Used command(s):

```shell
composer phpstan:generate-baseline
```

Resolves #804
Related: #787 https://github.com/TYPO3/Fluid/commit/1f1dc5ac2d23aff659a8b305f0ac765d15b7fba7

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -114,3 +114,8 @@ parameters:
 			message: "#^Call to static method PHPUnit\\\\Framework\\\\Assert\\:\\:assertSame\\(\\) with TYPO3Fluid\\\\Fluid\\\\Tests\\\\Functional\\\\Fixtures\\\\Various\\\\UserWithoutToString and string will always evaluate to false\\.$#"
 			count: 1
 			path: tests/Functional/ViewHelpers/Format/HtmlspecialcharsViewHelperTest.php
+
+		-
+			message: "#^Unreachable statement \\- code above always terminates\\.$#"
+			count: 1
+			path: tests/Functional/ViewHelpers/StaticCacheable/NotSharedStaticCompilableViewHelperTest.php
--- a/src/Core/Parser/SyntaxTree/ViewHelperNode.php
+++ b/src/Core/Parser/SyntaxTree/ViewHelperNode.php
@@ -122,18 +122,42 @@ class ViewHelperNode extends AbstractNode
      * If the view helper implements \TYPO3Fluid\Fluid\Core\ViewHelper\ChildNodeAccessInterface,
      * it calls setChildNodes(array childNodes) on the view helper.
      *
-     * Afterwards, checks that the view helper did not leave a variable lying around.
+     * Afterward, checks that the view helper did not leave a variable lying around.
      *
      * @param RenderingContextInterface $renderingContext
      * @return string evaluated node after the view helper has been called.
      */
     public function evaluate(RenderingContextInterface $renderingContext)
     {
+        // This is added as a safe-off, currently no evidence that we need this here like in convert().
+        // See: https://github.com/TYPO3/Fluid/issues/804
+        $this->updateViewHelperNodeInViewHelper();
         return $renderingContext->getViewHelperInvoker()->invoke($this->uninitializedViewHelper, $this->arguments, $renderingContext);
     }
 
     public function convert(TemplateCompiler $templateCompiler): array
     {
+        // We need this here to avoid https://github.com/TYPO3/Fluid/issues/804.
+        $this->updateViewHelperNodeInViewHelper();
         return $this->uninitializedViewHelper->convert($templateCompiler);
+    }
+
+    /**
+     * Ensure correct ViewHelperNode (this) reference in the uninitialized ViewHelper instance.
+     */
+    protected function updateViewHelperNodeInViewHelper(): void
+    {
+        // Custom ViewHelperResolver can and are implemented providing the ability to instantiate ViewHelpers through
+        // a DependencyInjection system like Symfony DI, for example done by TYPO3. Due to the nature, instances may be
+        // set as shared, which means that changes to property reflects the latest set state. Therefore, we need to set
+        // the current ViewHelperNode to a viewhelper instance to ensure correct context.
+        // See https://github.com/TYPO3/Fluid/issues/804
+        // @todo We should evaluate if we can get rid of this state and better pass it around.
+        // @todo The ViewHelperInterface does not contain the setViewHelperNode() method. Most likely ViewHelper are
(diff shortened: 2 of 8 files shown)


s2b (approved):

bmack (commented):
works for me!

NamelessCoder:
Guys. Here's the thing:

ViewHelpers are by design created to *not* be shared instances. The problem isn't that Fluid doesn't handle a shared instance (it was never supposed to, and this is something that TYPO3 specifically is trying to force it to do via a ViewHelperResolver override). The problem is that you/users are declaring every ViewHelper to be a **shared** service. Which is why you now have to jump through hoops and do various workarounds in Fluid, adding new mutability and relying on transient states.

The fact that DI is used does not also imply that instances must be shared. Instances are, however, shared by default unless you configured it otherwise (and one might argue that 1) either this is a bad design choice, or 2) making ViewHelpers behave like DI candidates is a bad design choice - I lean towards the latter).

> but `shared ViewHelper instances` are valid too.

This is the assertion that is incorrect. Shared instances are not valid. Instances are, by design, supposed to be unique per parsed node.

In short: the problem isn't Fluid. The problem is the attempt to make ViewHelper instances shared. The right solution is to declare ViewHelpers specifically `shared: false` in DI configuration or add a compiler pass that scans for the interface and forces `shared: false` unless explicitly told not to.

FYI: this is likely to cause other problems down the road, for example when nesting the same ViewHelper within itself.

If I were you I would rethink these decisions and come up with a solution within context of DI to prevent shared VH instances.

PR #811, opened 2023-07-18
[TASK] Raise phpunit and align changes

sbuerk:
PHPUnit removed the `-v` cli argument
with 10.2.4 as a followup. That flag
was the short form for the --verbose
removed with 10.0 and did not had any
effect.

Due to the fact that no composer.lock
file is used, the newer version is
installed leading to failing tests.

This change raises the version of
PHPUnit and aligns the test calls
to avoid the removed cli argument.

Used command(s):

```shell
composer require --dev --no-update \
    "phpunit/phpunit":"^10.2.6"
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -44,4 +44,4 @@ jobs:
         run: vendor/bin/phpstan analyze --no-progress
 
       - name: Phpunit v10
-        run: vendor/bin/phpunit -v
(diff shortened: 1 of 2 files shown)


sbuerk:
https://github.com/sebastianbergmann/phpunit/blob/10.2/ChangeLog-10.2.md#1024---2023-07-10

PR #813, opened 2023-09-26
[BUGFIX] Fix missing “null” annotation for parameter $sectionName

magicsunday:
Correct $sectionName's annotation that this can be null. Add missing default values to method parameters.


mbrodala:
Can you explain in more detail why this change is necessary?

Currently the only call to this method in Fluid itself is in the `RenderViewHelper` which does provide all necessary arguments:

https://github.com/TYPO3/Fluid/blob/64f4e46fab194b180b4f5db4ca071eb3e01ae528/src/ViewHelpers/RenderViewHelper.php#L164

magicsunday:
Because, for example, I create an instance of StandaloneView in middleware to render a partial. I don't need a section there. Phpstan then issues an error here (Parameter #2 $sectionName of method TYPO3Fluid\Fluid\View\AbstractTemplateView::renderPartial() expects string, null given.). Furthermore, the annotation in @param is simply wrong because the method checks for !== null, but the type is "string". And if $sectionName is checked for null, so the value can be null, then it can also be defined and initialized as a default value in the method. Everything else is unclean in my opinion.

mbrodala:
Thanks for the explanation, makes sense. And `$variables` is either passed to `renderSection()` where it is optional or to `VariableProviderInterface::getScopeCopy()` which enforces a constructor where `$variables` are optional.

So all in all :+1:

mbrodala (approved):

s2b:
LGTM, can be merged once #814 is merged and the branch has been rebased.

s2b:
Can you also adjust your commit message to use one of the prefixes described here:
https://docs.typo3.org/m/typo3/guide-contributionworkflow/main/en-us/Appendix/CommitMessage.html#summary-line-first-line

and a shorter title with an optional longer description if necessary.

s2b:
Can you fix the CGL issue in your changes?

s2b:
Just did some CGL changes, nothing substantial, so I'm merging without additional review.

kitsunet:
This change breaks API as the signature of  `AbstractTemplateView::renderPartial` is no longer in line with the Interface which does not declare those arguments optional, this minor release breaks for Neos. :(

s2b:
@kitsunet Ugh, sorry about that. Will take a look at it tomorrow!

You're talking about https://github.com/neos/fluidadaptor, right?

kitsunet:
In this case https://github.com/neos/form/blob/master/Classes/Core/Renderer/FluidFormRenderer.php#L203 
The fix is easy, will do it tomorrow, was just unexpected to see CI breaking

lolli42:
So neos/form adapted its implementation. There is nothing more to do for us, here, right?

PR #814, opened 2023-09-26
[TASK] Apply single_line_empty_body from PER-CS2.0

s2b:
PER Coding Style 2.0, which fluid respects with its php-cs-fixer configuration, added a CGL rule to define empty classes, interfaces and functions in one line rather than each brace on its separate line. php-cs-fixer already enforces this, so this commit adjusts the existing code base.

Executed command:

composer cgl:fix


s2b:
I'm currently coordinating this change with the TYPO3 Core, stay tuned.

Issue #666, opened 2022-10-06
AliasViewHelper has side effects


cundd:
I just recognized that `<f:alias>` has side effects if one of the alias-names matches an outer variable name.

Example:

```php
<?php

$this->view->assign('seminars', [
    'open'   => ['Seminar 1', 'Seminar 2'],
    'closed' => ['Seminar 3', 'Seminar 4'],
]);
```

```html
<!-- `seminars` contains the nested array -->
<f:alias map="{seminars : seminars.open}">
    <f:render section="list" arguments="{_all}"/>
</f:alias>

<!-- `seminars` is `NULL` because of https://github.com/TYPO3/Fluid/blob/main/src/ViewHelpers/AliasViewHelper.php#L87 -->
<f:alias map="{seminars : seminars.closed}">
    <f:render section="list" arguments="{_all}"/>
</f:alias>
```

The reason is, that the original values are not preserved in [AliasViewHelper.php](https://github.com/TYPO3/Fluid/blob/main/src/ViewHelpers/AliasViewHelper.php) but any value with the mapped key will simply be dropped.

Would it be a viable solution to clone the variable provider in https://github.com/TYPO3/Fluid/blob/main/src/ViewHelpers/AliasViewHelper.php#L80 and reuse the original one after `$renderChildrenClosure()`?

cundd:
The same is true for the `<f:for>` View Helper.

masi:
Odd. I thought the whole idea was to encapsulate the assignments.

PR #819, opened 2023-10-16
[TASK] Demonstration of Fluid bug with render()

s2b:
This PR aims to demonstrate a bug in Fluid with ViewHelpers that use `render()` instead of `renderStatic()`.

The API for ViewHelpers without a static implementation states that `$this->renderChildren()` should be used to render the child nodes of the ViewHelper. However, with the adjusted `ForViewHelper`, this doesn't work as expected.

I can get it to work if I reproduce the context of a static ViewHelper:

```php
        $renderChildrenClosure = $this->buildRenderChildrenClosure();
        foreach ($this->arguments['each'] as $keyValue => $singleElement) {
            // ...
            $output .= $renderChildrenClosure();
```

Note that this issue only affects uncached templates, which suggests to me that this is unwanted behavior.

### DIFF
--- a/src/ViewHelpers/ForViewHelper.php
+++ b/src/ViewHelpers/ForViewHelper.php
@@ -7,10 +7,8 @@
 
 namespace TYPO3Fluid\Fluid\ViewHelpers;
 
-use TYPO3Fluid\Fluid\Core\Rendering\RenderingContextInterface;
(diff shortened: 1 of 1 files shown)


s2b:
This seems to be a special variant of #666, and #820 seems to solve both problems.

PR #820, opened 2023-10-16
[BUGFIX] Eliminate side-effects from ForViewHelper

s2b:
The previous implementation overwrote template variables with the same name as the specified ViewHelper parameters "as", "key", or "iteration". With the new implementation, the variable state before the ViewHelper call gets restored once the ViewHelper call is completed.

### DIFF
--- a/src/ViewHelpers/ForViewHelper.php
+++ b/src/ViewHelpers/ForViewHelper.php
@@ -102,7 +102,6 @@ class ForViewHelper extends AbstractViewHelper
      */
     public static function renderStatic(array $arguments, \Closure $renderChildrenClosure, RenderingContextInterface $renderingContext)
     {
-        $templateVariableContainer = $renderingContext->getVariableProvider();
         if (!isset($arguments['each'])) {
             return '';
         }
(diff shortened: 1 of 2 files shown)


s2b:
I think we might need to consider that this is used as a feature in some projects, such as:

```xml
<f:for each="{items}" as="item">
  do something
</f:for>
Last item: {item}
```

Not saying that this is good or recommended, just something to think about.

s2b:
Just noticed that this would break the following hack/work-around:

```xml
<f:for each="{items}" as="item">
  <f:variable name="lastItem" value="{item}" />
</f:for>
```

Maybe we need some kind of merge logic here?

s2b:
The new version of this patch should mitigate this.

s2b:
I think that there is still an edge case here:
```xml
<f:variable name="lastName" value="fallback" />
<f:for each="{items}" as="item">
  <f:variable name="lastItem" value="{item}" />
</f:for>
```
In this case, `lastName` would always be `fallback` because the variable gets restored to its initial state after the loop.

lolli42 (approved):

Issue #826, opened 2023-10-31
[BUG] Array to string conversion warning in AbstractNode


sbuerk:
Note: Created based  on https://forge.typo3.org/issues/102264

Hi,
Exception throws at: /var/www/html/vendor/typo3fluid/fluid/src/Core/Parser/SyntaxTree/AbstractNode.php line 73
That means:

    {
        if (is_object($value) && !method_exists($value, '__toString')) {
            throw new Parser\Exception('Cannot cast object of type "' . get_class($value) . '" to string.', 1273753083);
        }
        $output = (string)$value;
        return $output;
    }

The line $output = (string)$value; throws it, but the error in mho is in the preceeding if statement: It's asking for object, but not for an array. (Which is the case there).

![Screenshot 2023-10-26 at 14-05-20 TYPO3 Exception](https://github.com/TYPO3/Fluid/assets/1453466/a91e086e-650a-4958-a7b7-f7033abca53f)

subnoodle:
Stefan, that is AWESOME. Thank you!

PR #827, opened 2023-10-31
[BUGFIX] Throw array conversion exception in `AbstractNode->castToString()`

sbuerk:
`AbstractNode->castToString()` already checked for objects which could not
converted to a string representation due to missing `__toString()` method.
It's possible that the value could be an array, which per nature is not
really convertable to a string anyway. Older php versions simply converted
it to an `Array` string along with a notices, which has been raised to a
warning in newer PHP versions.

There is no way to properly convert an array, therefore the PHP warning
should always lead to an parser exception like for object not convertable.

This change now checks explicitly for an array and throws a corresponding
`TYPO3Fluid\Fluid\Core\Parser\Exception` with a meaningfull message.

Test is extended to cover this case.

Resolves: #826


s2b (approved):

s2b:
I agree that a proper exception should be thrown for this. However, this might break existing templates where PHP warnings are suppressed. Not sure if this is a problem?

PR #828, opened 2023-11-20
[FEATURE] Proper support of local variables in ViewHelpers

### LINKED ISSUE #666: AliasViewHelper has side effects

s2b:
In its current implementation, the `f:alias` ViewHelper has side effects: It adds variables to the global variable container, which are not restored properly after the ViewHelper call. If a variable name is used as alias which already exists, that variable is unset after the ViewHelper call.

To resolve this, a new `ScopedVariableProvider` is introduced to handle local variables
inside ViewHelper calls properly. After the local scope, global variables are restored to their original value in case they existed. At the same time, other modifications to the variable provider inside the `f:alias` call, such as a call to `f:variable`, should still be persistent after the ViewHelper call. This ensures that the change is non-breaking for the vast majority of usages. The only situation where this could be breaking is when you relied on the arguably broken behavior that `f:alias` unsets the variables afterwards.

This is also the reason why we can't simple store and reset the whole variable container. This would be a breaking change and would also be a diversion from other ViewHelpers, such as `f:if`, where `f:variable` also leaks out. To change this behavior for select ViewHelpers would be inconsistent and hard to understand for users of Fluid.

before:

```xml
<f:variable name="myVariable" value="initialValue" />
<f:alias map="{myVariable: \'myValue\'}">...</f:alias>
{myVariable} <!-- outputs empty string -->
```

after:

```xml
<f:variable name="myVariable" value="initialValue" />
<f:alias map="{myVariable: \'myValue\'}">...</f:alias>
{myVariable} <!-- outputs "initialValue" -->
```

This behavior also applies to other ViewHelpers that provide variables to their child nodes, such as `f:for`, `f:render`, `f:groupedFor`, `f:cycle`. These will be modified accordingly in separate PRs.

Resolves #666

### DIFF
--- a/src/ViewHelpers/AliasViewHelper.php
+++ b/src/ViewHelpers/AliasViewHelper.php
@@ -8,6 +8,8 @@
 namespace TYPO3Fluid\Fluid\ViewHelpers;
 
 use TYPO3Fluid\Fluid\Core\Rendering\RenderingContextInterface;
+use TYPO3Fluid\Fluid\Core\Variables\ScopedVariableProvider;
+use TYPO3Fluid\Fluid\Core\Variables\StandardVariableProvider;
 use TYPO3Fluid\Fluid\Core\ViewHelper\AbstractViewHelper;
 use TYPO3Fluid\Fluid\Core\ViewHelper\Traits\CompileWithRenderStatic;
 
@@ -77,15 +79,15 @@ class AliasViewHelper extends AbstractViewHelper
      */
     public static function renderStatic(array $arguments, \Closure $renderChildrenClosure, RenderingContextInterface $renderingContext)
     {
-        $templateVariableContainer = $renderingContext->getVariableProvider();
-        $map = $arguments['map'];
-        foreach ($map as $aliasName => $value) {
-            $templateVariableContainer->add($aliasName, $value);
-        }
+        $globalVariableProvider = $renderingContext->getVariableProvider();
+        $localVariableProvider = new StandardVariableProvider($arguments['map']);
+        $scopedVariableProvider = new ScopedVariableProvider($globalVariableProvider, $localVariableProvider);
+        $renderingContext->setVariableProvider($scopedVariableProvider);
+
         $output = $renderChildrenClosure();
-        foreach ($map as $aliasName => $value) {
-            $templateVariableContainer->remove($aliasName);
-        }
+
+        $renderingContext->setVariableProvider($globalVariableProvider);
+
         return $output;
     }
 }
--- a/Documentation/Usage/Syntax.rst
+++ b/Documentation/Usage/Syntax.rst
@@ -151,3 +151,52 @@ types. To be able to cast a variable in this case, simply wrap it with quotes:
 cast the variable and finally remove the quotations and use the variable
 directly. Semantically, the quotes mean you create a new `TextNode` that
 contains a variable converted to the specified type.
+
+Variable Scopes
+===============
+
+Each Fluid template, partial and section has its own variable scope. For templates,
+these variables are set via the PHP API, for partials and sections the `<f:render>`
+ViewHelper has a `arguments` argument to provide variables.
+
+Inside templates, partials and sections there are two variable scopes: global
+variables and local variables. Local variables are created by ViewHelpers that
+provide additional variables to their child nodes. Local variables are only valid
+in their appropriate context and don't leak out to the whole template.
+
+For example, `<f:alias>` and `<f:for>` create local variables:
+
+.. code-block:: xml
+
+    <f:for each="{items}" as="item">
+        <!-- {item} is valid here -->
+    </f:for>
+    <!-- {item} is no longer valid here -->
+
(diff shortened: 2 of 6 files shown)


src/ViewHelpers/AliasViewHelper.php
```diff
@@ -78,14 +78,23 @@ public function initializeArguments()
     public static function renderStatic(array $arguments, \Closure $renderChildrenClosure, RenderingContextInterface $renderingContext)
     {
         $templateVariableContainer = $renderingContext->getVariableProvider();
+        $originalVariableContainer = clone $templateVariableContainer;
```
mbrodala, line 81: How about `$renderingContext->setVariableProvider($originalVariableContainer)` afterwards instead?

s2b: This is what the additional tests I just added cover: If you use `f:variable` inside `f:alias`, those variables imho shouldn't be reset (and this would be breaking).

mbrodala: This is getting ridiculously complicated though:

```html
<f:variable name="foo" value="1"/>
<f:alias map="{foo: 2}">
    <f:variable name="foo" value="3"/>
</f:alias>
```
What should be the value of `foo` after this?

s2b: In my opinion, the rule should be: ViewHelpers injecting variables into their children (like `f:alias` and `f:for`) create local variables, all others are global.

I recognize that this gets complicated, but `f:for` and `f:variable` are out there... Do you think that the current alternative that creates side effects is better/easier?

mbrodala: I'm thinking about how to actually solve this. The example I just gave is hard to implement as expected. The `f:alias` viewhelper can only restore `foo` to the state it had before it was rendered. This way it overwrites the value set by the 2nd `f:variable` viewhelper.

So a possible workflow:

1. Back up variables before rendering
2. Render
3. Go through all `map` keys
4. If a `map` value is different from the one in the `map`, keep it
5. If a `map` value is the same as the one in the `map` and a matching variable is present in the backup, restore the backup value
6. Otherwise remove the variable again

s2b: If I understood this correctly, this would lead to the following rule:

`f:variable` (and other potential ViewHelpers that create/modify variables directly, something like `f:array.push`) always create global variables.

mbrodala: I'd avoid the term "global variables" until Fluid actually has variable scopes. But yeah, unless variables are created by a viewhelper for child rendering, they will stay around after rendering.

s2b: Maybe the implementation would be clearer if there would be a special VariableProvider implementation for these circumstances?

It could encapsulate two StandardVariableProviders ("global/outside" and "local/inside"). `get()` would check local variables first and fall back to global variables. `add()` would always write to global variables. Before `renderChildren()`, the new VariableProvider would be constructed, afterwards it could just be reset to the global VariableProvider.

mbrodala: That's basically implementing variable scopes in Fluid. ;-)

Small correction: `add()` would always write to local variables here, that's what `f:alias` is expected to do. Children rendered within would access the "global" provider instead.

mbrodala:
Only a few words: bam! :boom:

src/ViewHelpers/AliasViewHelper.php
s2b: I've created a separate VariableProvider that can be used by all ViewHelpers that create local variables. This PR now depends on a pre-patch #829 that removes the constructor from the underlying interface.

lolli42 (approved):

mbrodala:
Thanks **a lot** for this missing piece!

PR #832, opened 2023-11-22
[FEATURE] Attributes for RenderingContext

s2b:
This adds the ability to attach arbitrary attributes to the rendering context. All popular usages of Fluid out there extend the rendering context implementation of Fluid and don't actually implement the whole interface, so we should be able to change the interface as well.

As a result of this, TYPO3 CMS will be able to drop its own rendering context implementation, which will make the integration more straightforward.

### DIFF
--- a/src/Core/Rendering/RenderingContext.php
+++ b/src/Core/Rendering/RenderingContext.php
@@ -111,6 +111,12 @@ class RenderingContext implements RenderingContextInterface
     ];
 
     /**
+     * Attributes can be used to attach additional data to the
+     * rendering context to be used e. g. in ViewHelpers.
+     */
+    protected array $attributes = [];
+
+    /**
      * Constructor
      *
      * Constructing a RenderingContext should result in an object containing instances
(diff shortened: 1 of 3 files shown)


mbrodala:
Hm, we should reconsider the name "attribute" here.

Given that Fluid is often used for HTML, we have a confusion with HTML attributes.
Given that Fluid is written in PHP, we have a confusion with PHP attributes.

So maybe simply `data` or `properties` would be better.

s2b:
I understand the naming conflicts, but I'm not sure if data or properties are a better fit. `withData()` has the singular/plural issue, `withProperty()` sounds to me like simple values, not complex objects (like a request object). `withAttribute()` is already known from PSR-7.

https://www.php-fig.org/psr/psr-7/#321-psrhttpmessageserverrequestinterface

mbrodala:
Hm, indeed. The TYPO3 core context uses the term "aspect" but that's maybe a different scope. Given these circumstances, "attribute" may indeed be a good choice.

lolli42 (approved):

PR #837, opened 2023-11-23
[FEATURE] TrimViewHelper for StandaloneFluid

s2b:
TrimViewHelper doesn't have any dependencies to TYPO3, so it can be moved to Fluid Standalone.

### DIFF
--- a/src/ViewHelpers/Format/TrimViewHelper.php
+++ b/src/ViewHelpers/Format/TrimViewHelper.php
@@ -1,0 +1,126 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file belongs to the package "TYPO3 Fluid".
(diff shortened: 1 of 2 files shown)


src/ViewHelpers/Format/TrimViewHelper.php
```diff
@@ -0,0 +1,126 @@
...
+ * Trim special characters
+ * -----------------------
+ *
+ * ::
+ *
+ *    #<f:format.trim characters=" St.">   String to be trimmed.   </f:format.trim>#
+ *
+ * .. code-block:: text
+ *
+ *    #ring to be trimmed#
```
lolli42, line 67: ring ring ^^

lolli42 (approved):

PR #841, opened 2023-11-24
[FEATURE] CaseViewHelper for StandaloneFluid

s2b:
CaseViewHelper doesn't have any dependencies to TYPO3, so it can be moved to Fluid Standalone.

### DIFF
--- a/src/ViewHelpers/Format/CaseViewHelper.php
+++ b/src/ViewHelpers/Format/CaseViewHelper.php
@@ -1,0 +1,151 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file belongs to the package "TYPO3 Fluid".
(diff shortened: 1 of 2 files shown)


src/ViewHelpers/Format/CaseViewHelper.php
```diff
@@ -0,0 +1,151 @@
...
+ *    SomeString
+ */
+final class CaseViewHelper extends AbstractViewHelper
+{
+    use CompileWithRenderStatic;
+
+    /**
+     * Directs the input string being converted to "lowercase"
+     */
+    public const CASE_LOWER = 'lower';
```
lolli42, line 75: all those can be at least protected, can they?!

lolli42 (approved):

PR #845, opened 2023-11-28
Do not check for content to be null in OrViewHelper.php

Alagts:
Defining an alternative and make a field (content) empty has not the expected result (we expect the alternative to be displayed). I found a condition that only checks if content is equal to null. But I never reach this, rather I reach an empty string.
The tests also have to be adapted. The TODO hint is correct.

### DIFF
--- a/src/ViewHelpers/OrViewHelper.php
+++ b/src/ViewHelpers/OrViewHelper.php
@@ -65,7 +65,7 @@ class OrViewHelper extends AbstractViewHelper
 
         $content = $renderChildrenClosure();
 
-        if (null === $content) {
+        if ('' === $content) {
             $content = $alternative;
         }
 


src/ViewHelpers/OrViewHelper.php
```diff
@@ -65,7 +65,7 @@ public static function renderStatic(array $arguments, \Closure $renderChildrenCl
 
         $content = $renderChildrenClosure();
 
-        if (null === $content) {
+        if ('' === $content) {
```
mbrodala, line 68: Now if `$content` really is `null`, the viewhelper would not work as expected anymore.

Thus one would actually also check against `null` here. Notice that `empty()` should not be used since the viewhelper would also render the alternative on `'0'` as `$content` then.

In any case I expect this to be breaking due to a change of behavior (even if desired). So this is only an option for the next major version.

NamelessCoder:
This ViewHelper is explicitly designed to only render output if the content is exactly `null` (implied: variable is undefined). It should not be comparing to anything other than explicitly `null`; that's what the `f:if` ViewHelper is for.

s2b:
Since the documentation of the VH is pretty self-explanatory that it covers only the `null` use case, I don't think that it makes sense to change that behavior.

PR #850, opened 2023-12-09
[BUGFIX] Choose global/local variable based on primary identifier

### LINKED ISSUE #848: Bug/Breaking Change with scoped variables, for-loop and sections
After the recent changes (v2.10.0) with scoped variables there changed also some logic in combination with `<f:for>` and `<f:section>` and their given parameters. If a parameter is given to the called section and that variable is redeclared within the `<f:for>` block with the same name, then the loop sequence is falsy.

I understand that redeclaring a variable is bad practice, but this code comes from older projects and suddenly after an update there is the main menu completely broken. Even if there were good intensions, I consider this change as breaking.

#### Example code
```
<f:variable name="menu" value="{
    0 : {
        title: '1',
        children: {
            0: {
                 title: '1.1',
                 children: {
                    0: { title: '1.1.1' },
                    1: { title: '1.1.2' },
                    2: { title: '1.1.3' }
                }
            },
            1: {
                 title: '1.2',
                 children: {
                    0: { title: '1.2.1' },
                    1: { title: '1.2.2' },
                    2: { title: '1.2.3' }
                }
            }
        }
    },
    1 : {
        title: '2',
        children: {
            0: {
                 title: '2.1',
                 children: {
                    0: { title: '2.1.1' },
                    1: { title: '2.1.2' },
                    2: { title: '2.1.3' }
                }
            }
        }
    }
}"/>

<ul>
<f:for each="{menu}" as="menuitem">
    <li>{menuitem.title}</li>
    <f:if condition="{menuitem.children}">
        <f:render section="SubMenu" arguments="{menu: menuitem.children, menuitem: menuitem}"/>
    </f:if>
</f:for>
</ul>

<f:section name="SubMenu">
    <f:comment><!-- this 'menuitem' and the one below in the for-loop are making now a problem --></f:comment>
    <f:if condition="{menuitem}">
        <div>{menuitem.title} (special label or so)</div>
    </f:if>

    <ul>
    <f:for each="{menu}" as="menuitem">
        <li>{menuitem.title}</li>
        <f:if condition="{menuitem.children}">
            <f:render section="SubMenu" arguments="{menu: menuitem.children, menuitem: menuitem}"/>
        </f:if>
    </f:for>
    </ul>
</f:section>
```

#### Actual/Falsy output:
```
1
1 (special label or so)
1.1
1.1 (special label or so)
1.1.1
1.1.1 (special label or so)
1.1.1
1.1.2
1.1.3
1.1.2
1.1.2 (special label or so)
1.1.1
1.1.2
1.1.3
1.1.3
1.1.3 (special label or so)
1.1.1
1.1.2
1.1.3
1.2
1.2 (special label or so)
1.2.1
1.2.1 (special label or so)
1.2.1
1.2.2
1.2.3
1.2.2
1.2.2 (special label or so)
1.2.1
1.2.2
1.2.3
1.2.3
1.2.3 (special label or so)
1.2.1
1.2.2
1.2.3
2
2 (special label or so)
2.1
2.1 (special label or so)
2.1.1
2.1.1 (special label or so)
2.1.1
2.1.2
2.1.3
2.1.2
2.1.2 (special label or so)
2.1.1
2.1.2
2.1.3
2.1.3
2.1.3 (special label or so)
2.1.1
2.1.2
2.1.3
```

#### Wanted/previous output
```
1
1 (special label or so)
1.1
1.1 (special label or so)
1.1.1
1.1.2
1.1.3
1.2
1.2 (special label or so)
1.2.1
1.2.2
1.2.3
2
2 (special label or so)
2.1
2.1 (special label or so)
2.1.1
2.1.2
2.1.3
```

With that code you will see (with the lastest fluid version) multiple duplicated entries in the lists, but in the previous version only the expected menu structure was rendered.
A quick fix would be by using another name for the passed parameter... but this is not so easy to do on each project in each for-loof with sections...

What would you suggest?

s2b:
With the previous implementation of ScopedVariableProvider, variables that contained arrays or object structures could behave incorrectly in a scoped context. When accessing a object path in a fluid template (e. g. myVar.subKey), the value of that "sub variable" was considered for the decision which variable (global or local) should be used.

This behavior is incorrect, only the "primary" variable name should determine if a local variable is set or not (and should fall back to a global variable). This would mean that a non-existing sub value in a local variable would trigger a fallback to the global variable.

This change restores the correct behavior: Only the "primary" variable name is checked initially. If the variable exists in the local variables, those will be used, even if the requested sub value doesn't exist there. Only if the "primary" variable name doesn't exist in local variables, global variables will be used instead.

Resolves #848

### DIFF
--- a/src/Core/Variables/ScopedVariableProvider.php
+++ b/src/Core/Variables/ScopedVariableProvider.php
@@ -96,7 +96,10 @@ final class ScopedVariableProvider extends StandardVariableProvider implements V
     public function getByPath($path): mixed
     {
         $path = $this->resolveSubVariableReferences($path);
-        return $this->localVariables->getByPath($path) ?? $this->globalVariables->getByPath($path);
+        $identifier = explode('.', $path, 2)[0];
+        return $this->localVariables->exists($identifier)
+            ? $this->localVariables->getByPath($path)
+            : $this->globalVariables->getByPath($path);
     }
 
     public function getAllIdentifiers(): array
--- a/tests/Unit/Core/Variables/ScopedVariableProviderTest.php
+++ b/tests/Unit/Core/Variables/ScopedVariableProviderTest.php
@@ -176,6 +176,13 @@ final class ScopedVariableProviderTest extends UnitTestCase
             'local',
         ];
 
+        yield 'local undefined subkey overrides global set subkey' => [
+            ['myVar' => ['myKey' => 'global']],
+            ['myVar' => ['myKey' => null]],
+            'myVar.myKey',
+            null,
+        ];
+
         yield 'variable variables using only globals' => [
             ['myVar' => ['sub' => 'global'], 'path' => 'sub'],
             [],


m41r (approved):
works as intended

PR #851, opened 2023-12-19
[FEATURE] Allow comparison with null in `<f:if>` ViewHelper

simonschaufi:
It is not possible to use a condition with the `<f:if />`
ViewHelper to check directly against null value, like it
is known from the PHP world.

For example, following php-code

```php
  // strict comparison
  if ($variable === null) {}
  if ($variable !== null) {}

  // loose comparision
  if ($variable == null) {}
  if ($variable != null) {}
```

can now be written with fluid as

```xhtml
  <!-- /* strict comparison */ -->
  <f:if condition="{var} === null"></f:if>
  <f:if condition="{var} !== null"></f:if>

  <!-- /* loose comparison */ -->
  <f:if condition="{var} == null"></f:if>
  <f:if condition="{var} != null"></f:if>
```

which extinguish the need of a custom `isNull` ViewHelper
in projects.


sbuerk (approved):

bnf (changes requested):
In my opinion `null` should be added as general language construct and not parsed from string at occasional places.
Supporting it in conditions, but not contexts like parameters makes no sense from a developer experience and makes it hard to grasp what fluid actually supports.

simonschaufi:
> In my opinion `null` should be added as general language construct and not parsed from string at occasional places.

I have no idea what exactly you mean by "general language construct". Can you given an example?

bnf:
There needs to be something like a `TYPO3Fluid\Fluid\Core\Parser\SyntaxTree\NumericNode` `TYPO3Fluid\Fluid\Core\Parser\SyntaxTree\TextNode` but for `null` values.

=> `TYPO3Fluid\Fluid\Core\Parser\SyntaxTree\NullNode`.

So that `null` is a language construct that is understood in all expressions, not just in boolean.

bnf:
Sorry, I can't provide further guidance without implementing it myself (which is currently not possible due to priorities).

s2b:
> In my opinion null should be added as general language construct and not parsed from string at occasional places.
> Supporting it in conditions, but not contexts like parameters makes no sense from a developer experience and makes it hard to grasp what fluid actually supports.

Further discussion about this can happen in #911. I'm closing this PR, since we're taking a different approach.

PR #852, opened 2023-12-19
[DOCS] Document and add tests for not (!) operator

simonschaufi:


src/ViewHelpers/IfViewHelper.php
```diff
@@ -19,6 +19,16 @@
  * Alternatively, you can write a boolean expression there.
  * Boolean expressions have the following form:
  *
+ * !{variable}
+ *
+ * ::
+ *
+ *       <f:if condition="!{foo}">
+ *           Will be shown if foo is false
```
mbrodala, line 27: We should be more verbose here. Is it actually `falsy`, so e.g. empty strings, `null`, etc. all evaluate to `true` when used with `!`?

simonschaufi: Needs to be tested. Maybe it differs for different php versions?

mbrodala: True, the actual implementation is very simple:

https://github.com/TYPO3/Fluid/blob/567a16691beedbcbefa5c12f35c5e71661cfbf53/src/Core/Parser/BooleanParser.php#L353

mbrodala: I see you did that by now, sounds good now.

mbrodala (approved):

s2b:
Thank you for this contribution.

I would move the example about the not operator after the paragraph about comparisons. Other than that this looks good to me.

simonschaufi:
The new example code only follows line 18: `As a condition is a boolean value, you can just use a boolean argument.` Otherwise the whole structure needs to be reorganized.

s2b (approved):

PR #855, opened 2024-01-12
[DOCS] Clarify IfViewHelper usage

garvinhicking:
We can't easily link to https://docs.typo3.org/m/typo3/reference-coreapi/12.4/en-us/ApiOverview/Fluid/Syntax.html due to versioning, so the auto-generated code is updated to reflect some more implementation details.

Also some rephrasing to make things easier to read (hopefully ;)).

### DIFF
--- a/src/ViewHelpers/IfViewHelper.php
+++ b/src/ViewHelpers/IfViewHelper.php
@@ -13,35 +13,51 @@ use TYPO3Fluid\Fluid\Core\ViewHelper\AbstractConditionViewHelper;
 /**
  * This ViewHelper implements an if/else condition.
  *
- * Conditions:
+ * Fluid Boolean Rules / Conditions:
+ * =================================
  *
- * As a condition is a boolean value, you can just use a boolean argument.
- * Alternatively, you can write a boolean expression there.
- * The fluid expression is evaluated as PHP expression without any change.
+ * A condition is evaluated as a boolean value, so you can use any
+ * boolean argument, like a variable.
+ * Alternatively, you can use a full boolean expression.
+ * The entered expression is evaluated as a PHP expression. You can
+ * combine multiple expressions via :php:`&&` (logical AND) and
+ * :php:`||` (logical OR).
+ *
+ * An expression can also be prepended with the :php:`!` ("not") character,
+ * which will negate that expression.
+ *
+ * Have a look into the Fluid section of the "TYPO3 Explained" Documentation
+ * for more details about complex conditions.
+ *
  * Boolean expressions have the following form:
  *
- * !{variable}
+ * `is true` variant: `{variable}`::
  *
- * ::
+ *       <f:if condition="{foo}">
+ *           Will be shown if foo is truthy.
+ *       </f:if>
+ *
+ * or `is false` variant: `!{variable}`::
  *
  *       <f:if condition="!{foo}">
  *           Will be shown if foo is falsy.
  *       </f:if>
  *
- * or
+ * or comparisons with expressions::
  *
- * XX Comparator YY
+ *       XX Comparator YY
  *
- * Comparator is one of: ==, !=, <, <=, >, >= and %
- * The % operator converts the result of the % operation to boolean.
+ * Comparator is one of: :php:`==, !=, <, <=, >, >=` and :php:`%`
+ * The :php:`%` operator (modulo) converts the result of the operation to
+ * boolean.
  *
- * XX and YY can be one of:
+ * `XX` and `YY` can be one of:
  *
- * - number
- * - Object Accessor
+ * - Number
+ * - String
+ * - Object Accessor (`object.property`)
  * - Array
  * - a ViewHelper
- * - string
  *
  * ::
  *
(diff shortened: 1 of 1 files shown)


src/ViewHelpers/IfViewHelper.php
```diff
@@ -12,36 +12,50 @@
...
 /**
  * This ViewHelper implements an if/else condition.
+ * The code entered within this ViewHelper is then only executed
+ * or shown, when the condition matches.
  *
- * Conditions:
+ * Fluid Boolean Rules / Conditions:
+ * =================================
+ *
+ * A condition is evluated as a boolean value, so you can use any boolean argument,
```
brotkrueml, line 21:
```suggestion
 * A condition is evaluated as a boolean value, so you can use any boolean argument,
```

src/ViewHelpers/IfViewHelper.php
```diff
@@ -105,6 +119,29 @@
...
+ * -----------------------------
+ *
+ * ::
+ *
+ *     <f:if condition="{user.rank} > 100 && {user.type} == 'contributor'">
+ *         <f:then>
+ *             This is being shown in case both conditions matches.
+ *         </f:then>
+ *         <f:else if="{user.rank} > 200 && ({user.type} == 'contributor' || {user.type} == 'developer')">
+ *             This is being displayed in case the first condition evaluates to TRUE and either of the second
```
brotkrueml, line 132: "first condition" is ambigious: First I thought of the "user.rank > 100 && ..." condition

src/ViewHelpers/IfViewHelper.php
```diff
@@ -105,6 +119,29 @@
...
+ *             conditionblock evalutes to TRUE.
```
brotkrueml, line 133:
```suggestion
 *             condition block evalutes to TRUE.
```

src/ViewHelpers/IfViewHelper.php
```diff
@@ -105,6 +119,29 @@
...
+ *         </f:else>
+ *         <f:else>
+ *             This is being displayed when none of the above conditions evaluated to TRUE.
+ *         </f:else>
+ *     </f:if> 
+ *
+ * Output::
+ *
+ *     Depending on which expression evaluated to TRUE, that value is displayed.
+ *     If no expression matched, the value of the final "else"-attribute is displayed.
```
brotkrueml, line 143:
```suggestion
 *     If no expression matched, the value of the final "else" attribute is displayed.
```

src/ViewHelpers/IfViewHelper.php
```diff
@@ -12,36 +12,50 @@
...
+ * Have a look into the Fluid section of the "TYPO3 Explained" Documentation for
+ * more details about complex conditions.
  *
- * As a condition is a boolean value, you can just use a boolean argument.
- * Alternatively, you can write a boolean expression there.
- * The fluid expression is evaluated as PHP expression without any change.
  * Boolean expressions have the following form:
  *
- * !{variable}
+ * (!){variable}
```
sbuerk, line 32: I don't think this get clearer with this kind of syntax.

I would suggest, that we show the two variantes one for one,
for example like following:

```php
/**
 *
 * `is true` variant:  {variable}
 * 
 * ::
 *
 *       <f:if condition="{foo}">
 *           Will be shown if foo is truthy.
 *       </f:if>
 *
 * `is false` variant: !${variable}
 *
 * ::
 *       <f:if condition="!{foo}">
 *           Will be shown if foo is falsy.
 *       </f:if> 
 *
 **/
```

Or similar - but let us really clarify this and not using even more semi inline conditional syntax not known by everyone.

sbuerk (changes requested):
@garvinhicking thank you for this pull-request.

See my code comment, I think we should really
clearify on this and not making it worser.

Can you do me a favour please, when you are on
it now ? Please squasch the commits into one
(with the commit message and description) and
ensure to not exceed line length.

We try to stay in nearly in the same line length
range like described the TYPO3 core contribution
rules, albeit not that hard limited.

Maybe something like:

```commit
[DOCS] Make IfViewHelper usage documentation clearer

We can't easily link to [1] due to versioning, so the
auto-generated code is updated to reflect some more
implementation details.

Also some rephrasing to make things easier to read
(hopefully ;)).

[1] https://docs.typo3.org/m/typo3/reference-coreapi/12.4/en-us/ApiOverview/Fluid/Syntax.html

Releases: main

```

s2b:
Sorry for the mixups in the PR, seems like multiple people were working on the same patch at the same time. However, it will be squashed anyway once this PR is accepted.

src/ViewHelpers/IfViewHelper.php
```diff
@@ -12,36 +12,50 @@
...
  * or
  *
- * XX Comparator YY
+ * ::
+ *       XX Comparator YY
  *
- * Comparator is one of: ==, !=, <, <=, >, >= and %
- * The % operator converts the result of the % operation to boolean.
+ * Comparator is one of: :php:`==, !=, <, <=, >, >=` and :php:`%`
+ * The :php:`%` operator (bitmask) converts the result of the operation to boolean.
```
s2b, line 50: `%` is the modulo operator. Not sure if this is related to bitmask, but it should be called modulo, I think.

src/ViewHelpers/IfViewHelper.php
```diff
@@ -12,36 +12,50 @@
 
 /**
  * This ViewHelper implements an if/else condition.
+ * The code entered within this ViewHelper is then only executed
+ * or shown, when the condition matches.
```
s2b, line 16: This sounds a bit too fuzzy to me. Not sure if this addition is helpful.

Apart from that, I would replace `when` with `if` because it has nothing to do with time. :)

garvinhicking:
I've implemented the suggestions and squashed the commit. Also reworked some formatting that I hope is better to understand.

s2b (approved):

s2b:
(approve once CGL is fixed)

PR #856, opened 2024-01-14
[FEATURE] SplitViewHelper

s2b:
This patch adds a new <f:format.split> ViewHelper to Fluid Standalone. It is named "split" instead of "explode" to be consistent with the naming of JavaScript.

The ViewHelper aims to support everything that TYPO3's utility functions "GeneralUtility::intExplode()" and
"GeneralUtility::trimExplode()" provide.

* items in the resulting array can be trimmed of whitespace
* items in the resulting array can be converted to integers
* empty items in the resulting array can be removed
* the "limit" option behaves similar to PHP's explode() behavior

In a following patch, its counterpart "join" will be added.

### DIFF
--- a/src/ViewHelpers/SplitViewHelper.php
+++ b/src/ViewHelpers/SplitViewHelper.php
@@ -1,0 +1,87 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file belongs to the package "TYPO3 Fluid".
+ * See LICENSE.txt that was shipped with this package.
+ */
+
+namespace TYPO3Fluid\Fluid\ViewHelpers;
+
+use TYPO3Fluid\Fluid\Core\Rendering\RenderingContextInterface;
+use TYPO3Fluid\Fluid\Core\ViewHelper\AbstractViewHelper;
+use TYPO3Fluid\Fluid\Core\ViewHelper\Traits\CompileWithRenderStatic;
+
+/**
+ * The SplitViewHelper splits a string by the specified separator, which
+ * results in an array. The number of values in the resulting array can
+ * be limited with the limit parameter, which results in an array where
+ * the last item contains the remaining unsplit string.
+ *
+ * This ViewHelper mimicks PHP's :php:`explode()` function.
+ *
+ *
+ * Examples
+ * ========
+ *
+ * Split with a separator
+ * -----------------------
+ * ::
+ *
+ *    <f:split value="1,5,8" separator="," />
+ *
+ * .. code-block:: text
+ *
+ *    {0: '1', 1: '5', 2: '8'}
+ *
+ *
+ * Split using tag content as value
+ * --------------------------------
+ *
+ * ::
+ *
+ *    <f:split separator="-">1-5-8</f:split>
+ *
+ * .. code-block:: text
+ *
+ *    {0: '1', 1: '5', 2: '8'}
+ *
+ *
+ * Split with a limit
+ * -------------------
+ *
+ * ::
+ *
+ *    <f:split value="1,5,8" separator="," limit="2" />
+ *
+ * .. code-block:: text
+ *
+ *    {0: '1', 1: '5,8'}
+ */
+final class SplitViewHelper extends AbstractViewHelper
+{
+    use CompileWithRenderStatic;
+
+    /**
+     * @var bool
+     */
+    protected $escapeChildren = false;
+
+    public function initializeArguments(): void
+    {
+        $this->registerArgument('value', 'string', 'The string to explode');
+        $this->registerArgument('separator', 'string', 'Separator string to explode with', true);
+        $this->registerArgument('limit', 'int', 'If limit is positive, a maximum of $limit items will be returned. If limit is negative, all items except for the last $limit items will be returned. 0 will be treated as 1.', false, PHP_INT_MAX);
+    }
+
+    public static function renderStatic(array $arguments, \Closure $renderChildrenClosure, RenderingContextInterface $renderingContext): array
+    {
+        $value = $arguments['value'] ?? $renderChildrenClosure();
+        if (!is_string($value)) {
+            throw new \InvalidArgumentException('Value to be split must be a string: ' . $value, 1705250408);
+        }
+
+        return explode($arguments['separator'], $value, $arguments['limit']);
+    }
+}
--- a/tests/Functional/ViewHelpers/SplitViewHelperTest.php
+++ b/tests/Functional/ViewHelpers/SplitViewHelperTest.php
@@ -1,0 +1,96 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file belongs to the package "TYPO3 Fluid".
+ * See LICENSE.txt that was shipped with this package.
+ */
+
+namespace TYPO3Fluid\Fluid\Tests\Functional\ViewHelpers;
+
+use TYPO3Fluid\Fluid\Tests\Functional\AbstractFunctionalTestCase;
+use TYPO3Fluid\Fluid\View\TemplateView;
+
+final class SplitViewHelperTest extends AbstractFunctionalTestCase
+{
+    /**
+     * @test
+     */
+    public function throwsExceptionForInvalidValue(): void
+    {
+        self::expectExceptionCode(1705250408);
+        $view = new TemplateView();
+        $view->getRenderingContext()->setCache(self::$cache);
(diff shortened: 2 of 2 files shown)


mbrodala:
TBH I see multiple viewhelpers / syntax features here:

1. Splitting a string with a custom delimiter
2. Filtering "empty" items from an array
3. Converting array items to integers

IMO this should be split accordingly to have small composable parts which could be chained:

```
{array -> f:iterator.split(delimiter: ",") -> f:iterator.filterEmpty() -> f:iterator.convert(type: "int")}
```

Notice that "iterator" would be a better NS here to hint at the subject type these viewhelpers operate on. The actually do not perform formatting (convert to string) as known in Fluid so far.

garvinhicking:
My two cents:

(Preface: Generally I'm not an advocate of using too much logic in fluid templates. The complexity/barrier should be low, so that integrators without too much coding knowledge can find their ways.)

I can see a need for converting a string to an array (for iteration) or the other way round (joining an array) so that this data can easily be passed to data-attributes or javascript-calls, where integrators have no ability/entrypoint to do that within a Controller or Service.

While the chaining is cool for developers, I'm afraid it adds too much cognitive complexity to a task, where integrators are more accustomed to specifying parameters (like "nonEmpty=true" and whathaveyou). A chained syntax looks a bit like map/reduce, which developers may find easy to read, but can create panic for "normal people"?

I think to stay within the "format" namespace would better suit established knowledge, where people search for any string transformations in that scope? TYPO3-fluid uses "transform" which would also fit somehow, but it's current application is focussed at HTML and not array/string conversion.
The NS "iterator" would lead me towards "f:for" and "f:cycle" or so, when in reality it's not addressing iteration, but rather type conversion/modification.

Closing note: Please do use the keywords "implode" and "explode" in the ViewHelper description, so that PHP developers can easily find the viewhelper (I myself wouldn't search for "split").

mbrodala:
Indeed, the input is not an iterator, so the NS should actually be `text` or `string`, so `f:text.split` or similar. Again, it's not a formatting as one is used to so IMO `format` is not the correct NS.

garvinhicking:
How about:
* Transform::split/join
* Convert::...
* Data::...
* Operations:..

IMO "formatting" would still fit, because I "format a string as array", and would use a well-known namespace.

(Filed under: There are only two hard things in Computer Science: cache invalidation, naming things and off-by-one errors)

benjaminkott:
Please let us avoid mistakes of the past and not add more bloat to our templates for base-level functionality.
It should just be `<f:split />` no need to add more unnecessary stuff here.

s2b:
Can you be more specific? What is unnecessary? What should and shouldn't the VH support?

benjaminkott:
@s2b it´s about the placement debate not about the functionality.

`<f:format.trim>` -> `<f:trim>`
`<f:text.split>` -> `<f:split>`

mbrodala:
Maybe we should also take some inspiration from Twig here: https://twig.symfony.com/doc/3.x/filters/index.html

The similarly have `format_*` filters and toplevel filters ...

benjaminkott:
> Maybe we should also take some inspiration from Twig here: https://twig.symfony.com/doc/3.x/filters/index.html
> 
> The similarly have `format_*` filters and toplevel filters ...

we almost need all of those, added the join on https://review.typo3.org/c/Packages/TYPO3.CMS/+/83442
split was next on my list, then merge

src/ViewHelpers/Format/SplitViewHelper.php
```diff
@@ -0,0 +1,72 @@
...
+     * @var bool
+     */
+    protected $escapeChildren = false;
+
+    public function initializeArguments(): void
+    {
+        $this->registerArgument('delimiter', 'mixed', 'Delimiter string to explode with', true);
+        $this->registerArgument('value', 'string', 'The string to explode');
+        $this->registerArgument('limit', 'int', 'If limit is positive, a maximum of $limit items will be returned. If limit is negative, all items except for the last $limit items will be returned. 0 will be treated as 1.', false, PHP_INT_MAX);
+        $this->registerArgument('trim', 'bool', 'Trim items in array', false, false);
```
benjaminkott, line 30: No need to overcomplicate stuff, there is no need for a splitter to trim stuff.

src/ViewHelpers/Format/SplitViewHelper.php
```diff
@@ -0,0 +1,72 @@
...
+        $this->registerArgument('removeEmpty', 'bool', 'Remove empty items from array. Items containing only whitespace are considered empty as well.', false, false);
```
benjaminkott, line 31: No need to overcomplicate stuff, there is no need for a splitter to remove empty stuff.

src/ViewHelpers/Format/SplitViewHelper.php
```diff
@@ -0,0 +1,72 @@
...
+        $this->registerArgument('convertToIntegers', 'bool', 'Convert items to integers', false, false);
```
benjaminkott, line 32: No need to overcomplicate stuff, there is no need for a splitter to convert something.

benjaminkott (changes requested):
If we continue this style, every view-helper is responsible for everything. We should stay to the basics. See twig or JS prototype implementation.

value -> split(separator, limit)

src/ViewHelpers/Format/SplitViewHelper.php
```diff
@@ -0,0 +1,72 @@
...
+    use CompileWithRenderStatic;
+
+    /**
+     * @var bool
+     */
+    protected $escapeChildren = false;
+
+    public function initializeArguments(): void
+    {
+        $this->registerArgument('delimiter', 'mixed', 'Delimiter string to explode with', true);
```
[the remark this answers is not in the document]
mbrodala: Delimiter would be a RegEx term and thus make this even more confusing whether RegEx are allowed here.

src/ViewHelpers/Format/SplitViewHelper.php
garvinhicking: My $0.02 on this: Use cases for integrators might be to receive a string like:

"cars, dogs, cats"

and they might want to put that into `data` attributes for an `<a>` tag or something like that. So they would need to trim each variable.

When the split viewhelper would not have a trim, we'd have these options:

* Do not use a viewhelper, but prepare the data in PHP scope (controller, dataprocessor, ...) -> might need more developer skills than integrator

* Create a <f:trim> viewhelper -> I think the more "small" viewhelpers we have then, that need chaining, the more fluid would be bent to contain more much logic. You'd have the whole split, foreach, trim logic there.

So this parameter to directly trim values (like in PHP the trimExplode does for us) does have some value for integrators to save one "logic" operation.

src/ViewHelpers/Format/SplitViewHelper.php
garvinhicking: Same here; if we would not have this parameter, integrators need to have one more f:if structure plus indentation to solve that. And I believe removing empty values could be a common use for this explosion.

src/ViewHelpers/Format/SplitViewHelper.php
garvinhicking: On this one I agree, I don't directly see the advantage of integer casting in a templating language that tries to hide types from the integrator. If you have for example a list of UIDs to pass to the `<f:image>` viewHelper as an image ID, this would get cast to an integer by that viewHelper.

@s2b Did you maybe have something specific in mind that could benefit from this?

src/ViewHelpers/Format/SplitViewHelper.php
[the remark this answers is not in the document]
garvinhicking: +1

src/ViewHelpers/Format/SplitViewHelper.php
andreaskienast: In Twig, you'd do something like `{{ array|map(v => v|trim) }}`. Not sure if this would be possible in Fluid at all.

s2b: > In Twig, you'd do something like `{{ array|map(v => v|trim) }}`. Not sure if this would be possible in Fluid at all.

I don't think a clean solution like this is possible without syntax changes. Something like this could be possible though:

```xml
{csv -> f:split(separator: ',') -> f:map(callback: 'f:trim')}
```

benjaminkott (approved):

sbuerk (approved):

Issue #857, opened 2024-03-01
PHP Warning: Array to string conversion in TagBuilder.php line 200


liayn:
Function parameter `$attributeValue` is allowed to be an array.
Therefore the straight conversion to string raises a warning in PHP 8.2.

liayn:
The usage seems to stem from an invalid Viewhelper implementation, still this case should be covered.

s2b:
The method seems to only support arrays if the attribute name is either data or aria. So I would suggest to throw an exception if an array is used for other attribute names.

In general, I think that this is bad API design, however I would argue that it should stay consistent to the existing API at the moment to not be breaking.

If your error occurred in a core ViewHelper, this would be good to know to be able to fix it there as well.

s2b:
@liayn Can you check if the PR fixes your problem?

liayn:
Thanks, will continue in the PR

PR #858, opened 2024-03-04
[BUGFIX] Prevent php warning when passing array as tag attribute

s2b:
TagBuilder::addAttribute() allows both strings and array/traversables as attribute value. However, in the implementation only "data" and "aria" are properly handled as arrays. For other attribute names, this leads to a type conversion warning.

With this patch, that edge case is properly handled with an exception. Also, test coverage is added both for data/aria and the exception.

Resolves: #857

### DIFF
--- a/src/Core/ViewHelper/TagBuilder.php
+++ b/src/Core/ViewHelper/TagBuilder.php
@@ -181,7 +181,10 @@ class TagBuilder
      * Adds an attribute to the $attributes-collection
      *
      * @param string $attributeName name of the attribute to be added to the tag
-     * @param string|\Traversable|array|null $attributeValue attribute value
+     * @param string|\Traversable|array|null $attributeValue attribute value, can only be array or traversable
+     *                                                       if the attribute name is either "data" or "area". In
+     *                                                       that special case, multiple attributes will be created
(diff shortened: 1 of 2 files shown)


src/Core/ViewHelper/TagBuilder.php
```diff
@@ -190,9 +193,14 @@ public function addAttribute($attributeName, $attributeValue, $escapeSpecialChar
         if ($escapeSpecialCharacters) {
             $attributeName = htmlspecialchars($attributeName);
         }
-        if (in_array($attributeName, ['data', 'aria'], true)
-            && (is_array($attributeValue) || $attributeValue instanceof \Traversable)
-        ) {
+        if (is_array($attributeValue) || $attributeValue instanceof \Traversable) {
+            if (!in_array($attributeName, ['data', 'aria'], true)) {
+                throw new \InvalidArgumentException(
+                    sprintf('Specified tag attribute "%s" cannot be of type array.', $attributeName),
```
liayn, line 199: wording suggestion: "Value of tag attribute %s can't be of type array"

s2b: fixed

lolli42 (approved):

PR #859, opened 2024-03-04
[FEATURE] Allow arbitrary arguments for TagBasedViewHelper

s2b:
Tag based ViewHelpers can now receive arbitrary arguments which
will be added to the resulting HTML tag automatically. This offers
a much more readable option compared to the "additionalAttributes"
argument, which is still available for tag based ViewHelpers. Also,
both "data" and "aria" arguments are still available to supply multiple
data or aria attributes as an array.

As a result of this change, the following methods of
AbstractTagBasedViewHelper are deprecated and will be removed
with the next major version of Fluid:

* registerUniversalTagAttributes(): The method is now a no-op since
all previously registered universal attributes can now be used without
registration.

* registerTagAttribute(): The method still has the same functionality
since it also supports the required flag as well as default values, but
is no longer necessary in its current form. For edge cases when
the required flag is still necessary, registerArgument() can be used.

### DIFF
--- a/src/Core/ViewHelper/AbstractTagBasedViewHelper.php
+++ b/src/Core/ViewHelper/AbstractTagBasedViewHelper.php
@@ -115,9 +115,7 @@ abstract class AbstractTagBasedViewHelper extends AbstractViewHelper
         }
 
         foreach ($this->additionalArguments as $argumentName => $argumentValue) {
-            if (strpos($argumentName, 'data-') === 0 || strpos($argumentName, 'aria-') === 0) {
-                $this->tag->addAttribute($argumentName, $argumentValue);
-            }
+            $this->tag->addAttribute($argumentName, $argumentValue);
         }
 
         if (isset(self::$tagAttributes[get_class($this)])) {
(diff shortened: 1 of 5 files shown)


benjaminkott:
Just to match expectations:

- data and aria attributes should only be handled IF they contain an array.
- data and aria attributes should just be passed through if they contain a "string"

Everything else:
- wonderful
- finally!

s2b:
> * data and aria attributes should just be passed through if they contain a "string"

I added tests for this. Currently, this throws an exception, which is consistent with what I would expect, since `aria` and `data` are registered parameters. This is also consistent with the current behavior. Why should it behave differently?

benjaminkott:
> > * data and aria attributes should just be passed through if they contain a "string"
> 
> I added tests for this. Currently, this throws an exception, which is consistent with what I would expect, since `aria` and `data` are registered parameters. This is also consistent with the current behavior. Why should it behave differently?

I think that the names are too generic and could mistakenly be used. It´s more like graceful to pass it through it it´s a string and cannot be handled.

lolli42 (approved):

PR #861, opened 2024-03-14
[FEATURE] JoinViewHelper

benjaminkott:
The JoinViewHelper combines elements from an array into a single string. You can specify both a general separator and a special one for the last element that is rendered between the elements.

### Simple join

```html
<f:join value="{0: '1', 1: '2', 2: '3'}" />   
```

Results in the output:

```html
123
```

### Join with separator

```html
<f:join value="{0: '1', 1: '2', 2: '3'}" separator=", " />
```

Results in the output:

```html
1, 2, 3
```

### Join with separator, and special one for the last

```html
<f:join value="{0: '1', 1: '2', 2: '3'}" separator=", " separatorLast=" and " />
```

Results in the output:

```html
1, 2 and 3
```

### DIFF
--- a/src/ViewHelpers/JoinViewHelper.php
+++ b/src/ViewHelpers/JoinViewHelper.php
@@ -76,24 +76,20 @@ final class JoinViewHelper extends AbstractViewHelper
      */
     public static function renderStatic(array $arguments, \Closure $renderChildrenClosure, RenderingContextInterface $renderingContext)
     {
-        $value = $arguments['value'] ?? null;
+        $value = $arguments['value'] ?? $renderChildrenClosure();
         $separator = $arguments['separator'] ?? '';
         $separatorLast = $arguments['separatorLast'] ?? null;
 
-        if ($value === null) {
-            $content = $renderChildrenClosure();
-            if (!is_array($content) && !$content instanceof \ArrayAccess && !$content instanceof \Traversable) {
-                $givenType = get_debug_type($content);
-                throw new \InvalidArgumentException(
-                    'The argument "value" was registered with type "array", but is of type "' .
-                    $givenType . '" in view helper "' . static::class . '".',
-                    1256475113
-                );
-            }
-            $value = $content;
+        if ($value === null || !is_iterable($value)) {
+            $givenType = get_debug_type($value);
+            throw new \InvalidArgumentException(
+                'The argument "value" was registered with type "array", but is of type "' .
+                $givenType . '" in view helper "' . static::class . '".',
+                1256475113
+            );
         }
 
-        $value = (array)$value;
+        $value = self::iteratorToArray($value);
 
         if (\count($value) === 0) {
             return '';
@@ -108,5 +104,13 @@ final class JoinViewHelper extends AbstractViewHelper
         }
 
         return implode($separator, \array_slice($value, 0, -1)) . $separatorLast . $value[\count($value) - 1];
+    }
+
+    /**
+     * This ensures compatibility with PHP 8.1
+     */
+    private static function iteratorToArray(\Traversable|array $iterator): array
+    {
+        return is_array($iterator) ? $iterator : iterator_to_array($iterator);
     }
 }
--- a/src/ViewHelpers/JoinViewHelper.php
+++ b/src/ViewHelpers/JoinViewHelper.php
@@ -1,0 +1,112 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file belongs to the package "TYPO3 Fluid".
+ * See LICENSE.txt that was shipped with this package.
+ */
+
+namespace TYPO3Fluid\Fluid\ViewHelpers;
+
+use TYPO3Fluid\Fluid\Core\Rendering\RenderingContextInterface;
+use TYPO3Fluid\Fluid\Core\ViewHelper\AbstractViewHelper;
+use TYPO3Fluid\Fluid\Core\ViewHelper\Traits\CompileWithRenderStatic;
+
+/**
+ * The JoinViewHelper combines elements from an array into a single string.
+ * You can specify both a general separator and a special one for the last
+ * element, which serves as the delimiter between the elements.
+ *
+ *
+ * Examples
+ * ========
+ *
+ * Simple join
+ * -----------
+ * ::
+ *
+ *    <f:join value="{0: '1', 1: '2', 2: '3'}" />
+ *
+ * .. code-block:: text
+ *
+ *    123
+ *
+ *
+ * Join with separator
+ * -------------------
+ *
+ * ::
+ *
+ *    <f:join value="{0: '1', 1: '2', 2: '3'}" separator=", " />
+ *
+ * .. code-block:: text
+ *
+ *    1, 2, 3
+ *
+ *
+ * Join with separator, and special one for the last
+ * -------------------------------------------------
+ *
+ * ::
+ *
+ *    <f:join value="{0: '1', 1: '2', 2: '3'}" separator=", " separatorLast=" and " />
+ *
+ * .. code-block:: text
+ *
+ *    1, 2 and 3
+ */
+final class JoinViewHelper extends AbstractViewHelper
+{
+    use CompileWithRenderStatic;
+
+    public function initializeArguments(): void
+    {
+        $this->registerArgument('value', 'array', 'An array', false);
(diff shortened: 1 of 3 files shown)


andreaskienast (changes requested):

andreaskienast (approved):

garvinhicking:
Nice, I like that "separatorLast" argument!

src/ViewHelpers/JoinViewHelper.php
```diff
@@ -0,0 +1,112 @@
...
+    public static function renderStatic(array $arguments, \Closure $renderChildrenClosure, RenderingContextInterface $renderingContext)
+    {
+        $value = $arguments['value'] ?? null;
+        $separator = $arguments['separator'] ?? '';
+        $separatorLast = $arguments['separatorLast'] ?? null;
+
+        if ($value === null) {
+            $content = $renderChildrenClosure();
+            if (!is_array($content) && !$content instanceof \ArrayAccess && !$content instanceof \Traversable) {
+                $givenType = is_object($content) ? get_class($content) : gettype($content);
```
mbrodala, line 86:
```suggestion
                $givenType = get_debug_type($content);
```

src/ViewHelpers/JoinViewHelper.php
```diff
@@ -0,0 +1,112 @@
...
+     */
+    public static function renderStatic(array $arguments, \Closure $renderChildrenClosure, RenderingContextInterface $renderingContext)
+    {
+        $value = $arguments['value'] ?? null;
+        $separator = $arguments['separator'] ?? '';
+        $separatorLast = $arguments['separatorLast'] ?? null;
+
+        if ($value === null) {
+            $content = $renderChildrenClosure();
+            if (!is_array($content) && !$content instanceof \ArrayAccess && !$content instanceof \Traversable) {
```
mbrodala, line 85:
```suggestion
            if (!is_iterable($content)) {
```

We don't need `ArrayAccess` here.

src/ViewHelpers/JoinViewHelper.php
```diff
@@ -0,0 +1,112 @@
...
+                throw new \InvalidArgumentException(
+                    'The argument "value" was registered with type "array", but is of type "' .
+                    $givenType . '" in view helper "' . static::class . '".',
+                    1256475113
+                );
+            }
+            $value = $content;
+        }
+
+        $value = (array)$value;
```
mbrodala, line 96:
```suggestion
        $value = iterator_to_array($value);
```

This ensures this really works for any `Traversable` object.

src/ViewHelpers/JoinViewHelper.php
```diff
@@ -0,0 +1,112 @@
...
+        if (\count($value) === 0) {
+            return '';
+        }
+
+        if ($separatorLast === null || $separatorLast === $separator) {
+            return implode($separator, $value);
+        }
+
+        if (\count($value) === 1) {
+            return $value[0];
```
mbrodala, line 107:
```suggestion
            return (string)$value[0];
```

The `$value` may contain arbitrary non-string values.

mbrodala (changes requested):

tests/Functional/ViewHelpers/JoinViewHelperTest .php
```diff
@@ -0,0 +1,145 @@
...
+            'expectation' => '123',
+        ];
+        yield 'value attribute inline array' => [
+            'arguments' => [
+                'value' => [1, 2, 3],
+            ],
+            'src' => '<f:join value="{0: \'1\', 1: \'2\', 2: \'3\'}" />',
+            'expectation' => '123',
+        ];
+        yield 'value inline' => [
```
s2b, line 65: Can you add another case where both inline and argument are used to test the fallback order?

tests/Functional/ViewHelpers/JoinViewHelperTest .php
```diff
@@ -0,0 +1,145 @@
...
+            'expectation' => '1, 2 and 3',
+        ];
+    }
+
+    #[DataProvider('renderValidDataProvider')]
+    #[Test]
+    public function renderValid(array $arguments, string $src, string $expectation): void
+    {
+
+        $view = new TemplateView();
```
s2b, line 104: Can you execute the rendering and testing twice here, similar to other tests? The first execution tests uncached, the second cached templates.

src/ViewHelpers/JoinViewHelper.php
benjaminkott: we still can have an array, this cannot be passed here

tests/Functional/ViewHelpers/JoinViewHelperTest .php
benjaminkott: added, found tests were not executed at all due to a space in the filename 🙈 - copy fail from Gerrit patch

src/ViewHelpers/JoinViewHelper.php
sbuerk: @mbrodala What do you mean with not needed ?

If that check is removed, object of \ArrayAccess throwing the exception instead of getting converted to an array later. This check is to prevent that. 

In case a object implements \ArrayAccess but not the \Traversable inteface (directly or indirectly) it would end in the exception with your suggestion, otherwise still hitting the conversion.

I'd say this should be then following:

```php
if (!is_iterable($content) && !$content instanceof \ArrayAccess)) {}
```

src/ViewHelpers/JoinViewHelper.php
sbuerk: benji is right due to PHP8.1 suppor. So this should be more like following:

```php
$value = ($value instanceof \Traversable)
  ? iterator_to_array($value)
  : (array) $value;
```

or what do you think ?

src/ViewHelpers/JoinViewHelper.php
s2b: Maybe we should cover the different cases with more tests to be sure everything works? I'm working on another VH and can create a basis for this (implementations of the different interfaces), so we can add more tests here as well.

s2b: Once #863 is merged, I will adjust this PR accordingly.

nhovratov:
It would be cool to have a `property` argument, so when you provide an array of objects/arrays like {data.categories} then you can map it to a specific property like `title`.

```
<f:split value="{data.categories}" property="title" separator=", "/>
```

s2b:
There seem to be very different opinions about features like this (add to join VH vs. add a separate VH). I would suggest to add the VH as-is in the first step and to extend it afterwards if it's general consensus to go this route.

chrissonntag:
Seems to be a useful ViewHelper and will be happy to use it. I'd rather have named it ImplodeViewHelper for that's what it is actually called in PHP, but that's just me.

Nevertheless I'd like to ask why the class has to be final? I've just checked and it seems that only a minority of ViewHelpers are declared final - for whatever reason even. It'd be quite convenient if we could extend this or any other ViewHelper for that matter.

src/ViewHelpers/JoinViewHelper.php
```diff
@@ -0,0 +1,116 @@
...
+
+        if (\count($value) === 0) {
+            return '';
+        }
+
+        if ($separatorLast === null || $separatorLast === $separator) {
+            return implode($separator, $value);
+        }
+
+        if (\count($value) === 1) {
```
sbuerk, line 102: Would it not make sense to move that up before L98-L100 ? I'd say, that if separatorLast is not given or spearatorLast and separator is the same it take precedence. For a single item the separator does not make any sense anyway, so this should be considerd simplified early return - or am I wrong ?

mbrodala:
> Nevertheless I'd like to ask why the class has to be final? I've just checked and it seems that only a minority of ViewHelpers are declared final - for whatever reason even. It'd be quite convenient if we could extend this or any other ViewHelper for that matter.

See e.g. https://matthiasnoback.nl/2018/09/final-classes-by-default-why/

In a perfect Fluid world each viewhelper is tiny and does a single thing, thus is atomic. Extending these wouldn't make sense in the first place and would also lead to tight coupling which would make changes to Fluid itself harder.

src/ViewHelpers/JoinViewHelper.php
s2b: I've combined the two edge-cases to one if condition.

PR #863, opened 2024-03-15
[FEATURE] ReplaceViewHelper

s2b:
The ReplaceViewHelper replaces one or multiple strings with other strings. This ViewHelper mimicks PHP's :php:`str_replace()` function. However, it's also possible to provide replace pairs as associative array via the "replace" argument.

Replace a single string:

```xml
<f:replace value="Hello World" search="World" replace="Fluid" /> <!-- Output: Hello Fluid -->
```

Replace multiple strings:

```xml
<f:replace value="Hello World" search="{0: 'World', 1: 'Hello'}" replace="{0: 'Fluid', 1: 'Hi'}" /> <!-- Output: Hi Fluid -->
```

Replace multiple strings using associative array:

```xml
<f:replace value="Hello World" replace="{'World': 'Fluid', 'Hello': 'Hi'}" /> <!-- Output: Hi Fluid -->
```

### DIFF
--- a/src/ViewHelpers/ReplaceViewHelper.php
+++ b/src/ViewHelpers/ReplaceViewHelper.php
@@ -1,0 +1,123 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file belongs to the package "TYPO3 Fluid".
+ * See LICENSE.txt that was shipped with this package.
+ */
+
+namespace TYPO3Fluid\Fluid\ViewHelpers;
+
+use TYPO3Fluid\Fluid\Core\Rendering\RenderingContextInterface;
+use TYPO3Fluid\Fluid\Core\ViewHelper\AbstractViewHelper;
+use TYPO3Fluid\Fluid\Core\ViewHelper\Traits\CompileWithRenderStatic;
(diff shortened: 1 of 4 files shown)


src/ViewHelpers/ReplaceViewHelper.php
```diff
@@ -0,0 +1,123 @@
...
+        $this->registerArgument('replace', 'mixed', '', true);
+    }
+
+    public static function renderStatic(array $arguments, \Closure $renderChildrenClosure, RenderingContextInterface $renderingContext)
+    {
+        $value = $arguments['value'] ?? $renderChildrenClosure();
+        $search = $arguments['search'];
+        $replace = $arguments['replace'];
+
+        if ($value === null) {
```
lolli42, line 75: we should probably check if $value can be converted to a string here, as well.
this however is an issue in other VH's as well, and we may want to take care of this in a more global fashion.

something like:
function can_be_string($var) {
    return $var === null || is_scalar($var) || $var instanceof Stringable);
}

lolli42 (approved):

PR #870, opened 2024-05-15
[TASK] Switch Documentation Rendering to PHP

Alagts:
You can try out the rendering locally with
```
make docs
```


Makefile
mbrodala, line 1: Please add this with `export-ignore` to `.gitattributes` since it's not useful for Composer dist archives.

garvinhicking: Couldn't it still be helpful to have in vendor/? Composer users could easily locally create documentation if they wanted to? It's a bit different than a makefile that spawns tests only...?

mbrodala: For development, sure. (But the docs should then be outputted somewhere else besides `vendor`). But on production not really. You only want to upload the bare minimum of code necessary to have a running system here. Documentation and similar should be left out here. (And yes, this would mean that the whole `Documentation` directory should also be skipped besides a few others.)

garvinhicking: I see! Maybe then a separate PR to sort this out for the other involved components/files?

Alagts: I added /Makefile export-ignore to .gitattributes.

mbrodala (approved):

Makefile
```diff
@@ -0,0 +1,14 @@
+.PHONY: help
```
s2b, line 1: Is the Makefile only used for local rendering? Or is it also necessary for CI processes to work properly?

Currently, we don't have a Makefile in the project, so I would prefer to either migrate all commands to the Makefile or to move the docs commands to composer.json.

linawolf: It is only used for the local rendering, however the issue with composer commands is, that you need php to run composer and many local machines run php in a docker container. Then composer is trying to execute docker in docker. Therefore I would not advise to add it to the composer.json. You could, however add the command to the runTests.sh if you want. You can also omit the Makefile and ppl can execute the docker or podman command directly. However we found that a Makefile is very helpful to a large number of contributors and we use it all over the official documentation. We also advise it to all Extensions. And also there is really no harm in having a Makefile. You might want to add shortcuts for the runTests.sh in future

sbuerk:
We will discuss this dedicated how to proceed and how which way of dispatching we use in standalone fluid (runTests.sh approach / composer script approach / Makefile (with all).
Also adding a CONTRIBUTION.md file to describe these kind of things (test execution and so on).

Thus remove the Makefile from this PR for now and integrate the CI run for it first and directly and not postpone this change until decision of the topics above are made.

Alagts:
I removed the Makefile. Now, you can try out the rendering locally with
```
docker run --rm --pull always -v $(pwd):/project -it ghcr.io/typo3-documentation/render-guides:latest --config=Documentation
```

s2b (approved):

s2b:
Thanks! :)

PR #871, opened 2024-05-22
[TASK] Use xxh3 instead of sha1

lolli42:
64bit xxh3 hash is significantly quicker. This is a micro optimization especially when creating hashes from template contents for caching.

### DIFF
--- a/examples/src/CustomVariableProvider.php
+++ b/examples/src/CustomVariableProvider.php
@@ -37,7 +37,7 @@ class CustomVariableProvider extends StandardVariableProvider implements Variabl
     public function getByPath($path)
     {
         if ($path === 'random') {
-            return 'random' . sha1(rand(0, 999999999));
+            return 'random' . hash('xxh3', (string)rand(0, 999999999));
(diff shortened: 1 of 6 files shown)


sbuerk (approved):

examples/src/CustomVariableProvider.php
```diff
@@ -37,7 +37,7 @@ class CustomVariableProvider extends StandardVariableProvider implements Variabl
     public function getByPath($path)
     {
         if ($path === 'random') {
-            return 'random' . sha1(rand(0, 999999999));
+            return 'random' . hash('xxh3', (string)rand(0, 999999999));
```
jdreesen, line 40: Why do you need to hash this at all?

lolli42: unsure. it's example code - didn't wanted to bother with this much at this point.

jdreesen: Oh, right, I didn't see it's example code.

Issue #873, opened 2024-06-11
ViewHelper to use PHP constants in Fluid


s2b:
There should be a ViewHelper to access the values of PHP constants from a Fluid template. To accomplish this, a new ViewHelper should be created that wraps the [constant()](https://www.php.net/manual/de/function.constant.php) function of PHP.

PR #874, opened 2024-06-11
[FEATURE] Add f:constant ViewHelper

andreaskienast:
This commit adds the `f:constant` ViewHelper allowing to use native PHP constants and enums in Fluid context.

Fixes: #873

### DIFF
--- a/src/ViewHelpers/ConstantViewHelper.php
+++ b/src/ViewHelpers/ConstantViewHelper.php
@@ -1,0 +1,62 @@
+<?php
+
+/*
+ * This file belongs to the package "TYPO3 Fluid".
+ * See LICENSE.txt that was shipped with this package.
+ */
+
+namespace TYPO3Fluid\Fluid\ViewHelpers;
+
+use TYPO3Fluid\Fluid\Core\Rendering\RenderingContextInterface;
+use TYPO3Fluid\Fluid\Core\ViewHelper\AbstractViewHelper;
+use TYPO3Fluid\Fluid\Core\ViewHelper\Traits\CompileWithRenderStatic;
+
+/**
+ * Wrapper for PHPs :php:`constant` function.
+ * See https://www.php.net/manual/function.constant.php.
+ *
+ * Examples
+ * ========
+ *
+ * Get built-in PHP constant
+ * -------------------------
+ *
+ * ::
+ *
+ *    {f:constant(value: 'PHP_INT_MAX')}
+ *
+ * Output::
+ *
+ *    9223372036854775807
+ *    (Depending on CPU architecture).
+ *
+ * Get class constant
+ * ------------------
+ *
+ * ::
+ *
+ *    {f:constant(value: '\Vendor\Package\Class::CONSTANT')}
+ *
+ * Get enum value
+ * --------------
+ *
+ * ::
+ *
+ *    {f:constant(value: '\Vendor\Package\Enum::CASE')}
+ */
+class ConstantViewHelper extends AbstractViewHelper
+{
+    use CompileWithRenderStatic;
+
+    public function initializeArguments(): void
+    {
+        parent::initializeArguments();
+        $this->registerArgument('name', 'string', 'String representation of a PHP constant or enum');
+    }
+
+    public static function renderStatic(array $arguments, \Closure $renderChildrenClosure, RenderingContextInterface $renderingContext): mixed
+    {
+        $name = $arguments['name'] ?? $renderChildrenClosure();
+        return constant($name);
+    }
+}
--- a/tests/Functional/ViewHelpers/ConstantViewHelperTest.php
+++ b/tests/Functional/ViewHelpers/ConstantViewHelperTest.php
@@ -1,0 +1,112 @@
+<?php
+
+declare(strict_types=1);
+
+/*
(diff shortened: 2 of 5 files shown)


src/ViewHelpers/ConstantViewHelper.php
```diff
@@ -0,0 +1,62 @@
...
+ * ::
+ *
+ *    {f:constant(value: '\Vendor\Package\Class::CONSTANT')}
+ *
+ * Get enum value
+ * --------------
+ *
+ * ::
+ *
+ *    {f:constant(value: '\Vendor\Package\Enum::CASE')}
```
mbrodala, line 45: We should clarify if the leading backslash is required or optional.

andreaskienast: Apparently both cases work, at least the tests didn't fail locally.

sascha-egerer, line 45: In my opinion, this leads to the misuse of enums. 
If these are used in a condition, for example, the enums should be compared and not the "backed value", which would be the case here. The enum is then cast into a "string" here. This works, but in my opinion it should not be used like this.

sascha-egerer (changes requested):

src/ViewHelpers/ConstantViewHelper.php
sascha-egerer: Ah I was wrong. Passing a enum to "constant()" does return the ENUM and not the "backed value". So ok, then I'm fine with it even if i would not expect that in the first place from a "Constant" ViewHelper.

So in this case it woult also be possible to call methods on the enum (if they're using a method name callable by fluid like "isXXX" or "getXXX")

src/ViewHelpers/ConstantViewHelper.php
```diff
@@ -0,0 +1,62 @@
...
+ *    (Depending on CPU architecture).
+ *
+ * Get class constant
+ * ------------------
+ *
+ * ::
+ *
+ *    {f:constant(value: '\Vendor\Package\Class::CONSTANT')}
+ *
+ * Get enum value
```
sascha-egerer, line 40:
```suggestion
 * Get enum instance
```

sascha-egerer (changes requested):

src/ViewHelpers/ConstantViewHelper.php
mbrodala: Actually `value` is the correct term for items of enumerations. There are no instances, even though enums in PHP may behave like such.

sascha-egerer: An instance represents a value but it's IMO still an instance. That term is also used in the PHP documentation.

https://www.php.net/manual/en/language.enumerations.overview.php
> ... The Enum itself is a class, and its possible cases are all single-instance objects of that class. ...

https://www.php.net/manual/en/language.enumerations.object-differences.php
> Cloning an Enum case is not supported, as cases must be singleton instances.

tests/Functional/ViewHelpers/ConstantViewHelperTest.php
```diff
@@ -0,0 +1,92 @@
...
+
+        $templateSources = [
+            '<f:constant value="{value}" />',
+            '<f:constant>{value}</f:constant>',
+            '{f:constant(value: \'{value}\')}',
+        ];
+
+        foreach ($templateSources as $templateSource) {
+            $view->getRenderingContext()->getTemplatePaths()->setTemplateSource($templateSource);
+            self::assertSame($expected, $view->render());
```
s2b, line 89: Currently, we execute all functional test cases twice to test both uncached and cached behavior, see other VH tests. Can you add this to your test as well?

src/ViewHelpers/ConstantViewHelper.php
```diff
@@ -0,0 +1,62 @@
...
+ *    {f:constant(value: '\Vendor\Package\Enum::CASE')}
+ */
+class ConstantViewHelper extends AbstractViewHelper
+{
+    use CompileWithRenderStatic;
+
+    public function initializeArguments(): void
+    {
+        parent::initializeArguments();
+        $this->registerArgument('value', 'string', 'String representation of a PHP constant or enum');
```
s2b, line 54: I'd suggest to use the parameter name `name` instead, as you specify the name of a constant to get its value as a result.

src/ViewHelpers/ConstantViewHelper.php
s2b: Can you add a test for this? I think that one (either constant or one of the enums) should be enough to cover this.

I'm not sure, but this might get relevant if we start to mess with cache representations of Fluid templates.

s2b: I think that this is not a big problem in the current implementation. I would leave it as-is.

src/ViewHelpers/ConstantViewHelper.php
s2b: I think, in the context of Fluid, `value` covers >95% of use cases, since methods can't be called anyways from Fluid (apart from get/is/has).

s2b (approved):

PR #880, opened 2024-06-25
[FEATURE] Support for boolean tag attributes

s2b:
TagBasedViewHelpers now have proper support for boolean attributes.
Before this change, it was very cumbersome to generate these with Fluid,
now it's implemented similar to popular JavaScript frameworks:

```
<my:viewhelper async="{true}" />
Result: <tag async="async" />

<my:viewhelper async="{false}" />
Result: <tag />

<my:viewhelper async="{null}" />
Result: <tag />
```

This can also be used in combination with variable casting:

```
<my:viewhelper async="{myString as boolean}" />
```

For compatibility reasons empty strings still lead to the attribute
being omitted from the tag. This might change in the future, however
we don't want to break templates now. ViewHelpers can define
an argument manually and implement the desired behavior in the
render method (this is how the ImageViewHelper in TYPO3 deals with
empty `alt` attributes).

### DIFF
--- a/src/Core/ViewHelper/AbstractTagBasedViewHelper.php
+++ b/src/Core/ViewHelper/AbstractTagBasedViewHelper.php
@@ -116,8 +116,11 @@ abstract class AbstractTagBasedViewHelper extends AbstractViewHelper
         }
 
         foreach ($this->additionalArguments as $argumentName => $argumentValue) {
+            // This condition is left here for compatibility reasons. Removing this will be a breaking change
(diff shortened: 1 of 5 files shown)


s2b:
#911

s2b:
This PR is based on #949

s2b:
Test run with TYPO3 core: https://review.typo3.org/c/Packages/TYPO3.CMS/+/85568

lolli42 (approved):

PR #882, opened 2024-06-26
[TASK] Tune RenderingContext withAttribute() and getAttribute()

lolli42:
getAttribute($className) now always returns an instance of $className, or throws. This avoids checking the returned instance and helps with type hinting in IDEs.


lolli42:
phpstorm issue: https://youtrack.jetbrains.com/issue/WI-75368/class-stringT-does-not-infer-the-right-type

s2b (approved):

PR #884, opened 2024-06-26
[TASK] Avoid composer 'scripts' and 'scripts-descriptions'

lolli42:
Fluid standalone devs should rely on the calls
performed by GH workflows. It is a burden to
keep composer scripts in sync.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -40,29 +40,5 @@
     },
     "bin": [
         "bin/fluid"
-    ],
-    "scripts": {
-        "cgl:check": "@php vendor/bin/php-cs-fixer fix --dry-run --diff",
-        "cgl:fix": "@php vendor/bin/php-cs-fixer fix --diff",
-        "ci:all": [
-            "@composer validate",
-            "@lint:php",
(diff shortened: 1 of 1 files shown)


sbuerk (approved):

lolli42:
Sorry, but no.

In general, I'm not a fan of composer scripts in the first place: They start with naming issues you then have to remember, gives additional headaches with additional arguments, is opinionated, and need maintenance.

In this very case, the 'dev' tasks are really simple:
```
composer update
vendor/bin/php-cs-fixer fix
vendor/bin/phpstan analyze
vendor/bin/phpunit
```

Plus the optional casual command options these binaries provide, if a dev wants that. Maybe also with a different php executor like this: /usr/bin/php8.3 vendor/bin/phpunit.

For this library, I see neither the need to encapsulate (and maintain) these binary calls every dev should know in composer.json, nor to abstract this away via something like runTests.sh - I think it's fair to require a local PHP in fluid standalone for devs who wish to develop things.

PR #892, opened 2024-06-27
[BUGFIX] Correct attribute name in ConstantViewHelper example

eliashaeussler:
The attribute in `ConstantViewHelper` is called `name`, but the documentation references it as `value`. This PR fixes the examples in the documentation.


s2b (approved):

s2b:
good catch, thx!

Issue #910, opened 2024-07-18
Consider throwing exception if namespace uses https


s2b:
It should be considered if it makes sense to throw an exception if a Fluid namespace is registered with `https://typo3.org/ns/` instead of `http://typo3.org/ns/`. This would be a breaking change.

liayn:
Related #385

mbrodala:
A few more options:
- Change to `typo3.org/ns/` (thus drop the "scheme" altogether)
- `org.typo3.ns` to prevent further confusion altogether

lolli42:
@mbrodala This would be heavily breaking for everyone. I don't think we should do this now.

PR #914, opened 2024-07-20
[!!!][TASK] Throw exception for https namespaces

s2b:
One way to make custom ViewHelpers available in Fluid templates is to define them as xml namespaces. Fluid parses `<html>` tags in templates, extracts all XML namespaces relevant to Fluid and removes them from the resulting HTML output.

XML namespaces by convention are usually named as URIs. However, they are actually not "callable" URLs but rather unique identifiers. Thus, there is only one valid way to reference Fluid namespaces as xml namespaces: By using the following schema:

```
http://typo3.org/ns/$Vendor/$Package/ViewHelpers/
```

Using https instead of http is invalid. Previously, Fluid namespaces defined with https instead of http were ignored and kept in the resulting output. Now, we catch this bug in templates and throw an exception instead to make tempate authors aware that there is an issue in their template.

This change might break existing templates. However, this is easily avoidable by checking that no template in your project includes the following string:

``` 
https://typo3.org/ns/
```

### DIFF
--- a/src/Core/Parser/Patterns.php
+++ b/src/Core/Parser/Patterns.php
@@ -13,6 +13,7 @@ namespace TYPO3Fluid\Fluid\Core\Parser;
 abstract class Patterns
 {
     public const NAMESPACEPREFIX = 'http://typo3.org/ns/';
+    public const NAMESPACEPREFIX_INVALID = 'https://typo3.org/ns/';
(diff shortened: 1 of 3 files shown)


s2b:
Resolves #910

lolli42 (approved):

PR #918, opened 2024-07-20
[TASK] Introduce type hints to ViewHelper-related API

s2b:
With v4 of Fluid, the goal is to not break any custom ViewHelpers out there. Thus, we only apply type hints to ViewHelper-related API for cases where we can be reasonably sure that it won't break ViewHelpers.

AbstractViewHelper is extended directly by custom ViewHelpers, so we can only type hint private properties.

ViewHelperInvoker and ViewHelperResolver are APIs that are implemented by frameworks that use Fluid for templating, so stricter types don't affect ViewHelper implementations. Still, there are some cases in which we won't enforce types because it is unclear if there are edge cases out there where the supplied types actually match existing PHPDOC annotations. For example, we can't really be sure if templates or ViewHelpers always return strings, so we stay away from those type hints for now.

TagBuilder, ArgumentDefinition and ViewHelperVariableContainer are often utilized from within ViewHelpers. However, because of their respected isolated functionality, we can be pretty sure about the types and thus apply them for most methods and properties.

### DIFF
--- a/src/Core/ViewHelper/ViewHelperVariableContainer.php
+++ b/src/Core/ViewHelper/ViewHelperVariableContainer.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 /*
  * This file belongs to the package "TYPO3 Fluid".
  * See LICENSE.txt that was shipped with this package.
@@ -7,6 +9,7 @@
 
 namespace TYPO3Fluid\Fluid\Core\ViewHelper;
 
+use Traversable;
 use TYPO3Fluid\Fluid\View\ViewInterface;
 
 /**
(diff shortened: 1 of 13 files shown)


src/Core/ViewHelper/ViewHelperVariableContainer.php
```diff
@@ -20,14 +23,11 @@ class ViewHelperVariableContainer
...
+     * @var array<string, array<string, mixed>>
      */
-    protected $objects = [];
+    protected array $objects = [];
 
-    /**
-     * @var ViewInterface
-     */
-    protected $view;
+    protected ?ViewInterface $view;
```
lolli42, line 30: protected ?ViewInterface $view = null;

lolli42 (commented):
should have a changelog entry.

tests/Unit/Core/ViewHelper/ViewHelperVariableContainerTest.php
```diff
@@ -105,7 +105,7 @@ public function getAllReturnsDefaultIfNotFound(): void
     #[Test]
     public function addAllThrowsInvalidArgumentExceptionOnUnsupportedType(): void
     {
-        $this->expectException(\InvalidArgumentException::class);
+        $this->expectException(\TypeError::class);
```
lolli42, line 108: remove test. there is no need to test PHP itself.

lolli42 (approved):

PR #930, opened 2024-07-21
[!!!][TASK] Apply more type hints

s2b:
More type hints are applied in the following areas:

* ConsoleRunner
* Cache interface and implementations
* Cache warmup interface and implementations
* Variable providers
* Error handlers

While these changes are breaking, they shouldn't affect the vast majority of users,
since in most cases the Fluid implementations of the affected interfaces are used.


lolli42 (approved):
I think this should be fine without too many headaches. Even FluidCacheInterface changes should be fine, i think.

PR #940, opened 2024-07-27
[BUGFIX] Restore previous empty tag attribute behavior

s2b:
With Fluid 2.12, we introduced the ability to provide arbitrary
ViewHelper arguments to TagBasedViewHelpers. Consequently,
we deprecated the method registerTagAttribute(), which is no
longer necessary due to the mentioned change.

However, the implementations differed until now: The previous
implementation ignored empty strings for tag attributes, while
the new implementation added such input as empty attribute
to the resulting tag. We now check for empty strings and only
add attributes to the tag if it's not an empty string.

This will be backported to Fluid 2.14.

### DIFF
--- a/src/Core/ViewHelper/AbstractTagBasedViewHelper.php
+++ b/src/Core/ViewHelper/AbstractTagBasedViewHelper.php
@@ -115,7 +115,9 @@ abstract class AbstractTagBasedViewHelper extends AbstractViewHelper
         }
 
         foreach ($this->additionalArguments as $argumentName => $argumentValue) {
-            $this->tag->addAttribute($argumentName, $argumentValue);
+            if ($argumentValue !== null && $argumentValue !== '') {
+                $this->tag->addAttribute($argumentName, (string)$argumentValue);
+            }
         }
 
         if (isset(self::$tagAttributes[get_class($this)])) {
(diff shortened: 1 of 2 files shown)


ErHaWeb (approved):

src/Core/ViewHelper/AbstractTagBasedViewHelper.php
```diff
@@ -115,7 +115,9 @@ public function initialize()
         }
 
         foreach ($this->additionalArguments as $argumentName => $argumentValue) {
-            $this->tag->addAttribute($argumentName, $argumentValue);
+            if ($argumentValue !== null && $argumentValue !== '') {
+                $this->tag->addAttribute($argumentName, (string)$argumentValue);
```
garvinhicking, line 119: I assume 'stringable' objects are explicitly okay here, but would some check on arrays or objects without string convert methods need to be included here? Or would an error at this place meet expectations? (Fine by me; mainly just asking)

garvinhicking (approved):

s2b:
Related TYPO3 Forge Issue: https://forge.typo3.org/issues/104490

src/Core/ViewHelper/AbstractTagBasedViewHelper.php
s2b: I think that we can depend on PHP's errors here. Considering the strange behavior here, I would expect that we change the behavior with a future Fluid version anyways.

garvinhicking: Ack, thx!

lolli42 (approved):

PR #947, opened 2024-08-06
[TASK] Add type hints to TemplateParser

s2b:
This patch adds as many types as is currently feasable. Minor changes to tests are necessary because of excessive mocking and tests of internal functionality.

With strict_types enabled, about half of all tests will be red. This is mainly due to invalid types being passed to php-internal functions, such as:

> Fatal error: Uncaught TypeError: substr(): Argument #2 ($offset) must be of type int, bool given in /var/www/html/Fluid/src/Core/Parser/TemplateParser.php:228

This class still needs a lot of work, but for now the types are good enough.


s2b:
Core CI is green (except for composer-integration): https://review.typo3.org/c/Packages/TYPO3.CMS/+/85544

lolli42 (approved):

PR #949, opened 2024-08-09
[!!!][FEATURE] Boolean and Null literals

s2b:
This patch enhances Fluid's capabilities of dealing with booleans and null significantly. Previously, the actual booleans `false` and `true` could be only created from within inline syntax or if a ViewHelper argument was defined as boolean (in which case strings were parsed as boolean). `null` values couldn't really be created at all. Now, there exist standard variables for true, false, and null, which can be used in all contexts.

This patch is marked as breaking because it disallows the usage of the variable names `true`, `false` and `null` in variable providers.


lolli42 (approved):

simonschaufi:
ok. This should really be documented :smile:

s2b:
Yes, you're right about that. We were busy with development, the Fluid documentation needs a general overhaul. On the TYPO3 side, this is already happening based on the RST files.

PR #950, opened 2024-08-09
[TASK] Trigger deprecation if true/false/null variable is used

s2b:
These variable names are no longer available in Fluid v4 because boolean literals will become part of the language.


lolli42 (changes requested):
this should go to 2.15, not 2.14: we need a new minor for this change.

Issue #895, opened 2024-07-11
Fluid Template Variable Parsing Error with Spaces around Operators


garvinhicking:
From: https://forge.typo3.org/issues/104363

h2. Steps to Reproduce:

# Use the following Fluid template code:
<pre><code class="html">
testVar: {testVar} <br/>
A: {testVar - 1} <br/>
B: {testVar - 1 } <br/>
C: {testVar-1} <br/>
D: {testVar + 1} <br/>
E: {testVar + 1 } <br/>
F: {testVar+1} <br/>
</code></pre>
# Set testVar to any integer, e.g., 2.

h3. Expected Output:

<pre><code>testVar: 2
A: 1
B: 1
C: 1
D: 3
E: 3
F: 3
</code></pre>

h3. Actual Output:

<pre><code>testVar: 2
A: 1
B: {testVar - 1 }
C:
D: 3
E: {testVar + 1 }
F: 3
</code></pre>

h2. Analysis:

* The expression {testVar - 1} and {testVar + 1} are parsed correctly when there are spaces around the operator.
* {testVar - 1 } and {testVar + 1 } result in no evaluation and return the expression as a string.
* {testVar+1} without space is parsed and evaluated correctly.
* {testVar-1} without space is not parsed and evaluated correctly.

s2b:
Case C isn't and won't be supported because `testVar-1` is treated as a variable identifier. This is especially useful for array keys properties:

`{myVar.my-array-key.test}`

The leading/trailing whitespace are addressed with the attached PR.

PR #954, opened 2024-08-11
[BUGFIX] Allow leading/trailing whitespace in math expression

s2b:
The parser of MathExpressionNode already allowed whitespace before
or after an expression. However, ExpressionNodes are only executed
if their `$detectionExpression` evaluates to true. This was not the case
for leading/trailing whitespace, which is why the regular expression is
extended to allow this.

Unfortunately, the test did only evaluate the parser but did not check
the regular expression beforehand. As there is no API for this (the
regular expression is the API and is executed directly in
`TemplateParser::textAndShorthandSyntaxHandler()`, we execute
the regex manually in the test as well. One existing test case is
removed because it never was actually supported.

In the future, we probably should implement a proper API for this.


s2b:
Resolves #895

lolli42 (approved):

Issue #894, opened 2024-07-10
Allow invalid Fluid tags inside f:comment


georgringer:
this is a duplicate of https://forge.typo3.org/issues/97835 to have it in correct project

> Currently i update all <f:widget.paginate>
> if i wrap this widget with f:comment the f:widget.paginate still throws an error.
> I now i have to remove this widget, but for testing i want wrap this in a comment to write the new code.
> i think fluid should ignore code which is wrapped with comments

georgringer:
wtf about the name

mbrodala:
Yup. In any case this doesn't work:

```
<f:cache.disable><f:widget.paginate></f:widget.paginate></f:cache.disable>
```

This still throws the `The ViewHelper "<f:widget.paginate>" could not be resolved.` error.

s2b:
Side note: These cache.* ViewHelpers are really confusing and will probably be deprecated in the future.

mbrodala:
As a first step making `f:comment` like expected and described here would likely help. Evidently "disabling the compiler" also is not enough, since we'd actually need a "disable the parser" flag.

lolli42:
Yes. We're deep in the parser here.

The current "solution" is documented in f:comment: "<f:comment><![CDATA[<f:some.invalid.syntax />]]></f:comment>".
Wrap the content in CDATA works. Reason: The NamespaceDetectionTemplateProcessor *removes* any CDATA wrapped content in an early step.

CDATA is of course ugly and it requires this manual interaction.

I think fluid has currently no solution to "stop" parsing sub elements.

And it gets worse: ext:styleguide has this usecase: It has a code viewhelper that should output html. If that code contains fluid VH's, those are interpreted - similar to the f:comment scenario. The CDATA "solution" does not work here, since the entire content vanishes ...

What we would need is a registry of VH's, to tell the parser if a VH is used in a template, the template parsing should stop. It looks as if this requires to make the regex $SPLIT_PATTERN_TEMPLATE_DYNAMICTAGS more dynamic, so fluid splits these parts differently on parser level. Such a change will require a bit of refactoring and a good solution for the registry.

lolli42:
Related: #808, #474

lolli42:
There is also `{parsing off}` to stop parsing an entire template, see #219

s2b:
This will be possible with Fluid v4, the change has just been merged.

rr-it:
This should work to use anything (invalid) inside a comment:

```
[...]

PR #956, opened 2024-08-11
[FEATURE] Fluid code in comments no longer interpreted

s2b:
Previously, the behavior of the `<f:comment>` ViewHelper was confusing:
Even if invalid Fluid code (e. g. syntax errors or a non-existent ViewHelper)
was wrapped with `<f:comment>`, it would still be interpreted by the
Fluid parser and an exception would be thrown.

This patch resolves this similarly to CDATA handling. It introduces a new
TemplateProcessor, which is executed before any other parsing and strips
out Fluid comments. This processor is then registered in the default
RenderingContext implementation. For systems that set their own
TemplateProcessors, this would probably need adjustments. However,
this change is not breaking since the old functionality is still there.

Fluid's own tests however now assume that this new TemplateProcessor
is registered and are adjusted properly in this patch.


s2b:
Resolves #894

lolli42 (approved):

PR #965, opened 2024-08-11
[FEATURE] Provide contentArgument API to all ViewHelpers

s2b:
Previously, `CompileWithContentArgumentAndRenderStatic` could
be used to create ViewHelpers that can get their default input both
from their child content and from a defined argument. This feature
has now been integrated into AbstractViewHelper. To use it, the
ViewHelper class needs to override the method
`resolveContentArgumentName()` and return the name of the
argument to be used.

Example:

```
public function initializeArguments(): void
{
    $this->registerArgument('value', 'string', '');
}

public function resolveContentArgumentName(): string
{
    return 'value';
}
```

`$this->renderChildren()` can then be used to get either the
defined argument or as fallback the ViewHelper's child content.

With this definition, the ViewHelper can be used in the following
ways:

```
<my:viewhelper value="test" />
<my:viewhelper>test</my:viewhelper>
{myInput -> my:viewhelper()}
```

Since it's no longer necessary, the trait is now deprecated and
will be removed with Fluid v5. There will be a soft deprecation
in Fluid v2 to let users know that this API will no longer be available
with Fluid v5 and will throw a deprecation error in v4.

All ViewHelpers still using the trait have been migrated to
`CompileWithRenderStatic`, which has been modified to support
the new implementation.

### DIFF
--- a/src/Core/ViewHelper/AbstractViewHelper.php
+++ b/src/Core/ViewHelper/AbstractViewHelper.php
@@ -147,6 +147,17 @@ abstract class AbstractViewHelper implements ViewHelperInterface
     }
 
     /**
+     * Returns the name of variable that contains the value to use instead of render children closure, if specified.
+     * ViewHelpers that want to use contentArgumentName are expected to override this method with their own implementation.
+     *
+     * @api
+     */
+    public function getContentArgumentName(): ?string
+    {
+        return null;
+    }
+
+    /**
      * Register a new argument. Call this method from your ViewHelper subclass
      * inside the initializeArguments() method. If an argument with the same name
      * is already defined, it will be overridden.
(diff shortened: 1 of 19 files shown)


src/Core/ViewHelper/AbstractViewHelper.php
```diff
@@ -146,6 +146,18 @@ public function isOutputEscapingEnabled()
         return $this->escapeOutput !== false;
     }
 
+    /**
+     * Returns the name of variable that contains the value to use instead of render children closure, if specified.
+     * ViewHelpers that want to use contentArgumentName are expected to override this method with their own implementation.
+     *
+     * @return string|null
```
mbrodala, line 153: This should use native type hints instead.

src/Core/ViewHelper/AbstractViewHelper.php
```diff
@@ -146,6 +146,18 @@ public function isOutputEscapingEnabled()
...
+     * @api
+     */
+    public function resolveContentArgumentName()
```
mbrodala, line 156: If this needs to be resolved or not shouldn't matter to the API. Thus `getContentArgumentName` would be more appropriate.

mbrodala (changes requested):

src/Core/ViewHelper/AbstractViewHelper.php
s2b: I thought a bit about it. While migrating ViewHelpers is a bit more effort now, on the whole it's best to introduce a better name (that can then also use a type hint, which wouldn't be possible with the old name).

lolli42 (approved):

PR #968, opened 2024-08-13
[BUGFIX] Respect getContentArgumentName() in TemplateParser

s2b:
With the introduction of the new `getContentArgumentName()`
API for ViewHelpers, there has been an oversight in `TemplateParser`: It had special handling for the old method name
`resolveContentArgumentName()`, which is now deprecated. Since
`resolveContentArgumentName()` is now called by the new API
`getContentArgumentName()`, we can just call the new API in the
`TemplateParser` now, without special handling for the deprecated
trait.

As a consequence, we need to adjust some tests and fixtures. Also,
we add two dedicated test cases to our new test file.

In the future, there needs to be a cleanup of `AbstractViewHelper`
and `ViewHelperInterface`, since internal classes depend on API
only defined in `AbstractViewHelper`. In fact, we arguably made things
a bit worse now, but this is a topic for Fluid v5.


src/Core/Parser/TemplateParser.php
```diff
@@ -531,7 +531,7 @@ protected function isArgumentEscaped(ViewHelperInterface $viewHelper, ?ArgumentD
         $isBoolean = $hasDefinition && ($argumentDefinition->getType() === 'boolean' || $argumentDefinition->getType() === 'bool');
         $escapingEnabled = $this->configuration->isViewHelperArgumentEscapingEnabled();
         $isArgumentEscaped = $hasDefinition && $argumentDefinition->getEscape() === true;
-        $isContentArgument = $hasDefinition && method_exists($viewHelper, 'resolveContentArgumentName') && $argumentDefinition->getName() === $viewHelper->resolveContentArgumentName();
+        $isContentArgument = $hasDefinition && $argumentDefinition->getName() === $viewHelper->getContentArgumentName();
```
lolli42, line 534: maybe keep the method_exists() since it's not in the interface? this should happify phpstan ...

s2b: My thinking was: Before it wasn't in `AbstractViewHelper`, only in the trait, so it made sense to check for its existence. Now it's in the abstract class every ViewHelper extends from, just like `$viewHelper->isChildrenEscapingEnabled()` two lines below it, which also has an entry in phpstan baseline. So at least it's consistent.

lolli42: fine with me. we'll have to resolve these at some point anyways. always two they are ;)

lolli42 (approved):

PR #978, opened 2024-08-18
[!!!][TASK] Extend ViewHelperInterface to cover current API

s2b:
In the past, some ViewHelper API methods that were used by Fluid
internally were only part of `AbstractViewHelper`, which lead to
either `method_exists()` checks or phpstan warnings in the baseline.

With this change, the `ViewHelperInterface` becomes the internal
API definition for ViewHelpers, only to be used by Fluid internally.
Thus, missing API methods are added to the interface, while
user-facing API only used by `AbstractViewHelper` has been
removed.

In the future, we might consider adding a separate interface for the
used-facing API of ViewHelpers. For now we highly recommend
to keep using `AbstractViewHelper` as the basis for all custom
ViewHelper implementations.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1,27 +1,7 @@
 parameters:
 	ignoreErrors:
 		-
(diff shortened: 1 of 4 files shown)


s2b:
TYPO3 Core CI is green(ish): https://review.typo3.org/c/Packages/TYPO3.CMS/+/85674

lolli42 (approved):

PR #979, opened 2024-08-20
Backport

s2b:


lolli42 (approved):

bwaidelich:
FYI: This change broke most of the Fluid adapter tests in Flow because mocking like
```php
$this->viewHelper->expects(self::once())->method('renderChildren')->will(self::returnValue('some text'));
```
won't work and, worse, because executing the tests leads to an exception
```
Error: Call to a member function evaluateChildNodes() on null
typo3fluid/fluid/src/Core/ViewHelper/AbstractViewHelper.php:329
```

We're investigating how to fix this on our side

bwaidelich:
@s2b No need to be sorry, it was just a FYI. I think we "fixed" the issue with https://github.com/neos/flow-development-collection/pull/3390

lolli42:
Quick note.
You may be better off avoiding mocks in VH tests.
Both fluid standalone and typo3 core basically transitioned to "functional" tests: Feed a TemplateView with some fluid template test string that calls the VH, render(), assert result string.
This turns out to be much more easy to read and avoids internal knowledge within tests.
Bonus: It's trivial to do this twice to test if a VH behaves identical with "template not cached and template cached" (we found and fixed quite a few issues due to this).

PR #982, opened 2024-08-20
[TASK] Use render() instead of renderStatic() in all ViewHelpers

s2b:
Performance tests have shown that the benefits of `renderStatic()` are negligent with current PHP versions. Thus, in preparation of the deprecation of the `CompileWithRenderStatic` trait, we migrate all ViewHelpers to use `render()`.


lolli42 (approved):

lolli42:
wow.

lolli42:
this will allow us to simplify tons and tons of stuff and will get state much more simplified.

PR #987, opened 2024-08-22
[TASK] Annotate correct type hints in AbstractTemplateView (#986)

s2b:
Partials and sections can (of course) return anything, as can any template.


s2b:
before we merge, we should test this against TYPO3 Core Nightly

s2b:
Core CI is green: https://git.typo3.org/typo3/CI/cms/-/pipelines/64132
Consequences for nightly-only are unlikely in my opinion, as only phpstan interprets these comments.

lolli42 (approved):

PR #993, opened 2024-08-26
[DOCS] Improve section about inline syntax

kitzberger:
I've updated the section on inline syntax and added some more (real-life) examples.

And I've added the available casting types to the section on expression nodes.

### DIFF
--- a/Documentation/Usage/Syntax.rst
+++ b/Documentation/Usage/Syntax.rst
@@ -60,25 +60,63 @@ This is called a "chain" and works like this:
 
 .. code-block:: xml
 
-    <f:if condition="{myArray}">
-        <f:for each="{myArray}" as="item">
-            <f:render section="MySection" />
-        </f:for>
-    </f:if>
+    <f:format.trim>
+        <f:replace search="foo" replace="bar">
+            <f:format.case mode="upper">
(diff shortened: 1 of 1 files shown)


Documentation/Usage/Syntax.rst
```diff
@@ -60,26 +60,69 @@ This is called a "chain" and works like this:
...
 .. code-block:: xml
 
-    {f:render(section: 'MySection') -> f:for(each: myArray, as: 'item') -> f:if(condition: myArray)}
+    {myString -> f:format.case(mode:'upper') -> f:replace(search:'foo',replace:'bar') -> f:format.trim()}
 
 The latter syntax is quicker to parse but obviously does not allow for inserting
-any (X)HTML elements around the tags.
+any (X)HTML elements around the single ViewHelper calls.
 
 **It will not always fit your design to use the inline mode but doing so
```
s2b, line 80: I don't think that this advice is relevant anymore. I would suggest to remove this whole paragraph, micro optimizations shouldn't be done in template files.

kitzberger: Agreed.

s2b (approved):

PR #1002, opened 2024-09-09
[FEATURE] Respect ContainerInterface in StandardVariableProvider

o-ba:


composer.json
```diff
@@ -10,7 +10,8 @@
     "license": ["LGPL-3.0-or-later"],
     "require": {
         "php": "^8.2",
-        "ext-mbstring": "*"
+        "ext-mbstring": "*",
+	"psr/container": "^2.0"
```
s2b, line 14: Not sure about this dependency. I don't think it's necessary for the code to work, is it?

s2b:
I like the change overall. Sure, it's special handling for one specific kind of object, but it's a PSR standard after all and it can make life easier without much downside.

composer.json
o-ba: No it's not necessary for the instanceof check. However, phpstan will fail if the interface is not present. Therefore I would suggest to add it as dev dependency.

**Side note:** `psr/container` is already present in fluid because it's a dependency of `symfony/service-contracts`, which is required by `symfony/console`, which is required by `friendsofphp/php-cs-fixer`.

lolli42:
late comment on this: i think its fine to have the psr/container depedency since i'm pretty sure we'll add functionality based on it at least within view helper invoker soon, anyways.

PR #1011, opened 2024-10-18
[FEATURE] Scoped variables

stefanr:
## Short intro

- Extended the `f:variable` view helper to support a new backwards compatible `scope` attribute
  - Value: "global" (default) or "local"
- When `f:variable` is used with `scope="local"` and is placed within a view helper opening a new variable scope via `ScopedVariableProvider`, the variable is only added to the local variable provider
- Added a new `f:scope` view helper to simplify the use of local variables
  - It's actually similar to the `f:alias` view helper but without the need to initially declare variables

## Example

```html
<f:variable name="foo" value="World" />
<f:scope variables="{foo: 'Fluid', bar: 'Lorem'}">
  <f:variable name="baz" scope="local" value="Ipsum" />
  {bar} {baz} {foo}!
</f:scope>
Hello {foo}!
```

Output:

```
Lorem Ipsum Fluid!
Hello World!
```

After the scope `{foo}` is restored, `{bar}` and `{baz}` is removed.

### DIFF
--- a/src/ViewHelpers/ScopeViewHelper.php
+++ b/src/ViewHelpers/ScopeViewHelper.php
@@ -1,0 +1,73 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file belongs to the package "TYPO3 Fluid".
+ * See LICENSE.txt that was shipped with this package.
+ */
+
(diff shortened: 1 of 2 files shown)


mbrodala:
What is the exact usecase which cannot implemented with the current state?

PR #1039, opened 2025-03-19
[DOCS] Remove reference to f:if viewhelper

linawolf:
It doesn't make much sense from f:variable...


mbrodala:
Maybe the intention was to reference the `f:alias` viewhelper? At least this one is mentioned and related functionally.

lolli42:
> Maybe the intention was to reference the `f:alias` viewhelper? At least this one is mentioned and related functionally.

maybe does not help us too much here. i hope it is ok to go with the proposed solution.

lolli42 (approved):

PR #1040, opened 2025-03-19
[BUGFIX] Use correct interlink code for Fluid docs

linawolf:
Like other official manuals the Fluid Explained manual has an explicit shortcode, 'fluid' to be used during linking. 

Using the composer key does not work as this manual is not displayed as extension, so the paths are different.

Additionally use standard shortcodes where possible

Please backport if possible


s2b (approved):

linawolf:
typo3fluid/fluid used in references must be replaced with 'fluid' in the docs for the test to go green again.

s2b (changes requested):
We need to fix the references that already use the existing short code

s2b (approved):

PR #1048, opened 2025-04-14
[BUGIFX] ViewHelpers without then/else return verdict does not work with null values

sascha-egerer:
This is a follow up for #1031 which somehow breaks if the result is a `null` value

### DIFF
--- a/src/Core/ViewHelper/AbstractConditionViewHelper.php
+++ b/src/Core/ViewHelper/AbstractConditionViewHelper.php
@@ -91,7 +91,7 @@ abstract class AbstractConditionViewHelper extends AbstractViewHelper
     protected function renderThenChild()
     {
         // Prefer "then" ViewHelper argument if present
-        if ($this->hasArgument('then')) {
+        if ($this->viewHelperNode instanceof ViewHelperNode && array_key_exists('then', $this->viewHelperNode->getArguments())) {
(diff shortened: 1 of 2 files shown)


src/Core/ViewHelper/AbstractConditionViewHelper.php
```diff
@@ -318,8 +286,58 @@ final public function convert(TemplateCompiler $templateCompiler): array
...
+            } elseif ($arguments[$argumentName] instanceof NodeInterface) {
+                // Argument *is* given to VH and is a node, resolve
+                $converted = $arguments[$argumentName]->convert($templateCompiler);
+                $accumulatedArgumentInitializationCode .= $converted['initialization'];
+
+                if ($argumentName === 'then' || $argumentName === 'else') {
+                    $argumentInitializationCode .= sprintf(
+                        '\'__%s\' => %s,' . chr(10),
+                        $argumentName,
+                        'function () use ($renderingContext) { return ' . $converted['execution'] . ';}',
```
sascha-egerer, line 311: thill will break access to the variable of the parent scope. We could try with arrow function. I'll try to create a testcase to reproduce the error and add my implementation with arrow function

s2b:
Will be applied with #1050 and #1051

PR #1050, opened 2025-04-16
[BUGFIX] Use null instead of empty string as default for condition VHs

s2b:
This partially reverts 3084952: Now all condition-based ViewHelpers use
`null` as default return value instead of `''` because this is more
consistent when `then` or `else` returns `null`, for example if a
non-existent variable is used there:

```
{f:if(condition: true, then: nonExistingVariable)}
```

This should return `null`, but now returns `''`. Note that this change
is only a pre-patch, a follow-up will actually achieve the desired
behavior from that example.

For all other use-cases, this change shouldn't be a problem because
`null` will be converted to `''` anyways.


sascha-egerer:
👍

sbuerk (approved):

Issue #1049, opened 2025-04-14
IfViewhelper returns "true" instead of empty string if condition result is `null`


sascha-egerer:
See #1048 for a failing use case.

PR #1051, opened 2025-04-16
[BUGFIX] Differentiate between null and unset then/else condition branches

s2b:
With Fluid 4.1 (in 96099a6ba60bcb6a73187d289a781903a05882e3) it became possible to
use `f:if` and other condition-based ViewHelpers without `then` or `else`, which
would return the result of the condition as boolean. However, the current
implementation of this feature didn't consider the difference between a ViewHelper
without any `then` or `else` and one with `then` that returns `null`. These two
examples would both return the boolean:

```
{f:if(condition: myCondition, then: myUndefinedVariable)}
{f:if(condition: myCondition)}
```

With this change, the two examples are treated differently. Instead of checking
the arguments of the ViewHelper (which are already merged with the defined
default values), the ViewHelperNode is used to find out if the argument has
actually been specified. This applies both for uncached rendering and compilation,
which is used for cached templates.

Since the implementation in `AbstractConditionViewHelper` became more and more
convoluted, both the rendering and the compilation step have been refactored.
Also, more test cases are added, which make sure that the issues of the current
implementation are still addressed with the refactored code.

Resolves: #1049

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -31,12 +31,6 @@ parameters:
 			path: src/Core/Variables/JSONVariableProvider.php
 
 		-
-			message: '#^Instanceof between TYPO3Fluid\\Fluid\\Core\\Parser\\SyntaxTree\\NodeInterface and TYPO3Fluid\\Fluid\\Core\\Parser\\SyntaxTree\\NodeInterface will always evaluate to true\.$#'
-			identifier: instanceof.alwaysTrue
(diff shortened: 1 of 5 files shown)


s2b:
Apart from composer integrity, TYPO3 Core CI is green: https://git.typo3.org/typo3/CI/cms/-/pipelines/77859

sbuerk (approved):

Issue #1020, opened 2024-11-28
Bug: f:else inside f:else always executed


linawolf:
TYPO3 11.5 - 13.4

I would have expected, the following would give me an error: 
```
<f:if condition="{link.link_icon_custom}">
	<f:then>
		<span class="link-icon"><i class="{link.link_icon_custom}"></i></span>
	</f:then>
	<f:else>
		<f:if condition="{link.link_icon}">
			<span class="link-icon"><i class="{link.link_icon}"></i></span>
		</f:if>
		<f:else>
			<span class="link-icon"><i class="fas fa-chevron-double-right"></i></span>
		</f:else>
	</f:else>
</f:if>
```
However the inner else is not only executed but always executed.

mbrodala:
Notice that the error is not exactly the `f:else` within `f:else` but that the 2nd one is not within the 2nd `f:if`. This would be perfectly valid:

```html
<f:if condition="{link.link_icon_custom}">
	<f:then>
		<span class="link-icon"><i class="{link.link_icon_custom}"></i></span>
	</f:then>
	<f:else>
		<f:if condition="{link.link_icon}">
			<f:then>
				<span class="link-icon"><i class="{link.link_icon}"></i></span>
			</f:then>
			<f:else>
				<span class="link-icon"><i class="fas fa-chevron-double-right"></i></span>
			</f:else>
		</f:if>
	</f:else>
</f:if>
```

(Yes, this could be optimized to use `f:else if="..."`.)

PR #1052, opened 2025-04-16
[BUGFIX] Ignore f:then/f:else outside condition context

s2b:
Before this change, both `f:then` and `f:else` could be used outside of
a condition tag and would then always output the content, regardless
of any condition. Now, nothing is outputted.

In the future, we might consider throwing an exception here to alert the
user about invalid Fluid syntax. However, this would be a breaking change.

Resolves: #1020


src/Core/Parser/SyntaxTree/ViewHelperNode.php
```diff
@@ -82,7 +82,8 @@ public function getViewHelperClassName(): string
 
     /**
      * @internal only needed for compiling templates
-     * @return NodeInterface[]
+     * @return array<NodeInterface|int|float|string|bool> For simple values, an argument might also be scalar
```
sascha-egerer, line 85: You can also use `scalar` here. It's a valid declaration for PHPDoc, PHPStan and PHPStorm. 
```suggestion
     * @return array<NodeInterface|scalar> For simple values, an argument might also be scalar
```

src/Core/ViewHelper/AbstractConditionViewHelper.php
```diff
@@ -90,59 +90,65 @@ public static function verdict(array $arguments, RenderingContextInterface $rend
...
+        }
+
+        // Search for f:then ViewHelper and identify possible f:else to decide how the tag's children should
+        // be interpreted afterwards
         $elseViewHelperEncountered = false;
         foreach ($this->viewHelperNode->getChildNodes() as $childNode) {
             if ($childNode instanceof ViewHelperNode
-                && substr($childNode->getViewHelperClassName(), -14) === 'ThenViewHelper') {
-                $data = $childNode->evaluate($this->renderingContext);
+                && str_ends_with($childNode->getViewHelperClassName(), 'ThenViewHelper')) {
```
sascha-egerer, line 120: That looks error prone to me - but as this not really a change i would not discuss this here. But what is if i have a "HereAndThenViewhelper" that does something total unrelated? Relying on the ending of the viewhelper class does not look very stable to me. Maybe there should be an interface to check for?
Should we create a ticket for this?

src/Core/ViewHelper/AbstractConditionViewHelper.php
```diff
@@ -90,59 +90,65 @@ public static function verdict(array $arguments, RenderingContextInterface $rend
...
+                $data = $childNode->evaluateChildNodes($this->renderingContext);
                 return $data;
             }
             if ($childNode instanceof ViewHelperNode
-                && substr($childNode->getViewHelperClassName(), -14) === 'ElseViewHelper') {
+                && str_ends_with($childNode->getViewHelperClassName(), 'ElseViewHelper')) {
```
sascha-egerer, line 125: Same as above with "ThenViewHelper"

src/Core/ViewHelper/AbstractConditionViewHelper.php
s2b: Yes, please do. I only touched this to get rid of that substr() call.

sascha-egerer: Done 1053

src/Core/Parser/SyntaxTree/ViewHelperNode.php
s2b: adjusted in #1051

PR #1056, opened 2025-04-18
[FEATURE] Introduce new nodeInitializedEvent for ViewHelpers

s2b:
In the current implementation, there already exists a "postParseEvent"
method which allows ViewHelpers to hook into the rendering process.
This is currently used by the f:layout ViewHelper to set the template's
layout. It is also used by the f:section ViewHelper to collect all sections
within a template. Currently, both implementation use the variable
container to transport that information to the compiler and the renderer,
which will then interpret that data.

While it makes sense to provide a generic way to transport data to
allow third-parties to extend Fluid, both cases described above are part
of Fluid's core functionality. Thus, we should aim to append that data
directly to the `ParsingState` instead of using magic strings as variable
names. This patch creates a new event, which has now access to the
whole `ParsingState` object, which can be extended in follow-up
patches to get rid of those magic strings.

Another advantage of direct access to the `ParsingState` is that
ViewHelpers can make more informed decisions during template
parsing. For example, it could be possible to allow certain ViewHelpers
only on the root-level of templates, but not nested within other
ViewHelpers. This will become necessary for the implementation of
upcoming Fluid component features.

For compatibility reasons, the old event keeps getting called by
`TemplateParser`, but emits a deprecation warning. This compatibility
layer will be removed with Fluid v5. ViewHelpers should be migrated
to the new `ViewHelperNodeInitializedEventInterface`.

This patch also adjust existing ViewHelpers to use the new event.

### DIFF
--- a/src/Core/ViewHelper/AbstractViewHelper.php
+++ b/src/Core/ViewHelper/AbstractViewHelper.php
@@ -9,7 +9,6 @@ namespace TYPO3Fluid\Fluid\Core\ViewHelper;
 
 use TYPO3Fluid\Fluid\Core\Compiler\TemplateCompiler;
 use TYPO3Fluid\Fluid\Core\Parser\SyntaxTree\NodeInterface;
-use TYPO3Fluid\Fluid\Core\Parser\SyntaxTree\TextNode;
 use TYPO3Fluid\Fluid\Core\Parser\SyntaxTree\ViewHelperNode;
 use TYPO3Fluid\Fluid\Core\Rendering\RenderingContextInterface;
 use TYPO3Fluid\Fluid\Core\Variables\VariableProviderInterface;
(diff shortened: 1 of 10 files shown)


s2b:
open questions:

* static method vs. object method?
* deprecation timeline ok?

src/Core/ViewHelper/AbstractViewHelper.php
```diff
@@ -558,14 +558,16 @@ public function compile($argumentsName, $closureName, &$initializationPhpCode, V
...
-     * @param ViewHelperNode $node
-     * @param array<string, TextNode> $arguments
-     * @param VariableProviderInterface $variableContainer
+     * @param array<string, NodeInterface> $arguments Unevaluated ViewHelper arguments
      */
-    public static function postParseEvent(ViewHelperNode $node, array $arguments, VariableProviderInterface $variableContainer) {}
+    public static function nodeInitializedEvent(ViewHelperNode $node, array $arguments, ParsingState $parsingState): void
+    {
+        // Call previous event method to not break compatibility with existing ViewHelpers
+        // @todo Trigger deprecation notice with Fluid 5, remove with Fluid 6
```
s2b, line 566: deprecate now, remove with Fluid 5

lolli42 (approved):

PR #1060, opened 2025-05-01
Striptags hrefs in brackets

tgrigoryan234:
https://forge.typo3.org/issues/102651


mbrodala:
This looks confusing but slightly like a conversion from HTML to markdown. IMO something for a separate 3rd party package.

s2b:
I would say that this conversion should happen before stripTags(), not during. So this should be a separate ViewHelper.

I'm not sure if it's generic enough to be part of Fluid Standalone though. In my opinion, it's too opinionated in contrast to other ViewHelpers. But I'm open to other opinions about that.

PR #1063, opened 2025-05-02
[FEATURE] Introduce argument definitions for templates and partials

s2b:
The new `<f:argument>` ViewHelper allows to define an API
for templates, partials and layouts that will be validated when
the template/partial/layout is rendered. For each argument,
a type can be defined. By default, arguments are defined as
non-optional, but they can be defined as optional by using
`optional="{true}"`. Optional arguments can have default
values, which are defined with `default="my default value"`.

To avoid unnecessary complexity, the ViewHelper ensures that
no dynamic content (like variables) can be used in an argument
definition. Also, `<f:argument>` can't be nested into other
ViewHelpers.

Template argument definitions are processed and verified
by the new `StrictArgumentProcessor`, which properly
ensures argument types by casting scalar values and checking
the types more thoroughly than the `LenientArgumentProcessor`.
This means that currently template arguments are validated
in a different way than ViewHelper arguments, however
this will be resolved with Fluid v5.

Currently, there is no way to prohibit additional arguments
that are not defined with the `<f:argument>` ViewHelper,
however this might still be added in the future.


src/Core/Parser/ParsedTemplateInterface.php
```diff
@@ -12,6 +12,7 @@
 use TYPO3Fluid\Fluid\Core\Parser\SyntaxTree\NodeInterface;
 use TYPO3Fluid\Fluid\Core\Rendering\RenderingContextInterface;
 use TYPO3Fluid\Fluid\Core\Variables\VariableProviderInterface;
+use TYPO3Fluid\Fluid\Core\ViewHelper\ArgumentDefinition;
 
 /**
  * This interface is returned by \TYPO3Fluid\Fluid\Core\Parser\TemplateParser->parse()
```
s2b, line 18: mark as internal

Documentation/ViewHelpers/Fluid/Argument.rst
```diff
@@ -0,0 +1,21 @@
...
+
+..  include:: /Includes.rst.txt
+
+..  _typo3fluid-fluid-argument:
+
+==================================
+Argument ViewHelper `<f:argument>`
+==================================
+
+.. ..  note::
```
s2b, line 15: Needs to be un-commented once the Fluid release is merged into TYPO3 Core.

src/View/AbstractTemplateView.php
```diff
@@ -144,6 +147,16 @@ public function render($actionName = null)
...
             $this->startRendering(self::RENDERING_TEMPLATE, $parsedTemplate, $this->baseRenderingContext);
+            try {
+                // @todo make argument processor configurable with Fluid v5
+                $this->processAndValidateTemplateVariables(
+                    $parsedTemplate,
+                    $this->baseRenderingContext->getVariableProvider(),
+                    new StrictArgumentProcessor(),
+                );
+            } catch (Exception $validationError) {
+                return $renderingContext->getErrorHandler()->handleViewError($validationError);
```
s2b, line 158: baseRenderingContext

src/View/AbstractTemplateView.php
```diff
@@ -159,6 +172,16 @@ function ($parent, TemplatePaths $paths) use ($layoutName) {
...
             $this->startRendering(self::RENDERING_LAYOUT, $parsedTemplate, $this->baseRenderingContext);
+            try {
+                // @todo make argument processor configurable with Fluid v5
+                $this->processAndValidateTemplateVariables(
+                    $parsedLayout,
+                    $this->baseRenderingContext->getVariableProvider(),
+                    new StrictArgumentProcessor(),
+                );
+            } catch (Exception $validationError) {
+                return $renderingContext->getErrorHandler()->handleViewError($validationError);
```
s2b, line 183: baseRenderingContext

tests/Functional/ViewHelpers/ArgumentViewHelperTest.php
```diff
@@ -0,0 +1,185 @@
...
+        $view = new TemplateView();
+        $view->assignMultiple($variables);
+        $view->getRenderingContext()->setCache(self::$cache);
+        $view->getRenderingContext()->getTemplatePaths()->setLayoutRootPaths([$layoutRootPath]);
+        $view->getRenderingContext()->getTemplatePaths()->setTemplateSource($templateSource);
+        $view->render();
+    }
+
+    #[Test]
+    public function templateArgumentsAreIgnoredWithLayout(): void
```
s2b, line 166: add test with section

lolli42 (approved):

PR #1077, opened 2025-05-12
[!!!][TASK] Remove namespaces from compiled templates

s2b:


s2b:
TYPO3 Core CI run: https://review.typo3.org/c/Packages/TYPO3.CMS/+/89360

s2b:
We might keep the namespace in the file, since #1080 already removes the global state from the cache.

PR #1090, opened 2025-05-23
[TASK] Establish proper API to set template layout

s2b:
The `ParsingState` now provides a proper method to set the
layout of a template during parsing/compilation. Previously, the
magic variable name `layoutName` has been used for that
purpose, but with the new `nodeInitializedEvent` we can now
access the `ParsingState` object directly.

The usage of the old variable name has been deprecated and
will no longer be evaluated with Fluid v5. Its main usage besides
the `<f:layout>` ViewHelper (see above) has been the cache
warmup command, which will receive a more straightforward
implementation with Fluid v5.


s2b:
TYPO3 Core CI is green (apart from composer integrity): https://review.typo3.org/c/Packages/TYPO3.CMS/+/89518

PR #1091, opened 2025-05-27
[TASK] Deprecate childNodes property for ViewHelpers

s2b:
Currently, the raw parser nodes are provided twice to each
ViewHelper instance: Once with `setViewHelperNode()` and
once with `setChildNodes()`. The latter is now deprecated,
together with the property `$childNodes`, since it can easily
be subsituted like this within a ViewHelper:

before:

```php
$this->childNodes
```

after:

```php
$this->viewHelperNode->getChildNodes()
```

### DIFF
--- a/Documentation/Changelog/4.x.rst
+++ b/Documentation/Changelog/4.x.rst
@@ -6,6 +6,17 @@
 Changelog 4.x
 =============
 
+4.3
+---
+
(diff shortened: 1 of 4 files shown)


s2b:
Associated TYPO3 Core patch: https://review.typo3.org/c/Packages/TYPO3.CMS/+/89571

lolli42 (approved):

Issue #1109, opened 2025-06-16
Named Slots


s2b:
As an extension to components, it should be possible to provide multiple slots to one component: Named Slots.

```xml
<my:component>
  <f:fragment name="slot1">some <b>content</b></f:fragment>
  <f:fragment name="slot2">some <b>other</b> content</f:fragment>
</my:component>
```

And within the component:

```xml
<f:slot name="slot1" />
<f:slot name="slot2" />
```

We might also consider some other additions:

* write available slot names to the cache so that they can be fetched easily
* allow default values for slots
* provide arguments to slots to separate integration and component

PR #1118, opened 2025-06-25
[FEATURE] Named Slots

s2b:
This patch enables Fluid's components to accept multiple slots. Each
slot is identified by a `name` argument. The corresponding
`<f:fragment>` ViewHelper can then be used to fill the slots of a
component.

The available slots of a component are now also exposed properly
via the `ComponentDefinition` DTO. Unfortunately, the `TemplateParser`
fails to propagate available slots of the current templates consistently to
the `ParsingState`, which currently leads to ugly state merging within the
parser. Hopefully, this issue can be addressed in the future to remove this
workaround, which has been documented in the code.

For this first implementation, we limit the usage if the new
`<f:fragment>` ViewHelper to the context of components. However, we
might expand this ViewHelper to other areas in Fluid, such as a way to
partially extend/overwrite templates.

Resolves: #1109


Kanti:
Possible Validations:

1. If `f:fragment` is used but `f:slot` is not, it should trigger an Exception and notify the Developer that the slot is not defined.
2. If `f:slot` is used but no matching `f:fragment` is used it should also trigger an Exception. (if the slot is not optional)

lolli42 (approved):

PR #1140, opened 2025-08-01
[!!!][TASK] Add final and type checks to SyntaxTree nodes

s2b:


src/Core/Parser/SyntaxTree/Expression/AbstractExpressionNode.php
```diff
@@ -52,7 +51,7 @@ public function __construct(string $expression, array $matches)
      * @param RenderingContextInterface $renderingContext
      * @return string the text stored in this node/subtree.
      */
-    public function evaluate(RenderingContextInterface $renderingContext)
+    public function evaluate(RenderingContextInterface $renderingContext): mixed
```
sbuerk, line 54: Interesting ... phpdocblock says `@return string` ... that does not match ...

src/Core/Parser/SyntaxTree/Expression/AbstractExpressionNode.php
```diff
@@ -72,7 +71,7 @@ public function evaluate(RenderingContextInterface $renderingContext)
      * instance - and the RenderingContext and other APIs
      * can be accessed via the TemplateCompiler.
      */
-    public function compile(TemplateCompiler $templateCompiler)
+    public function compile(TemplateCompiler $templateCompiler): array
```
sbuerk, line 74: I'd say we can add a phpdocblock return type with an array-shape 

```
* @return array{
*   initialization: string,
*   execution: string,
* }
```

src/Core/Parser/SyntaxTree/Expression/AbstractExpressionNode.php
s2b: The `CastingExpressionNode` can return various types, so `mixed` is correct here. I removed the phpdoc.

PR #1150, opened 2025-08-02
[TASK] Treat f:argument with default value as optional

### LINKED ISSUE #1149: Automatically treat arguments with `default` value as optional
Hey,

I’d like to propose an improvement to the `f:argument` ViewHelper behavior, specifically around the use of the `default` and `optional` attributes.

Currently, when using f:argument with a default value, we must explicitly set both:

```html
<f:argument name="open" type="boolean" optional="{true}" default="{false}" />
```

If the `optional` attribute is omitted, even when a default is provided, the ViewHelper throws an exception if the argument is missing. This feels slightly redundant, since a `default` value generally implies that the argument is optional.

**Proposal**

- If a `default` is provided and `optional` is omitted, the argument could be treated as optional automatically.
- Additionally, it might be helpful to throw an exception when `optional="{false}"` is explicitly set alongside a `default` value, as that combination seems contradictory.

I think that would improve DX and reduce boilerplate, without changing existing expected behavior.

Happy to hear your thoughts and provide a PR, if so.

Joost

jramke:
Closes #1149

### DIFF
--- a/src/ViewHelpers/ArgumentViewHelper.php
+++ b/src/ViewHelpers/ArgumentViewHelper.php
@@ -138,7 +138,7 @@ final class ArgumentViewHelper extends AbstractViewHelper implements ViewHelperN
             $argumentName,
             (string)$evaluatedArguments['type'],
             array_key_exists('description', $evaluatedArguments) ? (string)$evaluatedArguments['description'] : '',
-            $required,
+            !$optional,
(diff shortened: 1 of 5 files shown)


jramke:
Updated the comma for php cs fixer

s2b (changes requested):

s2b:
And can you add a test case that covers this?

PR #1155, opened 2025-08-02
[BUGFIX] allow usage of interfaces in StrictArgumentProcessor

Kanti:
Got this error:

> The argument "file" for template "Default_action_ContentElements_rte_only_templates_frontend_3a0ee286e27afc95" is registered with type "TYPO3\CMS\Core\Resource\FileInterface", but the provided value is of type "TYPO3\CMS\Core\Resource\File".

This fix makes it possible to use interfaces as typed arguments for components.


s2b:
Nice catch, thanks!

PR #1161, opened 2025-08-04
[TASK] Allow argument types to be interfaces

### LINKED ISSUE #1160: Argument `type` in `f:argument` cannot be interface
When I want to use this:

```
<f:argument name="file" type="\TYPO3\CMS\Core\Resource\FileInterface" optional="{true}" />
```

I get that:

> (1/1) #1746637333 TYPO3Fluid\Fluid\View\Exception
> The argument "file" for template "Default_action_DownloadItem_DownloadItem_1725a6bf79fdf9cb" is registered with type "TYPO3\CMS\Core\Resource\FileInterface", but the provided value is of type "TYPO3\CMS\Core\Resource\File".

Screenshot of the error message

koehnlein:
This change does not only allow `$type` to be a class but also an interface.

Resolves #1160

### DIFF
--- a/src/Core/ViewHelper/StrictArgumentProcessor.php
+++ b/src/Core/ViewHelper/StrictArgumentProcessor.php
@@ -101,7 +101,7 @@ final readonly class StrictArgumentProcessor implements ArgumentProcessorInterfa
             }
             return true;
         }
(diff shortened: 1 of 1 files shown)


s2b:
This has already been fixed in a previous commit: #1156

Issue #1124, opened 2025-06-26
f:length ViewHelper


s2b:
A new `<f:length>` ViewHelper should be added to Fluid, which calculates the length of a string.

Arguments:

* string $value

See also [the php mb_strlen() function](https://www.php.net/manual/en/function.mb-strlen.php).

Examples:

```xml
<f:variable name="length" value="{f:length(value: 'my string')}" />
<f:variable name="length" value="{myVariable -> f:length()}" />
```

PR #1169, opened 2025-08-09
[FEATURE] LengthViewHelper returns the length of a given string (#1124)

staatzstreich:
A new `<f:length>` ViewHelper should be added to Fluid, which calculates the length of a string.

Arguments:

string $value
See also [the php mb_strlen() function](https://www.php.net/manual/en/function.mb-strlen.php).

Examples:

```
<f:variable name="length" value="{f:length(value: 'my string')}" />
<f:variable name="length" value="{myVariable -> f:length()}" />
```


src/ViewHelpers/LengthViewHelper.php
```diff
@@ -0,0 +1,70 @@
...
+ * Output::
+ *
+ *      13
+ *
+ * Inline notation
+ * ---------------
+ *
+ * ::
+ *
+ *      {f:length(value: 'Hello, World!')}
```
s2b, line 38: Could you add an example for the encoding argument?

tests/Functional/ViewHelpers/LengthViewHelperTest.php
```diff
@@ -0,0 +1,110 @@
...
+        $view->getRenderingContext()->getTemplatePaths()->setTemplateSource($src);
+        $view->assignMultiple($arguments);
+        self::assertSame($expectation, $view->render());
+    }
+
+    public static function renderInvalidDataProvider(): iterable
+    {
+        yield 'invalid value' => [
+            'arguments' => [
+                'value' => 13,
```
s2b, line 73: I would argue that this should be a valid use case:

```
<f:length value="13" />

Result: 2
```

Maybe use `is_scalar()` in your code above and then cast to string?

s2b (changes requested):

src/ViewHelpers/LengthViewHelper.php
staatzstreich: Should there also be a link to the possible encoding from php documentation?

s2b (approved):

s2b:
Thank you for your contribution!

Issue #1125, opened 2025-06-26
f:contains ViewHelper


s2b:
A new `<f:contains>` ViewHelper should be added to Fluid, which checks if a provided value either exists in the provided array or in the provided string. If both the array and the string are provided, an exception should be thrown. This ViewHelper should extend the `AbstractConditionViewHelper`.

Arguments:

* mixed $value (needs to be scalar if f:contains is used with the string argument)
* array $array
* string $string

See also [the php in_array()](https://www.php.net/manual/de/function.in-array.php) and [the php str_contains()](https://www.php.net/manual/en/function.str-contains.php) functions.

Examples:

```xml
<f:contains value="foo" array="{myArray}">

</f:contains>
<f:contains value="foo" string="{myString}">

</f:contains>

<f:variable name="result" value="{f:contains(value: 'foo', array: myArray)}" />
```

staatzstreich:
The "in_array" function also has a parameter "strict" ... should this also be usable?

PR #1193, opened 2025-08-14
[FEATURE] <f:contains> ViewHelper

staatzstreich:
This patch adds a new `<f:contains>` ViewHelper to Fluid, which allows
to check if a provided string or array contains a specified value.
Depending on the input, this mimicks PHP's `in_array()` or `str_contains()`.

The initial version of this patch is intended to be backported to Fluid 4.
In a follow-up, the code and argument types can be improved for Fluid 5.

Resolves: #1125


mbrodala:
I wonder if we should consider extending conditions instead, e.g. using Symfony EL. Then we'd not need to add a viewhelper for each and every bit of functionality.

s2b:
@mbrodala I agree that we should improve conditions overall. However, in the short term it's probably wise to add ViewHelpers for the most common use cases nonetheless. Also, the following syntax is already possible:

```xml
<f:if condition="{f:contains(value: 'foo', array: myArray)} && ...">
```

s2b:
The more I think about it, the more I struggle with the VH API. I'm open to other suggestions, but these come to mind:

* Allow both array and string as one argument and differentiate in the VH based on the type
* Split into two: f:contains for strings and f:in for arrays

What do you think?

brotkrueml:
> * Split into two: f:contains for strings and f:in for arrays

IMHO there should be two dedicated view helpers as you suggested to handle the different types. Additionally, for an f:in view helper (better f:inArray as the purpose is clearer) we should think about the possibility for checking the type (like in_array in PHP with the "strict" argument):

    <f:inArray value="some-value" array="{someArray}" strict="{true}">

With "strict" be false by default (like in PHP). Strictness might be difficult when setting a value directly in Fluid:

    <f:inArray value="42" array="{someArray}" strict="{true}">

Here the value is a string while the array values might be ints. But that would work:

    <f:inArray value="{someValue}" array="{someArray}" strict="{true}">

if {someValue} is passed to Fluid and is really an int.

s2b:
I've adjusted the ViewHelper and documentation. WDYT?

PR #1205, opened 2025-08-15
[TASK] Avoid context-dependent prefixes for cache files

s2b:
In order to implement a reasonable cache warmup, the template cache needs to be independent of the context where the template will be used later in the project. For cache files, this means that they cannot limit how a template will be used later on (partial, layout or template).

This patch removes the context-dependent prefixes for cache file names to achieve this.

### DIFF
--- a/src/View/TemplatePaths.php
+++ b/src/View/TemplatePaths.php
@@ -338,9 +338,7 @@ class TemplatePaths
     public function getLayoutIdentifier(string $layoutName = 'Default'): string
     {
         $filePathAndFilename = $this->getLayoutPathAndFilename($layoutName);
-        $layoutName = str_replace('.', '_', $layoutName);
-        $prefix = 'layout_' . $layoutName . '_' . $this->getFormat();
-        return $this->createIdentifierForFile($filePathAndFilename, $prefix);
+        return $this->createIdentifierForFile($filePathAndFilename);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


s2b:
I've validated with the main branch of TYPO3 Core that this works as expected. Test setup:

* flush all caches
* login to backend, click through entry pages of all backend modules
* store snapshot of Fluid cache folder (I've used git for this)
* flush all caches
* `composer req typo3fluid/fluid:dev-task/cacheFiles`
* repeat same action
* compare cache folder with previous snapshot

This resulted in mostly moved files where only the class name changed, e. g.:

```diff
-class BackendUser_action_list_47e99727da31866b extends \TYPO3Fluid\Fluid\Core\Compiler\AbstractCompiledTemplate {
+class List_html_47e99727da31866b extends \TYPO3Fluid\Fluid\Core\Compiler\AbstractCompiledTemplate {
```

There were also a couple of files that weren't recreated at all. Further investigation showed that these were actually duplicate cache entries that don't exist anymore with the new context-independent naming scheme, for example:

Old:

* Default_action_BackendUser_List_47e99727da31866b
* BackendUser_action_list_47e99727da31866b

New:

* List_html_47e99727da31866b

Note the same hash at the end. Because controller and action are no longer part of the filename, one cache entry is enough to cover both use cases.

Issue #1166, opened 2025-08-09
Add {array} to give the possibility to get an empty array


garbast:
In a template I want to start a variable with an empty array to fill it later on with f:merge.

```
<f:variable name="filteredItems" value="{array}"/>
<f:for each="{allItems}" as="item">
    <f:if condition="{item.uid} > 10">
      <f:variable name="filteredItems" value="{f:merge(a: filteredItems, b: {item.uid: item})"/>
    </f:if>
</f:for>
```

mbrodala:
Native workaround for now:

```xml
<f:split value="" separator="," limit="-1"/>
```

This will produce an empty array. Notice that `limit` is the important key here, without you'd get an array like `[0 => '']` instead.

In your particular case:

```xml
<f:variable name="filteredItems" value="{f:split(value: '', separator: ',', limit: '-1')}"/>
```

Issue #1041, opened 2025-03-24
Cannot pass \n as seperator to split viewhelper


benjaminkott:
bodytext
```
 Lorem ipsum dolor sit amet, consectetur adipiscing elit
 Maecenas blandit mauris vitae vehicula volutpat
 Duis feugiat risus efficitur ligula condimentum tempus
 Curabitur eget quam in ipsum tincidunt pulvinar ut a eros
 Nam placerat erat sit amet euismod fermentum
```

fluid
```
{bodytext -> f:split(separator: '\n') -> f:variable(name: 'rows')}
```

benjaminkott:
that's the current workaround ... it hurts

mbrodala:
Notice that this should be handled very carefully to not have unexpected side-effects in cases `\n` is literally used, e.g. in a PHP FQCN like `acme\notinventedhere\solution`.

s2b:
I was also thinking about cases like that. We probably need to introduce some kind of constant. It might make sense to define a dedicated namespace for things like this to avoid breaking changes in the future.

So something like this:

```
{_LF}
{_TAB}
...
```

PR #1225, opened 2025-08-31
[!!!][TASK] Disallow variables that start with a "_"

s2b:
[!!!][TASK] Disallow variables that start with a "_"

In preparation for the possible introduction of more internal
constants in Fluid, custom variable names that start with an
underscore are no longer allowed.

For Fluid 4, a deprecation will be introduced in a follow-up patch
to prepare template authors for the upcoming change.

Related: #1166
Related: #1041


s2b:
Core CI is green with this change, so I think we can go ahead with it.

PR #1233, opened 2025-09-11
[!!!][TASK] Remove tag attributes if set to null

billdagou:
For compatibility reasons, Fluid interpreted tag attributes that were set
to `null` as empty strings in `TagBuilder`, which resulted in an empty
tag attribute (`data-attribute=""`). With this patch, that behavior is
changed: `null` values now lead to the attribute being omitted from the
tag, just like `false` does.

Before:

```xml
<my:tagBased data-attribute="{null}" />
Result: <div data-attribute=""></div>
```

After:

```xml
<my:tagBased data-attribute="{null}" />
Result: <div></div>
```


s2b:
Can you adjust the test case as well?

s2b:
thx. 8.5 is a general issue, so your PR is technically green now. ;)

billdagou:
Thank you.

PR #1234, opened 2025-09-11
[TASK] Use modern HTML syntax for boolean attributes

billdagou:
Fluid used the XHTML syntax to represent boolean values. This change adjusts
the `TagBuilder` to use the modern syntax. This change is considered non-breaking
because it results in the same interpretation of the browser. However, it will
not be backported to Fluid v4 to be safe.

Before:

```xml
<my:tagBased required="{true}" />
Result: <input required="required" />
```

After:

```xml
<my:tagBased required="{true}" />
Result: <input required />
```

### DIFF
--- a/src/Core/ViewHelper/TagBuilder.php
+++ b/src/Core/ViewHelper/TagBuilder.php
@@ -43,8 +43,7 @@ class TagBuilder
 
     protected bool $ignoreEmptyAttributes = false;
     /**
-     * I'd prefer to set the default value to FALSE.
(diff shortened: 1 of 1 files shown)


s2b:
What was the reason for the additional `requireAttributeValues` configuration? From my point of view, this isn't necessary, we can just switch to the alternative syntax.

billdagou:
To switching between `<input required />` and `<input required="required" />`, like for `config.doctype = xhtml_*` or `config.doctype = html5` in TYPO3.

Issue #1012, opened 2024-11-11
Don't remove CDATA sections from templates


s2b:
Since `<f:comment>` now works correctly for invalid Fluid, we should consider not removing `CDATA` sections from templates anymore.

Issue #808, opened 2023-06-29
Template parser issue when using curly braces for inline js or css


stefanr:
Sometimes when using curly braces for inline javascript or css styles inside a fluid template the parser breaks and does not further resolve variables from this point up until some other view helper is used via xml syntax or the scope (section, partial, ...) is closed.

I didn't look into the implementation in detail so it currently seems a little "random" at which point the parser stops working.

Example:

```
<f:variable name="test" value="foo" />
<div
  x-data="{
  test: null,
  init() {
    test = 'foo';
  }
  }"
>
  {test}
</div>
```

This correctly (or let's say as expected), gives: `foo`

This on the other hand (focus on the indentation):

```
<f:variable name="test" value="foo" />
<div
    x-data="{
      test: null,
      init() {
        test = 'foo';
      }
    }"
>
  {test}
</div>
```

... renders as: `{test}`

Placing a view helper before the variable output seems to fix wrong parser state:

```
<f:variable name="test" value="foo" />
<div
    x-data="{
      test: null,
      init() {
        test = 'foo';
      }
    }"
>
  <f:variable name="void" value="" />
  {test}
</div>
```

... renders again: `foo`

Any ideas? This issue is kind of a problem as one might not expect to destroy the following template code by changing some attributes that also render completely fine in the frontend.

Wrapping the problematic blocks like so `x-data="<f:format.raw>{ ... }</f:format.raw>"` is the current work around to fix this issue.

sbuerk:
@stefanr Thank you for the issue report. However, I would say, this is not a bug. It is how it is worked and documentented. 

`{}` are fluid specific chars which marks the start end end of variables or inline code. So they have a special meaning.

You are trying to mix this some how. There are multiple ways to mitigate this,
which are already documentend and explained, at least for the TYPO3 world here:

https://docs.typo3.org/m/typo3/guide-extbasefluid/main/en-us/Fluid/ThingsToKnow/JsAndInline.html

Granted, the context describes it for inline javascript, but this also counts for "json" or "JavaScript object like" markup creation.

It's a casual issue if you deal with kind of a parsed language, that special meaning chars needs to be escaped in the one or other way.

Not tested, but I guess following "should" work.

```html
<f:variable name="test" value="foo" />
<f:format.cdata>
<div
  x-data="{
    test: null,
    init() {
      test = 'foo';
    }
  }"
>
</f:format.cdata>
  {test}
</div>
```

[...]

PR #1246, opened 2025-10-11
[!!!][FEATURE] Alternative Fluid syntax in CDATA

s2b:
Fluid has so far removed any CDATA sections from templates. This has
been introduced with 978071f, without any clear indication why this
was done. This means that Fluid templates either expect CDATA sections
to be removed or avoid them altogether since they don't do anything.
This gives us the opportunity to not only re-add those sections to
Fluid templates (which is a breaking change), but to also introduce
new syntax within those section.

One common issue with the Fluid parser has always been that the
variable/inline syntax collides with both CSS and JS, which both use
curly braces quite extensively. In most cases, this leads to the
Fluid parser trapping over a curly brace that's not intended for Fluid
and then stops interpreting valid Fluid syntax after that.

This patch extends the Fluid parser to use alternative variable and
inline syntax within CDATA sections in the template. Both ViewHelper
tag syntax and the normal curly braces syntax no longer work there,
instead three curly braces `{{{ }}}` can be used to access Fluid
variables, call ViewHelpers and use expressions. The CDATA keyword
itself is removed from the output.

Resolves: #1012
Resolves: #808

### DIFF
--- a/Documentation/Changelog/5.x.rst
+++ b/Documentation/Changelog/5.x.rst
@@ -90,4 +90,5 @@ Changelog 5.x
   results in an exception. Note that this doesn't affect array keys or object property names, only the name of the variable itself.
 * Deprecation: Class :php:`TYPO3Fluid\Fluid\Core\ViewHelper\LenientArgumentProcessor`
   is no longer being used by Fluid v5 and will be removed with Fluid v6.
-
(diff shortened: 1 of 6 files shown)


s2b:
TODO: We may declare this as experimental for now.

PR #1248, opened 2025-10-15
Explain how to check versions

linawolf:


mbrodala:
Strictly speaking we have quite a few `versionadded` blocks throughout the docs. Do you think it would make sense to mention this everywhere? It should be compact though (e.g. a link) so it won't take too much effort for future changes.

linawolf:
sounds like a good idea

PR #1274, opened 2025-12-10
[DOCS] Document `<f:argument>` does not work when rendering sections

### LINKED ISSUE #1273: `<f:argument>` not evaluated when partial is rendered with `section=…`
Assume there is a partial named `Argu`:

```html
<f:argument name="property" type="string" />

<f:section name="main"><p>Property in section: {property}</p></f:section>
<p>Property outside section: {property}</p>
```

When rendering this partial the `<f:argument>` is not always evaluated:

1. Check triggered and validation passed:

```html
<f:render partial="Argu" arguments="{property:'age'}" />
```

2. Check triggered and validation failed (correctly):

```html
<f:render partial="Argu" arguments="{}" />
```

3. Check not triggered - no validation

```html
<f:render partial="Argu" section="main" arguments="{}" />
```

## Expected behavior:

I expect example 3. to throw an exception, which informs me about the missing argument.

## Environment:

| Package           | Version |
| ----------------- | ------- |
| typo3/cms-core    | 13.4.21 |
| typo3/cms-fluid   | 13.4.21 |
| typo3fluid/fluid | 4.5.0   |

cundd:
Fix #1273


src/ViewHelpers/ArgumentViewHelper.php
```diff
@@ -34,6 +34,9 @@
  * in any of its arguments is not possible (e. g. you can't define an argument name
  * by using a variable).
  *
+ * Currently rendering of specific sections will not validate argument constraints.
```
sorenmalling: Avoid using "currently" - it implies that something might change.. but you don't know :)

Keep it clean:

`Rendering of specific sections will not validate argument constraints. 
Comment`

PR #1288, opened 2026-01-15
[FEATURE] UnsafeHTML interface to allow passing variables to template without being escaped

Kanti:
Fluid escapes variable output by default to prevent XSS. In some cases,
however, values originate from a trusted/sanitized source (e.g. HTML
generated from a sanitizer, CMS RTE output, a Markdown renderer with
a strict allow-list, …) and should be rendered as HTML without forcing
template authors to opt out of escaping via `f:format.raw` or similar.

This change introduces a marker interface
`TYPO3Fluid\Fluid\Core\Parser\UnsafeHTML` for values that should be
rendered unescaped. Any object implementing this interface will
bypass Fluid’s escaping and will be output as-is via `__toString()`.
For now, this interface is marked as `@internal` to be able to make
adjustments within the Fluid 5 branch, if necessary.

A small helper value object `UnsafeHTMLString` is included for convenience.

Example (PHP):
````php
use TYPO3Fluid\Fluid\Core\Parser\UnsafeHTMLString;

// $safeHtml must already be sanitized/escaped appropriately
$view->assign('content', new UnsafeHTMLString($safeHtml));
````
Template:
````html
{content}
````
Technical notes:

- `EscapingNode` now detects `UnsafeHTML` and returns the value unescaped.
- The compiled escaping closure generated by the `TemplateCompiler`
  includes the same check, so compiled templates behave identically.
- Boolean expression evaluation unwraps `UnsafeHTML` to a string first,
  so empty HTML values behave like empty strings in conditions
  (e.g. `{content}` is false when it’s `''`).

Tests/examples have been adjusted accordingly.

### DIFF
--- a/src/Core/Parser/SyntaxTree/EscapingNode.php
+++ b/src/Core/Parser/SyntaxTree/EscapingNode.php
@@ -10,6 +10,7 @@ declare(strict_types=1);
 namespace TYPO3Fluid\Fluid\Core\Parser\SyntaxTree;
 
 use TYPO3Fluid\Fluid\Core\Compiler\TemplateCompiler;
+use TYPO3Fluid\Fluid\Core\Parser\UnsafeHTML;
 use TYPO3Fluid\Fluid\Core\Rendering\RenderingContextInterface;
 
 /**
@@ -38,6 +39,9 @@ final class EscapingNode extends AbstractNode
     public function evaluate(RenderingContextInterface $renderingContext): mixed
     {
         $evaluated = $this->node->evaluate($renderingContext);
+        if ($evaluated instanceof UnsafeHTML) {
+            return (string)$evaluated;
+        }
         if (is_string($evaluated) || (is_object($evaluated) && method_exists($evaluated, '__toString'))) {
             return htmlspecialchars((string)$evaluated, ENT_QUOTES);
         }
@@ -65,6 +69,7 @@ final class EscapingNode extends AbstractNode
         if ($configuration['execution'] !== '\'\'') {
             $configuration['execution'] = sprintf(
                 'call_user_func_array( function ($var) { '
+                . 'if ($var instanceof ' . UnsafeHTML::class . ') { return (string)$var; }'
                 . 'return (is_string($var) || (is_object($var) && method_exists($var, \'__toString\')) '
                 . '? htmlspecialchars((string) $var, ENT_QUOTES) : $var); }, [%s])',
                 $configuration['execution'],
--- a/src/Core/Parser/UnsafeHTML.php
+++ b/src/Core/Parser/UnsafeHTML.php
@@ -1,0 +1,23 @@
+<?php
+
+declare(strict_types=1);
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

File: tests/Unit/Core/Parser/BooleanParserTest.php [hunk 1]
```diff
@@ -105,6 +106,12 @@ public static function getSomeEvaluationTestValues(): array
             ['{foo} == FALSE', true, ['foo' => false]],
             ['!{foo}', true, ['foo' => false]],
 
+            ['{foo}', false, ['foo' => new UnsafeHTMLString('')]],
+            ["{foo} == ''", true, ['foo' => new UnsafeHTMLString('')]],
+            ['{foo}', true, ['foo' => new UnsafeHTMLString('test')]],
+            ['{foo} == FALSE', false, ['foo' => new UnsafeHTMLString('test')]],
+            ["{foo} == 'test'", true, ['foo' => new UnsafeHTMLString('test')]],
+
```
s2b: let's add some more test cases, e. g. to document the `===` behavior or comparison with `{false}`. Also, I'd like to have at least one higher-level example in a `<f:if>` test.

src/Core/Parser/SyntaxTree/EscapingNode.php
```diff
@@ -38,6 +39,9 @@ public function __construct(NodeInterface $node)
     public function evaluate(RenderingContextInterface $renderingContext): mixed
     {
         $evaluated = $this->node->evaluate($renderingContext);
+        if ($evaluated instanceof UnsafeHTML) {
+            return $evaluated->__toString();
```
s2b: let's switch all castings to `(string)` to stay consistent with the existing code.

src/Core/ViewHelper/UnsafeHTML.php
```diff
@@ -0,0 +1,28 @@
...
+
+use Stringable;
+
+/**
+ * Interface for values that are considered safe HTML
+ * and will not be escaped by the Fluid rendering engine.
+ *
+ * Use with caution and ensure the HTML has already been sanitized, as it will not be escaped by the Fluid rendering engine.
+ *
+ * @api
```
s2b: Let's stick to `@internal` at the moment.

s2b (changes requested):

src/Core/Parser/UnsafeHTML.php
```diff
@@ -0,0 +1,28 @@
...
+ * Use with caution and ensure the HTML has already been sanitized, as it will not be escaped by the Fluid rendering engine.
+ *
+ * @internal
+ */
+interface UnsafeHTML extends Stringable
+{
+    /**
+     * @return string returns HTML that has already been sanitized and will not be escaped by the Fluid rendering engine.
+     */
+    public function __toString(): string;
```
ohader: I think this definition can be omitted, as it is defined in `\Stringable` already

PR #1302, opened 2026-02-06
[BUGFIX] make ContentArgumentName non required argument

Kanti:
if you have a ViewHelper that has a required argument like 'record'.
And you use 
````php
    public function getContentArgumentName(): string
    {
        return 'record';
    }
````

and then you want to use:

````html
{record -> my:vh()}
````

you currently get this error:
````
Fluid parse error in template template_..., line 19 at character 1. Error: Required argument "record" was not supplied. (error code 1237823699). Template source chunk: ...
````


s2b:
As discussed: I'd rather not do it like this and instead fix it properly with Fluid 6, see #1303.

PR #1314, opened 2026-02-17
[FEATURE] Annotations API

s2b:
This patch introduces new data structures that represent annotations
for ViewHelpers/Components or their individual arguments. The purpose
is to be able to attach arbitrary information to definitions of
ViewHelpers/Components or arguments, either to pass them along as-is
or to react to them within Fluid. There are places within Fluid where
something like this already exists (e. g. XSD schema), and there are
multiple additional use cases for this, both by Fluid itself and by
third-party libraries that extend Fluid, for example:

* Marking ViewHelpers/Components or individual arguments as deprecated
* Collecting information for documentation from PhpDoc comments
* Attaching additional validations/constraints to component arguments

In this first step, only the low-level API is provided, which
consists of interfaces and a generic implementation. Since this concept
is new to Fluid, the API is still classified as internal and might
change with future revisions.

Since the `ArgumentDefinition` needs to be written to Fluid's template
cache files, each annotation needs to implement the `compile()`
method, which returns the PHP code that re-creates the annotation object
when a template is read from cache.

### DIFF
--- a/src/Core/Component/ComponentDefinition.php
+++ b/src/Core/Component/ComponentDefinition.php
@@ -9,6 +9,7 @@ declare(strict_types=1);
 
 namespace TYPO3Fluid\Fluid\Core\Component;
 
+use TYPO3Fluid\Fluid\Core\Definition\Annotation\ViewHelperAnnotationInterface;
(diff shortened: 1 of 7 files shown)


s2b:
@jramke Could you test if this would cover your use case for fluid-primitives?

PR #1325, opened 2026-02-20
[BUGFIX] Address inconsistencies in object comparisons (#1320)

s2b:
Fluid's `BooleanParser` attempts to cover a specific edge case: If
two objects should be compared, according to the code it should always
be a strict comparison. However, the condition that should have
accomplished this depends on the variable content during the first
run. If a template is compiled (which is almost always the case), the
comparison is still dependent on the variable content during initial
template compilation.

Above all, this special handling doesn't really make sense, since strict
comparison operators are available as well.

This patch adds test coverage for object comparisons on multiple levels.
Also, the special handling is removed from `BooleanParser` to get the
tests to pass. Without the adjustment, either uncached or cached
templates would be correct, but never both. The new version delivers
the same results as cached templates before this change, so it should
be in line with real-world projects.

### DIFF
--- a/src/Core/Parser/BooleanParser.php
+++ b/src/Core/Parser/BooleanParser.php
@@ -347,14 +347,6 @@ class BooleanParser
      */
     protected function evaluateCompare(mixed $x, mixed $y, string $comparator)
     {
-        // enfore strong comparison for comparing two objects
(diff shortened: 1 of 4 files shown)


s2b:
Will only be part of 4.6

PR #1384, opened 2026-07-01
[BUGFIX] Handle generators and traversables correctly in ForViewHelper

Kanti:
## What changed

This PR improves how `ForViewHelper` handles generators and other traversables.

The ViewHelper sometimes has to look at all items before rendering them, for example when `reverse="true"` is used or when `{iterator.total}` is needed.
Until now, that was done with `iterator_to_array()`.
That works for simple cases, but it loses entries when a generator yields the same key more than once.
The implementation now keeps materialized traversable items as key-value pairs instead.
That way, every yielded item is preserved, including duplicate keys.

Plain generator loops are still rendered directly whenever possible, 
so we avoid consuming or buffering generators unless Fluid actually needs to know the full item list up front.

## Why
PHP `foreach` renders every yielded generator item, even if keys are duplicated.
`ForViewHelper` should behave the same way.
Without this change, using `iteration` or `reverse` could silently drop entries from a valid generator.

## Tests

This adds test coverage for:

  - non-countable traversables with `{iterator.total}`
  - plain generator rendering without look-ahead
  - generators with `{iterator.total}`
  - generators that yield duplicate keys
  - duplicate generator keys without exposing the key argument
  - empty generators with `{iterator.total}`
  - empty generators with `reverse="true"`
  - duplicate generator keys together with `reverse="true"`
  - generators with both `reverse="true"` and `{iterator.total}`
  - duplicate generator keys together with `reverse="true"` and `{iterator.total}`
  - countable traversables with iteration metadata

### DIFF
--- a/src/ViewHelpers/ForViewHelper.php
+++ b/src/ViewHelpers/ForViewHelper.php
@@ -99,8 +99,18 @@ class ForViewHelper extends AbstractViewHelper
         if (is_object($this->arguments['each']) && !$this->arguments['each'] instanceof \Traversable) {
             throw new InvalidArgumentValueException('ForViewHelper only supports arrays and objects implementing \Traversable interface', 1248728393);
         }
+        $isKeyValuePair = false;
+        if ($this->arguments['reverse'] === true || (isset($this->arguments['iteration'])) && !is_countable($this->arguments['each'])) {
+            $isKeyValuePair = true;
+            $traversable = $this->arguments['each'];
+            $this->arguments['each'] = [];
+            foreach ($traversable as $key => $singleElement) {
+                // unpack the traversable, but keep duplicate keys.
+                $this->arguments['each'][] = [$key, $singleElement];
+            }
+        }
         if ($this->arguments['reverse'] === true) {
-            $this->arguments['each'] = array_reverse(iterator_to_array($this->arguments['each']), true);
+            $this->arguments['each'] = array_reverse($this->arguments['each'], true);
         }
         if (isset($this->arguments['iteration'])) {
             $iterationData = [
@@ -114,6 +124,9 @@ class ForViewHelper extends AbstractViewHelper
         $this->renderingContext->setVariableProvider(new ScopedVariableProvider($globalVariableProvider, $localVariableProvider));
         $output = '';
         foreach ($this->arguments['each'] as $keyValue => $singleElement) {
+            if ($isKeyValuePair === true) {
+                [$keyValue, $singleElement] = $singleElement;
+            }
             $localVariableProvider->add($this->arguments['as'], $singleElement);
             if (isset($this->arguments['key'])) {
                 $localVariableProvider->add($this->arguments['key'], $keyValue);
--- a/tests/Functional/ViewHelpers/ForViewHelperTest.php
+++ b/tests/Functional/ViewHelpers/ForViewHelperTest.php
@@ -107,6 +107,147 @@ final class ForViewHelperTest extends AbstractFunctionalTestCase
             . '</ul>',
         ];
 
+        $value = new class () implements \IteratorAggregate {
+            public function getIterator(): \Traversable
+            {
+                return new \ArrayIterator([
+                    'first' => 'foo',
+                    'second' => 'bar',
+                ]);
+            }
+        };
+        yield 'iterator total works for non countable traversable' => [
+            '<f:for each="{value}" as="item" iteration="iterator">{item}: {iterator.total}, </f:for>',
+            ['value' => $value],
+            'foo: 2, bar: 2, ',
+        ];
+
+        yield 'generator gets traversed' => [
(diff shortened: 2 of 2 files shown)


src/ViewHelpers/ForViewHelper.php
```diff
@@ -106,7 +106,7 @@ public function render(): string
             $iterationData = [
                 'index' => 0,
                 'cycle' => 1,
-                'total' => count($this->arguments['each']),
+                'total' => count(iterator_to_array($this->arguments['each'])),
```
mbrodala: IMO we should not do this silently but fail explicitly instead. For example unwinding a `Generator` only for counting can be expensive if it is then traversed again.

mbrodala (changes requested):

src/ViewHelpers/ForViewHelper.php
Kanti: resolved with `is_countable($this->arguments['each']) ? $this->arguments['each'] : iterator_to_array($this->arguments['each'])`

mbrodala: Sorry, this does not solve the issue I mentioned. Given a `Generator` we would still convert it to an array purely for counting.

As said we should explicitly fail with an error instead. This is way better than any silent assumption. It already fails with an error but this only implicitly hints at the missing `Countable` implementation. This should be spelled out explicitly instead. Maybe with a suggestion to skip `iteration` if implementing `Countable` is not possible.

src/ViewHelpers/ForViewHelper.php
```diff
@@ -99,21 +99,39 @@ public function render(): string
...
             throw new InvalidArgumentValueException('ForViewHelper only supports arrays and objects implementing \Traversable interface', 1248728393);
         }
+        $items = null;
         if ($this->arguments['reverse'] === true) {
-            $this->arguments['each'] = array_reverse(iterator_to_array($this->arguments['each']), true);
+            if (is_array($this->arguments['each'])) {
+                $this->arguments['each'] = array_reverse($this->arguments['each'], true);
+            } else {
+                $items = [];
+                foreach ($this->arguments['each'] as $keyValue => $singleElement) {
```
mbrodala: I'm still not a fan of preemptively unwinding a `Generator` which may produce a lot of items. One point of generators is reducing memory usage which would be rendered useless here.

Still that's the behavior we always had so far so I guess it's acceptable.

src/ViewHelpers/ForViewHelper.php
```diff
@@ -99,21 +99,39 @@ public function render(): string
...
+                'total' => count($items ?? $this->arguments['each']),
             ];
         }
         $globalVariableProvider = $this->renderingContext->getVariableProvider();
         $localVariableProvider = new StandardVariableProvider();
         $this->renderingContext->setVariableProvider(new ScopedVariableProvider($globalVariableProvider, $localVariableProvider));
         $output = '';
-        foreach ($this->arguments['each'] as $keyValue => $singleElement) {
+        foreach ($items ?? $this->arguments['each'] as $keyValue => $singleElement) {
+            if ($items !== null) {
```
mbrodala: See above about readability.

src/ViewHelpers/ForViewHelper.php
```diff
@@ -99,21 +99,39 @@ public function render(): string
...
+            } else {
+                $items = [];
+                foreach ($this->arguments['each'] as $keyValue => $singleElement) {
+                    $items[] = [$keyValue, $singleElement];
+                }
+                $items = array_reverse($items);
+            }
         }
         if (isset($this->arguments['iteration'])) {
+            if ($items === null && !is_countable($this->arguments['each'])) {
```
mbrodala: This is rather hard to read and understand. We could instead always use $items as the result of a normalization from various input formats.

Also we completely skip `Countable` here. One can in fact implement a very memory friendly object which iterates using a generator and implement `Countable`. (Think e.g. `SELECT *` & `COUNT(*)` for DB data.)

mbrodala (changes requested):

src/ViewHelpers/ForViewHelper.php
Kanti: The generator can only be read once and cannot be reversed unless we collect all the items and reverse them.
Therefore, there is no other possible solution here.

The same problem applies if we need the count for each of the elements.

src/ViewHelpers/ForViewHelper.php
Kanti: I could not find a solution where we could always use `$items` while keeping the generators streaming the items in case `iteration` and `reverse` are not used.
This sadly gives us two different ways to iterate.

I will try to make it more readable

Kanti: I think it's more readable now. What do you think?

